PR #228, opened 2015-01-27
Placeholder icon redecorating

daines:
screen shot 2015-01-27 at 12 12 44 pm

* Remove per-format image placeholders in favor of generic icons with format labels
* Change the folder and collection icons to a flatter style, use an "archive box" icon for generic collections
* Change display of deleted items in access interface to be translucent and have a "trash" icon badge
* Replace icons in the admin interface to match the new style in the access interface
* Display a lock badge on items in the access interface which either have an active embargo or do not have the "patron" role for the "public" group, even if the current user has access to those items
* Add Font Awesome webfont for icon badges
* Centralize thumbnail stuff in one template
* Remove the jQuery.preload plugin in favor of simple class swapping for thumbnail loading placeholders


access/src/main/external/static/js/public/fullRecord.js
```diff
@@ -20,7 +19,7 @@ require.config({
 	},
 	shim: {
 		'jquery-ui' : ['jquery'],
-		'thumbnails' : ['jquery', 'preload'],
+		'thumbnails' : ['jquery'],
```
bbpennel: This shim may not be needed anymore with preload out of the picture

PR #241, opened 2015-02-18
Admin public linkage

bbpennel:
Improved linkage back and forth between the public and admin UI's.


bbpennel:
I would like to wait for a bit more feedback from cdr-discuss before merging

PR #255, opened 2015-03-18
Introduce resubmitting ingests

daines:
- Add ResubmitIngestController which checks resubmitted packages and saves them to the batch ingest directory
- Add PrepareResubmitJob which backs up the previous ingest, arranges the resubmitted ingest package for processing by the rest of the pipeline, and marks the deposit as a resubmission
- Modify IngestDeposit to consider "object exists" exceptions confirmation that the object has already been ingested when the deposit is marked as a resubmission
- Add ReingestPackageForm and "Reingest" link to details pane in status monitor

### DIFF
--- a/admin/src/main/java/edu/unc/lib/dl/admin/controller/ResubmitIngestController.java
+++ b/admin/src/main/java/edu/unc/lib/dl/admin/controller/ResubmitIngestController.java
@@ -73,14 +73,14 @@ public class ResubmitIngestController {
 		}
 		
 		
-		// Check that the deposit is in the failed state
+		// Check that the deposit is in the failed or paused state
 		
 		String state = status.get(DepositField.state.name());
 		
-		if (state == null || !state.equals(DepositState.failed.name())) {
+		if (state == null || !(state.equals(DepositState.failed.name()) || state.equals(DepositState.paused.name()))) {
 			response.setStatus(400);
-			result.put("error", "The deposit isn't in the failed state.");
-			log.error("Tried to resubmit deposit {} but it isn't in the failed state", depositPid);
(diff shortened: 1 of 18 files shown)


admin/src/main/java/edu/unc/lib/dl/admin/controller/ResubmitIngestController.java
```diff
@@ -0,0 +1,176 @@
...
+		
+		
+		// Scan the ingest file
+		
+		MetsHeaderScanner scanner = new MetsHeaderScanner();
+		
+		try {
+			scanner.scan(resubmitPackageFile, uploadedFile.getOriginalFilename());
+		} catch (Exception e) {
+			response.setStatus(500);
```
bbpennel: I'm glad that it is giving less crazy error messages to the client, but we probably do need to have some way to get to the stacktrace of why the ingest failed.  That could just be dumping a stacktrack through log4j

admin/src/main/java/edu/unc/lib/dl/admin/controller/ResubmitIngestController.java
```diff
@@ -0,0 +1,176 @@
...
+			response.setStatus(401);
+			return result;
+		}
+		
+		
+		// Check that the deposit is in the failed state
+		
+		String state = status.get(DepositField.state.name());
+		
+		if (state == null || !state.equals(DepositState.failed.name())) {
```
bbpennel: Perhaps it should allow you to resubmit if the ingest is paused?  Mainly we want to prevent resubmission when an ingest is active or could become active at any moment.

PR #268, opened 2015-03-30
Ensure breadcrumb value will reflect changes to parent container titles

bbpennel:
No longer indexing ancestor names so that they will not get out of date after descriptive metadata changes.  They are now retrieved at access time, either from the solr index or from a cache of path information.

Introduced ObjectPath and ObjectPathFactory classes, which retrieve and store name information for objects along a path since that is no longer tracked in AncestorPath.


solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/BriefObjectMetadata.java
```diff
@@ -26,6 +26,10 @@
 public interface BriefObjectMetadata {
 	public CutoffFacet getAncestorPathFacet();
 
+	public ObjectPath getObjectPath();
+
+	public void setObjectPath(ObjectPath pathObject);
```
daines: Change "pathObject" to "objectPath"?

access/src/main/webapp/WEB-INF/jsp/common/hierarchyTrail.jsp
```diff
@@ -41,24 +41,24 @@
 		</c:url>
 		<a href="<c:out value="${shiftFacetUrl}"/>">Home</a>
 	</c:if>
-	<c:forEach items="${facetNodes}" var="facetNode" varStatus="status">
+	<c:forEach items="${pathObject.entries}" var="pathEntry" varStatus="status">
```
daines: pathObject or objectPath?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/CutoffFacet.java
```diff
@@ -145,7 +145,7 @@ public String getDisplayValue() {
 		CutoffFacetNode lastNode = this.getHighestTierNode();
 		if (lastNode == null)
 			return null;
-		return lastNode.getDisplayValue();
+		return lastNode.getSearchKey();
```
daines: Why use getSearchKey rather than getDisplayValue?

bbpennel: CutoffFacets don't have a display value anymore, that was previously that label.  However, I should replace the contents of this method with a call to CutoffFacet.getSearchKey instead of reimplementing it.

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/BriefObjectMetadata.java
bbpennel: Got a little dyslexic when I entered the name originally and looks like I missed a few places when cleaning up.  Thanks for catching, I'll fix it!

PR #279, opened 2015-04-23
Ensure that we are actually asking for optimistic locking for moves

daines:
- Make sure the lastModifiedDate parameter is actually added when attempting to modify a datastream
- Retrieve a datastream's metadata before retrieving its content so that we have its last modified date to millisecond resolution, rather than second resolution as available from the Last-Modified header
- Clean up debug logging in move methods (removeChildren, addChildren, cleanupRemovedChildren)

### DIFF
--- a/persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
+++ b/persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
@@ -26,9 +26,11 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -615,7 +617,7 @@ public class DigitalObjectManagerImpl implements DigitalObjectManager {
 
 	/**
 	 * Returns response containing the jdom document representing the datastream and the last modified date. If it does
-	 * not exist, then null is returned. If the document cannot be parsed
+	 * not exist, then null is returned. If the document cannot be parsed, a ServiceException is thrown.
 	 *
 	 * @param pid
 	 * @param datastreamName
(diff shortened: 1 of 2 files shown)


persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
```diff
@@ -623,29 +625,42 @@ public String getLastModified() {
...
+			
+			while (true) {
+		
+				edu.unc.lib.dl.fedora.types.Datastream datastream = managementClient.getDatastream(pid, datastreamName);
+			
+				if (datastream == null) {
+					return null;
 				}
+				
+				log.debug("Got datastream, attempting to get dissemination version with create date " + managementClient.getDatastream(pid, datastreamName).getCreateDate());
```
bbpennel: Guessing this is just in here for testing?  Probably don't want it re-retrieval the datastream profile.

persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
```diff
@@ -623,29 +625,42 @@ public String getLastModified() {
...
-					lastModified = header.getValue();
-					break;
+		
+		log.debug("Attempting to get datastream " + datastreamName + " for object " + pid);
+		
+		try {
+			
+			while (true) {
+		
+				edu.unc.lib.dl.fedora.types.Datastream datastream = managementClient.getDatastream(pid, datastreamName);
```
bbpennel: This is likely the better way to check for the presence of a datastream.  Do you have an idea of if it has a comparable execution time to the triplestore check in the previous version?

persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
```diff
@@ -623,29 +625,42 @@ public String getLastModified() {
...
+			
+				if (datastream == null) {
```
bbpennel: Are both this conditional and the NotFoundException catch necessary?  Is the exception not being thrown every time?

persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
```diff
@@ -623,29 +625,42 @@ public String getLastModified() {
...
+				
+				log.debug("Got datastream, attempting to get dissemination version with create date " + managementClient.getDatastream(pid, datastreamName).getCreateDate());
+			
+				try {
+			
+					MIMETypedStream mts = accessClient.getDatastreamDissemination(pid, datastreamName, datastream.getCreateDate());
+				
+					try (ByteArrayInputStream bais = new ByteArrayInputStream(mts.getStream())) {
+						Document dsDoc = new SAXBuilder().build(bais);
+						return new DatastreamDocument(dsDoc, datastream.getCreateDate());
```
bbpennel: So it ended up being okay to use the timestamp returned by Fedora in the getDatastream request?  It has enough precision to be useful for optimistic locking, unlike the modified header in the getDatastreamDissemination call?

persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
daines: I'm not sure about execution time. This isn't only a check for existence, though. We need the "create date" for the latest version of the datastream so we can set lastModified on the DatastreamDocument object we return.

persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
daines: It seems to always be null, but getDatastream does throw FedoraException.

persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
daines: oops, should just be datastream.getCreateDate()

persistence/src/main/java/edu/unc/lib/dl/services/DigitalObjectManagerImpl.java
daines: Yes.

PR #286, opened 2015-05-05
Sending Add message after a deposit completes

bbpennel:


deposit/src/main/resources/service-context.xml
```diff
@@ -122,6 +122,12 @@
 		<property name="htmlTemplate" ref="htmlTemplate" />
 		<property name="textTemplate" ref="textTemplate" />
 	</bean>
+	
+	<bean id="depositMessageHandler" class="edu.unc.lib.deposit.work.DepositMessageHandler" >
```
daines: Should also be added to test service-context.xml?

PR #294, opened 2015-05-15
Updated to spring 4.1.6

bbpennel:


bbpennel:
I have tested: simple object ingest, sip based ingest, moves, edit access/publish/delete, verified that at least one scheduled task is running, navigation in admin and access, file retrieval, thumbnail generation.

All were tested in my vm, while some of them were tested on dev as well.

PR #296, opened 2015-05-20
Move reporting

bbpennel:
Tracks ongoing move operations on the server.  Clients poll server for changes to set of ongoing move operations, displaying which items are actively being moved to all users.  Informs originating user when their move operations complete.

### DIFF
--- a/access/src/main/external/static/js/admin/src/MoveActionMonitor.js
+++ b/access/src/main/external/static/js/admin/src/MoveActionMonitor.js
@@ -8,7 +8,7 @@ define('MoveActionMonitor', [ 'jquery'], function($) {
 		this.options = $.extend({}, defaultOptions, options);
 		
 		this.alerts = alertHandler;
-		this.processedCompletes = [];
+		this.moveData = {};
 	}
 	
 	function setDifference(A, B) {
@@ -34,8 +34,7 @@ define('MoveActionMonitor', [ 'jquery'], function($) {
 		
 		// Mark objects that were previously known to be moving
 		var self = this;
-		var moveData = this.getMoveData();
-		$.each(moveData, function(moveId, moveDetails) {
+		$.each(this.moveData, function(moveId, moveDetails) {
 			self.markMoving(moveDetails.pids);
 		});
 		
@@ -52,16 +51,13 @@ define('MoveActionMonitor', [ 'jquery'], function($) {
 			contentType: "application/json; charset=utf-8",
 			dataType: "json",
 			success : function(data) {
-				var moveData = self.getMoveData();
-				var localMoves = Object.keys(moveData);
-				
 				var remoteMoves = data == null? {} : data;
 				
 				try {
-					
-					self.cleanMoveTombstones(moveData, remoteMoves);
+					// Clean up inactive move operations that have been removed remotely
+					self.cleanMoveTombstones(remoteMoves);
 					// Clean up completed move operations
-					self.completeMoves(moveData, remoteMoves);
+					self.completeMoves(remoteMoves);
 				
 					// Indicate new move options
 					self.addMoves(remoteMoves);
@@ -73,45 +69,38 @@ define('MoveActionMonitor', [ 'jquery'], function($) {
 			},
 			error : function() {
 				console.error("Failed to retrieve move information from server");
+				// Keep refreshing alive, but wait a bit longer
+				setTimeout($.proxy(self.update, self), self.options.updateInterval * 2);
 			}
 		});
 	};
 	
-	MoveActionMonitor.prototype.cleanMoveTombstones = function(moveData, remoteMoves) {
+	MoveActionMonitor.prototype.cleanMoveTombstones = function(remoteMoves) {
+
 		var localInactive = [];
-		for (var moveId in moveData) {
-			if (moveData[moveId].inactive) {
+		for (var moveId in this.moveData) {
+			if (this.moveData[moveId].inactive) {
 				localInactive.push(moveId);
 			}
 		}
 		
 		if (localInactive.length == 0)
 			return;
-		
-		// Catch moves that completed in another browser window
-		var unprocessedComplete = setDifference(localInactive, this.processedCompletes);
-		for (var index in unprocessedComplete) {
-			this.cleanupResults(moveData[unprocessedComplete[index]].pids);
-		}
-		
+
 		// Cleanable tombstones = Inactive Local Moves - Remote Moves
 		var cleanable = setDifference(localInactive, Object.keys(remoteMoves));
 		
 		if (cleanable.length == 0)
 			return;
 		
-		try {
-			for (var index in cleanable) {
-				var cleanableId = cleanable[index];
-				delete moveData[cleanableId];
-				delete this.processedCompletes[cleanableId];
-			}
-		} finally {
-			this.setMoveData(moveData);
+		for (var index in cleanable) {
+			var cleanableId = cleanable[index];
+			delete this.moveData[cleanableId];
+			this.removeUserMove(cleanableId);
 		}
 	};
 	
-	MoveActionMonitor.prototype.completeMoves = function(moveData, remoteMoves) {
+	MoveActionMonitor.prototype.completeMoves = function(remoteMoves) {
 
 		// Get the set of move ids which were in progress according to the server
 		var remoteActive = [];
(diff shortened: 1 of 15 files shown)


lfarrell:
As best as I can tell it looks fine. One thing with the Ajax requests per the jQuery docs:

"Deprecation Notice: The jqXHR.success(), jqXHR.error(), and jqXHR.complete() callbacks are deprecated as of jQuery 1.8. To prepare your code for their eventual removal, use jqXHR.done(), jqXHR.fail(), and jqXHR.always() instead."

Might want to switch to the new Promises model???

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
```diff
@@ -0,0 +1,255 @@
...
+		this.active = true;
+		
+		// Mark objects that were previously known to be moving
+		var self = this;
+		$.each(this.moveData, function(moveId, moveDetails) {
+			self.markMoving(moveDetails.pids);
+		});
+		
+		// Start refresh loop
+		this.update();
```
daines: Why not use setInterval?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
```diff
@@ -85,8 +127,12 @@ public void setDigitalObjectManager(DigitalObjectManager digitalObjectManager) {
 	public static class MoveRequest {
 		private String newParent;
 		private List<String> ids;
+		private String user;
+		private String moveId;
```
daines: Maybe just "id"?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
```diff
@@ -85,8 +127,12 @@ public void setDigitalObjectManager(DigitalObjectManager digitalObjectManager) {
 	public static class MoveRequest {
 		private String newParent;
 		private List<String> ids;
```
daines: Maybe indicate specifically that these are the ids being moved? "moved"?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
```diff
@@ -85,8 +127,12 @@ public void setDigitalObjectManager(DigitalObjectManager digitalObjectManager) {
 	public static class MoveRequest {
 		private String newParent;
```
daines: Why not "destination"?

bbpennel: I would prefer that too, it just that it was already using the name newParent before this PR.  Changing the names of fields in this class will require updating things on the javascript side as well, which is fine, I was just avoiding extra changes.  I'll revisit the names though.

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
bbpennel: I did at first, but switched to make sure that the next run of the function wouldn't start until the ajax call from the previous one had finished in case the server decided to take a really long time to respond

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
```diff
@@ -46,32 +50,70 @@
...
+	public @ResponseBody
+	Object listMoves() {
+		Map<String, Boolean> moves = new HashMap<>(this.activeMoveRequests.size());
+		
+		Iterator<Entry<String, MoveRequest>> moveIt = activeMoveRequests.entrySet().iterator();
+		while (moveIt.hasNext()) {
+			Entry<String, MoveRequest> move = moveIt.next();
+			long finishedAt = move.getValue().finishedAt;
+			if (finishedAt == -1 || finishedAt + TIME_TO_LIVE_AFTER_FINISH > System.currentTimeMillis()) {
+				moves.put(move.getKey(), finishedAt == -1);
```
daines: What is the difference between `moves` having a value of `false` for a key and that key not being present?

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
```diff
@@ -0,0 +1,250 @@
...
+		
+		// Get the list of move ids considered in progress locally
+		var localActive = [];
+		for (var moveId in this.moveData) {
+			if (!this.moveData[moveId].inactive) {
+				localActive.push(moveId);
+			}
+		}
+		
+		var localMoves = Object.keys(this.moveData);
```
daines: If `localActive` is the list of move ids considered in progress locally, what is `localMoves`?

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
```diff
@@ -0,0 +1,250 @@
...
+			var cleanableId = cleanable[index];
+			delete this.moveData[cleanableId];
+			this.removeUserMove(cleanableId);
+		}
+	};
+	
+	MoveActionMonitor.prototype.completeMoves = function(remoteMoves) {
+
+		// Get the set of move ids which were in progress according to the server
+		var remoteActive = [];
```
daines: Why does removeMoves need to be converted into an array? (Besides the fact that the setDifference function takes an array as input.)

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
```diff
@@ -0,0 +1,250 @@
...
+		var localMoves = Object.keys(this.moveData);
+		// Completed jobs = Local Moves - Active Remote Moves
+		var completed = setDifference(localActive, remoteActive);
+		
+		if (completed.length == 0)
+			return;
+		
+		var userMoves = this.getUserMoves();
+		// Process the set of completed moves to remove them from local memory and inform the user
+		for (var index in completed) {
```
daines: If `completed` is an array, why are you using for ... in?

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
```diff
@@ -0,0 +1,250 @@
...
+	};
+	
+	function MoveActionMonitor(alertHandler, options) {
+		this.options = $.extend({}, defaultOptions, options);
+		
+		this.alerts = alertHandler;
+		this.moveData = {};
+	}
+	
+	function setDifference(A, B) {
```
daines: Why are arrays used to represent sets here? Why not use objects with move ids as the keys?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
bbpennel: having a value of false means that the job is not active, but the server is still reporting on it.  It prevents the client from looking up info about the job after it finishes but before it has been cleaned up

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
bbpennel: localMoves is all moves tracked locally, inactive or active.  localActive is just the moves tracked locally that are still in progress.

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
bbpennel: It is generating a subset of which remove operations are considered to still be in progress on the server side.  The server will respond with moves which have finished, but then hold the moves for another 12 seconds as completed.  So the client determines which moves are newly inactive by subtracting the remotely active moves from the locally active ones.

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
bbpennel: Which structure would you prefer?  It seems about the same as a for (var i = 0; i < completed.length; i++) and I generally haven't been using .each

access/src/main/external/static/js/admin/src/MoveActionMonitor.js
```diff
@@ -0,0 +1,216 @@
...
+			contentType: "application/json; charset=utf-8",
+			dataType: "json"
+		}).done(function(data) {
+			try {
+				// Clean up inactive move operations that have been removed remotely
+				self.cleanMoveTombstones(data);
+				// Clean up completed move operations
+				self.completeMoves(data);
+			
+				// Indicate new move options
```
daines: Typo? Should be "operations"?

PR #300, opened 2015-06-01
Rename premis event

lfarrell:
Add a premis even for the file renaming operation


daines:
Is the meat of this PR just commit 22f2de1 or does it involve the other commits? If it's just 22f2de1, could you change this PR to be based on master?

lfarrell:
Yeah, it's just the last commit. I think I branched off a branch.

PR #301, opened 2015-06-01
Add Premis event to the file/container renaming process.

lfarrell:
Add a Premis event to track when a file/container has been renamed

### DIFF
--- a/admin/src/main/java/edu/unc/lib/dl/admin/controller/EditLabelController.java
+++ b/admin/src/main/java/edu/unc/lib/dl/admin/controller/EditLabelController.java
@@ -35,8 +35,8 @@ public class EditLabelController {
 
 				this.client.modifyObject(pidObject, label, null, null, null);
 				
-				Element event = logger.logEvent(PremisEventLogger.Type.NORMALIZATION, "Object renamed to " + label, pidObject);
-				PremisEventLogger.addDetailedOutcome(event, "success", "Message: Object renamed successfully", null);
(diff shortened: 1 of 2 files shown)


admin/src/main/java/edu/unc/lib/dl/admin/controller/EditLabelController.java
```diff
@@ -31,7 +30,14 @@ Object saveLabel(@PathVariable("pid") String pid,
 		
 		if (label != null && label.trim().length() > 0) {
 			try {
-				this.client.modifyObject(new PID(pid), label, null, null, null);
+				PremisEventLogger logger = new PremisEventLogger(pid);
+				PID pidObject = new PID(pid);
+
+				this.client.modifyObject(pidObject, label, null, null, null);
+				
+				Element event = logger.logEvent(PremisEventLogger.Type.NORMALIZATION, "Object renamed to " + label, pidObject);
```
bbpennel: We used MIGRATION as the event type for move operations, so even though that doesn't exactly match the definition of that type of event, we should switch this for consistency.

admin/src/main/java/edu/unc/lib/dl/admin/controller/EditLabelController.java
```diff
@@ -31,7 +30,14 @@ Object saveLabel(@PathVariable("pid") String pid,
...
+				PremisEventLogger.addDetailedOutcome(event, "success", "Message: Object renamed successfully", null);
```
bbpennel: Remove the "Message: " part from the beginning here

PR #302, opened 2015-06-01
Mods controller encoding

lfarrell:


lfarrell:
Explicitly set encoding for JSON response.

PR #305, opened 2015-06-04
Use Jesque for queues in services

daines:
- Introduce ApplyEnhancementServicesJob and SolrUpdateJob
- EnhancementConductor and SolrUpdateConductor now use Jesque
- Remove existing conductors from service context
- Remove unused classes and interface code
- Remove unused MessageConductor methods


services/src/main/resources/log4j.properties
```diff
@@ -15,7 +15,7 @@
 #
 
 # Root logger set to DEBUG, but SERVER.Threshold blocks below WARN
-log4j.rootLogger=INFO, SERVER, CDRSERVICES
+log4j.rootLogger=DEBUG, SERVER, CDRSERVICES
```
bbpennel: we'll want to switch this back before merging

PR #306, opened 2015-06-11
Default ordering

lfarrell:
Change default and collection searches to sort by title and label only


bbpennel:
Tests aren't passing:
getACollectionTest(edu.unc.lib.dl.ui.service.StructureQueryTest): sort param field can't be found: null
  getEmptyFolderStructureTest(edu.unc.lib.dl.ui.service.StructureQueryTest): sort param field can't be found: null
  getCollectionsTest(edu.unc.lib.dl.ui.service.StructureQueryTest): sort param field can't be found: null

My guess is that you need to define a missing field, probably label.

access-common/src/test/resources/search.properties
```diff
@@ -109,8 +109,8 @@ search.sort.type.creator=CREATOR_SORT|asc
 search.sort.type.dateUpdated=DATE_UPDATED|desc
 search.sort.type.dateCreated=DATE_CREATED|desc
 search.sort.type.dateAdded=DATE_ADDED|desc
-search.sort.type.collection=ANCESTOR_IDS|asc,RESOURCE_TYPE_SORT|asc,DISPLAY_ORDER|asc
-search.sort.type.default=RESOURCE_TYPE_SORT|asc,SCORE|desc,ANCESTOR_IDS|asc,DISPLAY_ORDER|asc
+search.sort.type.collection=TITLE|asc,LABEL|asc
+search.sort.type.default=TITLE|asc,LABEL|asc
```
bbpennel: I think the default sort should be:
RESOURCE_TYPE_SORT|asc,SCORE|desc,TITLE|asc,LABEL|asc

This sort is used for search results, so we want the score to be a factor.  I am not sure about excluding RESOURCE_TYPE_SORT, which causes collections and folders to show up at the top of the result.  I think that is preferable behavior when browsing for navigation reasons, but doesn't make sense for search results.  So perhaps searches should default to a different sort.

bbpennel:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/312

switching to squashed PR

PR #307, opened 2015-06-16
Change resource types of containers

bbpennel:
* Implemented operation for changing resource type/content models of a container
* Added dialog, context menu options, and button for performing the operation
* Added indexing action for recursive indexing of sets of pids using just relations
* Moved ResourceType enum to the metadata module


access/src/main/external/static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -134,8 +134,13 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'E
 			}
 		}
 		
+		if ($.inArray('changeResourceType', metadata.permissions) != -1
+				&& $.inArray('info:fedora/cdr-model:Container', metadata.model) != -1) {
+			items["changeType"] = {name : 'Change Type'};
```
daines: Menu items that modify the object should be grouped together. We could have something like this:

```
--
Unpublish
Edit Access
Edit Label
Edit Description
Change Type
--
Export as CSV
Copy PID to Clipboard
--
```

daines: Also, could this be called "Edit Type" instead? Why is this "change" but the other options are "edit"?

access/src/main/external/static/templates/admin/changeTypeForm.html
```diff
@@ -0,0 +1,22 @@
+<form id="change_type_form" method="post" action="services/api/edit/changeType" class="edit_form change_type_form">
+	<p>
+		<div class="form_field">
+			<label>New Type</label>
+			<select id="change_type_new_type">
+				<option value="Aggregate" <%= "Aggregate Work" == defaultType? "disabled" : "" %>>Aggregate Work</option>
```
daines: I think all of the options should be available -- otherwise it's easy to accidentally change the type.

In that case, the label above should read "Type" rather than "New Type".

access/src/main/external/static/templates/admin/changeTypeForm.html
```diff
@@ -0,0 +1,22 @@
...
+				<option value="Folder" <%= "Folder" == defaultType? "disabled" : "" %>>Folder</option>
+				<option value="Collection" <%= "Collection" == defaultType? "disabled" : " "%>>Collection</option>
+			</select>
+		</div>
+	
+		<div class="update_field">
+			<input type="submit" id="change_type_submit_button" class="update_button" value="Change" />
```
daines: We should standardize on a title for buttons like this. The access controls modal has "Update", this has "Change", and edit label and edit description have "Save Changes".

access/src/main/external/static/js/admin/src/action/ChangeTypeBatchAction.js
```diff
@@ -0,0 +1,84 @@
...
+	};
+	
+	ChangeTypeBatchAction.prototype.execute = function() {
+		var self = this;
+		
+		this.targets = this.getTargets();
+		var title;
+		var defaultType;
+		if (this.targets.length == 1) {
+			title = "Change type for " + this.targets[0].metadata.title.substring(0, 30);
```
daines: "Type" should be capitalized.

PR #324, opened 2015-07-01
Mimetypes assignment changes

bbpennel:


services/src/main/java/edu/unc/lib/dl/cdr/services/techmd/TechnicalMetadataEnhancement.java
```diff
@@ -131,45 +136,51 @@ public Element call() throws EnhancementException {
...
-				}
+				
+				// If fedora mimetype is not meaningful, then override with FITS generated value
+				if (fedoraMimetype == null || fedoraMimetype.trim().length() == 0
+						|| fedoraMimetype.contains("octet-stream")) {
+					
+					// IDENTIFICATION LOGIC
+					// get mimetype out of FITS XML
+					Element identity = null;
+					Element idn = fits.getRootElement().getChild("identification", ns);
```
daines: What about using an XPath expression to find the right identity element?

bbpennel: I don't think I would be able to select the preferred identity with an xpath expression?  I think I'd prefer to stick with the current implementation.

services/src/main/java/edu/unc/lib/dl/cdr/services/techmd/TechnicalMetadataEnhancement.java
```diff
@@ -131,45 +136,51 @@ public Element call() throws EnhancementException {
...
-				if (fedoraMimetype != null && fedoraMimetype.trim().length() > 0
-						&& !fedoraMimetype.contains("octet-stream")) {
-					fitsMimetype = fedoraMimetype;
+					if (identity != null) {
+						fitsMimetype = identity.getAttributeValue("mimetype");
+						format = identity.getAttributeValue("format");
+					} else {
+						format = "Unknown";
+						LOG.warn("FITS unable to conclusively identify file: {}/{}", pid, dsid);
+						LOG.info(new XMLOutputter().outputString(fits));
```
bbpennel: Never noticed this before, not sure about dumping the FITS XML at info level.

daines: Maybe debug?

services/src/main/java/edu/unc/lib/dl/cdr/services/techmd/TechnicalMetadataEnhancement.java
daines: Maybe, but the logic here is a little weird. Why get a reference to the <identification> element in order to test one of its attribute values for every iteration of the loop below?

daines: What about this:

```
//identification[not(@status)]/identity[@mimetype!='application/x-symlink' and not(tool[@toolname='Exiftool'])]
```

bbpennel: True, the status test every loop through is weird.  The xpath looks like it'd work, although i haven't tested it.

bbpennel: Selecting identity exclusively with an xpath doesn't seem like it is going to work without getting pretty ugly, since we can't just exclude identities that include an Exiftool entry, since FITS rolls multiple tools up into a single identity.  For example:

```xml
<fits xmlns="http://hul.harvard.edu/ois/xml/ns/fits/fits_output" xsi:schemaLocation="http://hul.harvard.edu/ois/xml/ns/fits/fits_output http://hul.harvard.edu/ois/xml/xsd/fits/fits_output.xsd" version="0.8.6" timestamp="7/14/15 2:34 PM">
  <identification>
    <identity format="Portable Document Format" mimetype="application/pdf" toolname="FITS" toolversion="0.8.6">
      <tool toolname="Jhove" toolversion="1.5" />
      <tool toolname="Tika" toolversion="1.3" />
      <tool toolname="Droid" toolversion="6.1.3" />
      <tool toolname="Exiftool" toolversion="9.13" />
      <version toolname="Jhove" toolversion="1.5">1.3</version>
      <externalIdentifier toolname="Droid" toolversion="6.1.3" type="puid">fmt/17</externalIdentifier>
    </identity>
  </identification>
....
```
wouldn't return an identity because exif is agreeing with everything else.  

I am going to stick with the loop approach and cleanup the status evaluation bit.  We can revisit our approach to deciding on what is the preferred FITS mimetype later if we are still having issues.

PR #326, opened 2015-07-09
Xml export

bbpennel:
Exports an XML document containing the MODS for a set of selected objects.  The set can either be specified as a tree starting from a selected folder or by selecting a set of specific objects via the admin interface.  The exported XML document is sent to the provided email address (or the user's default email address).

Moved getXMLDatastreamIfExists from DigitalObjectManager to ManagementClient so that it can be more broadly used, and because it's a more fitting home

### DIFF
--- a/admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
+++ b/admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
@@ -1,0 +1,151 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.dl.admin.controller;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.Resource;
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.solr.client.solrj.beans.Field;
+import org.jdom2.Document;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.core.io.ByteArrayResource;
+import org.springframework.core.io.InputStreamResource;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.jdom2.output.XMLOutputter;
+
+import edu.unc.lib.dl.acl.util.GroupsThreadStore;
+import edu.unc.lib.dl.fedora.AccessClient;
+import edu.unc.lib.dl.fedora.FedoraDataService;
+import edu.unc.lib.dl.fedora.FedoraException;
+import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.fedora.types.MIMETypedStream;
+import edu.unc.lib.dl.search.solr.model.BriefObjectMetadata;
+import edu.unc.lib.dl.search.solr.model.SearchRequest;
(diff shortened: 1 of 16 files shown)


daines:
What do we do about Outlook web blocking XML file attachments?

bbpennel:
If we view it as a major problem, we could  1) change extensions, 2) put it
in a zip, 3) make the body of the email be the xml document or 4) switch to
providing a url instead

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
```diff
@@ -0,0 +1,300 @@
...
+			this.user = user;
+			this.groups = groups;
+			this.request = request;
+		}
+
+		@Override
+		public void run() {
+			long startTime = System.currentTimeMillis();
+			String seperator = System.getProperty("line.separator");
+			byte[] seperatorBytes = System.getProperty("line.separator").getBytes();
```
daines: exportHeaderBytes doesn't use this property. Is it important to use it?

bbpennel: it should be consistent, i'll fix that.

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
```diff
@@ -0,0 +1,300 @@
...
+							objectDoc.addContent(objectEl);
+
+							DatastreamDocument modsDS = managementClient
+									.getXMLDatastreamIfExists(pid, Datastream.MD_DESCRIPTIVE.getName());
+
+							if (modsDS != null) {
+								Element modsUpdateEl = new Element("update");
+								modsUpdateEl.setAttribute("type", "MODS");
+								modsUpdateEl.setAttribute("lastModified", modsDS.getLastModified());
+								modsUpdateEl.addContent(modsDS.getDocument().detachRootElement());
```
daines: Can you guarantee that these MODS elements will use the "mods" namespace prefix?

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
```diff
@@ -0,0 +1,300 @@
...
+								objectEl.addContent(modsUpdateEl);
+							}
+
+							xmlOutput.output(objectEl, xfop);
+
+							xfop.write(seperatorBytes);
+							xfop.flush();
```
daines: Why is flush() necessary here?

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
```diff
@@ -0,0 +1,300 @@
...
+				try (FileOutputStream xfop = new FileOutputStream(mdExportFile)) {
+					xfop.write(exportHeaderBytes);
+
+					Format format = Format.getPrettyFormat();
+					format.setLineSeparator(seperator);
+					XMLOutputter xmlOutput = new XMLOutputter(Format.getPrettyFormat());
+					for (String pidString : request.getPids()) {
+						PID pid = new PID(pidString);
+
+						if (!aclService.hasAccess(pid, groups, Permission.editDescription)) {
```
daines: I assume we are checking this here because it would slow down the controller method to check for every pid? We might end up with a lot of INFO statements.

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
```diff
@@ -0,0 +1,300 @@
...
+		public XMLExportRunnable(XMLExportRequest request, String user, AccessGroupSet groups) {
+			this.user = user;
+			this.groups = groups;
+			this.request = request;
+		}
+
+		@Override
+		public void run() {
+			long startTime = System.currentTimeMillis();
+			String seperator = System.getProperty("line.separator");
```
daines: Fix spelling of "separator"

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
```diff
@@ -0,0 +1,300 @@
...
+			byte[] seperatorBytes = System.getProperty("line.separator").getBytes();
+
+			try {
+				GroupsThreadStore.storeGroups(groups);
+				GroupsThreadStore.storeUsername(user);
+
+				File mdExportFile = File.createTempFile("md_export", ".xml");
```
daines: It might be preferable to name this "xml_export" or something else similar to the controller class name.

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
bbpennel: I was concerned about that too, but it actually doesn't matter and I could probably remove this.  The MODS documents will have the MODS namespace declared in them with whatever prefix they need, since they are all standalone documents.

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
bbpennel: Its probably not necessary, I just prefer to have it there since you will be able to see progress when debugging

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportXMLController.java
bbpennel: Perhaps I should turn it down to debug?  And yeah, its in here so that the request doesn't have to wait on it.

daines:
When multiple objects are selected and the context menu is activated, the menu item for this reads, eg, "Export MODS 2 objects". It should read "Export MODS **for** 2 objects", or just "Export MODS". This may be a separate issue, though, because there is also a "Edit Type 2 objects" menu item.

PR #327, opened 2015-07-13
Wildcard searching

bbpennel:
Not particularly elegant, but it seemed better to duplicate the solr parameter cleaning method minus the * character versus unescaping the character.


daines:
We should probably add a unit test for this to SolrSettingsTest.

PR #333, opened 2015-07-29
Display deposit monitor results sorted by submitTime.

daines:


bbpennel:
Thanks for doing this mike, it is long overdue and makes the monitor vastly more usable.

PR #350, opened 2015-08-12
Bulk Metadata Importing

bbpennel:
Adds option for importing bulk metadata update files, as exported by the XML export feature.  The option is currently listed in the "+Add" menu


services-worker/src/main/java/edu/unc/lib/dl/cdr/services/worker/WorkerDaemon.java
```diff
@@ -38,18 +40,22 @@ public void start() throws Exception {
...
 		}
 		
-		WorkerPool workerPool = appContext.getBean(WorkerPool.class);
-		workerPool.getWorkerEventEmitter().addListener(this);
-		workerPool.run();
+		Map<String, WorkerPool> workerPools = appContext.getBeansOfType(WorkerPool.class);
+		for (WorkerPool workerPool : workerPools.values()) {
+			workerPool.getWorkerEventEmitter().addListener(this);
+		}
+		workerPools.values().iterator().next().run();
```
bbpennel: This is weird, but calling run() on each workerPool resulted in it calling start on threads which had already started in the second worker pool.  There are no messages for workers in the second pool starting, but they still do work

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/worker/WorkerDaemon.java
```diff
@@ -38,18 +40,22 @@ public void start() throws Exception {
...
 	public void stop() throws Exception {
 		LOG.info("Stopping the services worker daemon");
 		
-		WorkerPool workerPool = appContext.getBean(WorkerPool.class);
-		workerPool.end(true);
-		appContext.stop();
+		Map<String, WorkerPool> workerPools = appContext.getBeansOfType(WorkerPool.class);
+		for (WorkerPool workerPool : workerPools.values()) {
+			workerPool.end(true);
+			appContext.stop();
```
daines: Move this line outside the loop?

PR #351, opened 2015-08-13
Prevent failure on duplicate pid

bbpennel:
Attempt to determine if an existing file is okay during ingest, including when the object is either from the same deposit or is the same size

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/IngestDeposit.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/IngestDeposit.java
@@ -384,7 +384,7 @@ public class IngestDeposit extends AbstractDepositJob implements ListenerJob {
 			log.info("Fedora ingest timed out, awaiting ingest confirmation and proceeding with the remainder of the deposit: "
 					+ e.getLocalizedMessage());
 		} catch (ObjectExistsException e) {
-			if (confirmExisting || isDuplicateOkay(pid, ingestPid)) {
+			if (confirmExisting || isDuplicateOkay(pid)) {
 				ingestsAwaitingConfirmation.remove(ingestPid);
 			} else {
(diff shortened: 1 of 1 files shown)


deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/IngestDeposit.java
```diff
@@ -382,6 +397,58 @@ private void ingestObject(String ingestPid, boolean confirmExisting) throws Depo
 		// TODO increment ingestedOctets
 
 	}
+	
+	private boolean isDuplicateOkay(PID pid, String depositId) {
```
daines: Where is depositId used in this method?

deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/IngestDeposit.java
```diff
@@ -382,6 +397,58 @@ private void ingestObject(String ingestPid, boolean confirmExisting) throws Depo
...
+		List<String> deposits = tsqs.fetchBySubjectAndPredicate(pid, Relationship.originalDeposit.toString());
+		
+		// Ensure that either the deposit ID of the file in the repository matches the current deposit
+		// or that there is no deposit id assigned, such as is the case for small ingests
+		if (deposits == null || deposits.size() == 0 || deposits.contains(this.getDepositPID().getURI())) {
```
daines: Could we use something like this logic to avoid having too many levels of scope?

```
if (there is a deposit record associated with this pid in the repository and it isn't the current deposit) {

  return false, because that means this pid definitely belongs to another deposit

}

if (there is not a staged file for this object) {

  return true
  
} else {

  if (we have a size for this object and the size of the staged file doesn't match) {
    return false
  }

  if (we have a checksum for this object and the checksum checksum of the staged file doesn't match) {
    return false
  }

  return true

}
```

PR #352, opened 2015-08-14
Prevent duplicate deposit jobs from enqueuing when resuming

bbpennel:
Verifying that jobs are not already queued for a deposit before resuming or re-enqueuing during service startup to prevent duplicate jobs being created

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
@@ -664,16 +664,18 @@ public class DepositSupervisor implements WorkerListener {
 			jobStatusFactory.clearStale(uuid);
 			depositStatusFactory.deleteField(uuid, DepositField.errorMessage);
 			
+			// since we already checked for queued jobs at startup, only check when resuming from a paused state
 			boolean enqueueNext = true;
 			if (DepositState.paused.name().equals(status.get(DepositField.state.name()))) {
 				Map<String, Set<String>> depositSet = getQueuedDepositsWithJobs();
 				enqueueNext = !depositSet.containsKey(uuid);
-				LOG.info("Resuming from paused state.  {} will enqueue a new job {}", uuid, enqueueNext);
 			}
 
 			if (enqueueNext) {
 				List<String> successfulJobs = jobStatusFactory.getSuccessfulJobNames(uuid);
 				queueNextJob(null, uuid, status, successfulJobs, delay);
+			} else {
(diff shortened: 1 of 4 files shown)


deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
```diff
@@ -225,15 +239,53 @@ public void run() {
...
+		addQueuedDeposits(
+				RedisWorkerConstants.RESQUE_QUEUE_PREFIX + RedisWorkerConstants.DEPOSIT_PREPARE_QUEUE, depositMap);
+		addQueuedDeposits(
+				RedisWorkerConstants.RESQUE_QUEUE_PREFIX + RedisWorkerConstants.DEPOSIT_DELAYED_QUEUE, depositMap);
+		addQueuedDeposits(
+				RedisWorkerConstants.RESQUE_QUEUE_PREFIX + RedisWorkerConstants.DEPOSIT_CDRMETS_QUEUE, depositMap);
+		return depositMap;
+	}
+	
+	private void addQueuedDeposits(String queueName, Map<String, Set<String>> depositMap) {
```
daines: Is there a way to do this using Jesque's API rather than going to Redis?

bbpennel: Nothing in the Client class, but it looks like jesque web is polling jobs using this:

https://github.com/gresrun/jesque/blob/master/src/main/java/net/greghaines/jesque/meta/dao/impl/QueueInfoDAORedisImpl.java#L152

daines: I don't know about peeking at the jobs in the queue, but we could use JedisUtils.isDelayedQueue for the first part of this method.

deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
```diff
@@ -607,9 +663,18 @@ private void resumeDeposit(String uuid, Map<String, String> status, long delay)
 			// Clear out the previous failed job if there was one
 			jobStatusFactory.clearStale(uuid);
 			depositStatusFactory.deleteField(uuid, DepositField.errorMessage);
+			
+			boolean enqueueNext = true;
+			if (DepositState.paused.name().equals(status.get(DepositField.state.name()))) {
+				Map<String, Set<String>> depositSet = getQueuedDepositsWithJobs();
+				enqueueNext = !depositSet.containsKey(uuid);
+				LOG.info("Resuming from paused state.  {} will enqueue a new job {}", uuid, enqueueNext);
```
daines: This log statement seems like it should go in the if block below?

PR #354, opened 2015-08-18
Fixes for deposit emails related to vocab enforcement failures

bbpennel:
* Wait for the repository to become available before initializing the departments vocabulary to prevent VocabularyEnforcementJob from failing due to an NPE
* Store a failure message for non-JobFailureExceptions so that the mail sender will realize the job failed
* Check that the ingested count variable is set in deposit email template so that it won't throw an exception when the deposit fails before the ingest job

### DIFF
--- a/fcrepo-clients/src/main/java/edu/unc/lib/dl/util/VocabularyHelperManager.java
+++ b/fcrepo-clients/src/main/java/edu/unc/lib/dl/util/VocabularyHelperManager.java
@@ -86,6 +86,9 @@ public class VocabularyHelperManager {
 	private Boolean initialized = false;
 
 	public synchronized void init() {
+		// Wait for the repository to be up before loading vocabularies
+		managementClient.waitForRepositoryAvailable();
+		
 		log.debug("Initializing vocabulary helpers");
 		initialized = true;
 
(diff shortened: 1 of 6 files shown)


fcrepo-clients/src/main/java/edu/unc/lib/dl/util/VocabularyHelperManager.java
```diff
@@ -86,6 +86,9 @@
 	private Boolean initialized = false;
 
 	public synchronized void init() {
+		// Wait for the repository to be up before loading vocabularies
+		managementClient.waitForRepositoryAvailable();
```
daines: Since other parts of the deposit service depend on the availability of the repository, could this be moved into DepositSupervisor?

bbpennel: It is probably needed in any code that will not be running in tomcat, such
as the services workers (in fact, it is needed in there), so I think it
makes sense to have the check for repository availability in
fcrepo-clients.  But having something that blocked startup of the deposit
service until fedora came online would actually be wise.  I can't recall if
the attempt to get the Collections object pid fails if fedora is offline,
it may be going directly to mulgara and therefore not care.

PR #368, opened 2015-08-25
Previews always visible

daines:


daines:
This doesn't take into account the "View" button on PDFs.

PR #371, opened 2015-08-26
Relationship updating cleanup

bbpennel:
* Created RDFXMLUtil to perform some of the basic RELS-EXT manipulation we frequently perform.
* Switched all relationship adding methods over to using optimstic locking.
* Removed relationship modifying methods from DigitalObjectManager, moved RELS-EXT updating method to ManagementClient.
* Normalized the parameters for relationship updating methods.
* Updated enhancements to use local client instead of referencing their services instance repeatedly, removed relationship setting methods, and switched to setting exclusive triples in a few cases where it makes more sense
* Removed some dead code and tests related to relationships

### DIFF
--- a/fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
+++ b/fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
@@ -17,6 +17,7 @@ package edu.unc.lib.dl.fedora;
 
 import static edu.unc.lib.dl.util.ContentModelHelper.Administrative_PID.REPOSITORY;
 import static edu.unc.lib.dl.util.ContentModelHelper.Datastream.MD_EVENTS;
+import static edu.unc.lib.dl.util.ContentModelHelper.Datastream.RELS_EXT;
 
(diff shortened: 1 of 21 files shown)


fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
```diff
@@ -258,16 +258,28 @@ public String addInlineXMLDatastream(PID pid, String dsid, boolean force, String
...
-		req.setObject(literal);
-		AddRelationshipResponse resp = (AddRelationshipResponse) this.callService(req, Action.addRelationship);
-		return resp.isAdded();
+		addTriple(pid, pred, ns, true, literal, datatype);
+	}
+	
+	public void addTriple(PID pid, String pred, Namespace ns, boolean isLiteral, String value, String datatype)
+			throws FedoraException {
+		
+		DatastreamDocument dsDoc = getRELSEXTWithRetries(pid);
```
daines: Shouldn't this be inside the loop? If we get an OptimisticLockException, when do we retrieve a new version of the document to work with?

PR #374, opened 2015-09-02
Action for manually triggering enhancements

bbpennel:


static/templates/admin/runEnhancementsForm.html
```diff
@@ -0,0 +1,19 @@
+<form id="run_enhancements_form"class="edit_form">
+	<p>
+		All appropriate enhancements will be performed on the selected objects and their children
```
daines: This reads a little strangely to me. What is an "appropriate" enhancement? Also, if this meant as a full sentence, it needs to end with a period.

(We could probably just drop this label.)

static/templates/admin/runEnhancementsForm.html
```diff
@@ -0,0 +1,19 @@
...
+	</p>
+	
+	<span>
+		<input type="checkbox" id="run_enhancements_force" name="run_enhancements_force" /><label for="run_enhancements_force">Force enhancement of objects which have been previously processed?</label>
```
daines: Instead of "force enhancements", it might be clearer to say something like "run enhancements again" or "re-run enhancements".

bbpennel: I agree that's clearer, but the only problem is that it will also run enhancements for the first time if they had not run previously

static/templates/admin/runEnhancementsForm.html
bbpennel: If we did so, we would likely want to change the title of the window to say that the children of the object would also get enhancements run.  The window looks slightly sad with nothing in it but the checkbox :(

static/templates/admin/runEnhancementsForm.html
bbpennel: Perhaps, "Re-run previous enhancements too" to indicate that it's bonus.

admin/src/main/java/edu/unc/lib/dl/admin/controller/RunEnhancementsController.java
```diff
@@ -0,0 +1,108 @@
...
+				result.put("error", "Insufficient permissions to perform operation");
+				return result;
+			}
+		}
+		
+		
+		Job job = new Job(RunEnhancementsTreeJob.class.getName(), data.getPids(), data.isForce());
+		jesqueClient.enqueue(runEnhancementQueueName, job);
+		
+		if (data.isForce()) {
```
daines: I don't think it's necessary to differentiate these cases in the result message -- just "enhancement ... has begun" is probably sufficient.

PR #383, opened 2015-09-21
Metadata editor

bbpennel:


fcrepo-clients/src/main/java/edu/unc/lib/dl/util/TripleStoreQueryServiceMulgaraImpl.java
```diff
@@ -1180,24 +1180,23 @@ public String lookupSourceMimeType(PID pid) {
 		Set<String[]> result = new HashSet<String[]>();
 		StringBuffer query = new StringBuffer();
 		query.append("select $group $role from <%1$s>")
-				.append(" where <%2$s> $role $group").append(" and (")
```
daines: I kind of like the old way because the entire query was visible, but looping over the enum values is better. Could you use String.format within the loop rather than at the end?

```java
query.append(String.format("select $group $role from <%s> ", this.getResourceIndexModelUri()))
query.append(String.format("where <%s> $role $group and (", pid.getURI()))

int i = 0;
for (UserRole role : UserRole.values()) {
  query.append(String.format("$role <mulgara:is> <%s>", role.getURI()));
  if (i > 0) {
    query.append(" or ");
  }
  i++;
}

query.append(")");
```

PR #384, opened 2015-09-24
Remove Capistrano-related files

daines:


bbpennel:
Do we want to merge this before the fabric approach is added?

PR #386, opened 2015-09-29
Track throughput metrics

bbpennel:


daines:
Maybe we should centralize metrics stuff in OperationMetricsClient? (and it could be given a different name)

PR #396, opened 2015-10-16
Increased solr timeouts

bbpennel:


bbpennel:
These are not scientifically chosen values, but I don't think they are unreasonable either.  Its very easy to trigger a timeout at the moment with blank searches.

PR #399, opened 2015-10-23
WordPress RSS Feed for Case Study

lfarrell:


access/src/main/webapp/WEB-INF/jsp/frontPage.jsp
```diff
@@ -131,41 +131,11 @@
...
-
-			<p>With one of the nation’s finest collections of archaeological materials from the South, the RLA curates more than seven million artifacts along with more than 60,000 photographic negatives, photographs, and slides. Over the past 70 years, virtually all of the major discoveries in the understanding of North Carolina's ancient past can be attributed to the RLA or to researchers trained there.</p>
-
-			<p>The Carolina Digital Repository is working with RLA staff to to preserve their large collection of photographs, slides, and publications, and to provide access to those collections online.</p>
-
-			<p class="button"><a href="record/uuid:8ae56bbc-400e-496d-af4b-3c585e20dba1">Browse the RLA Collection</a></p>
+	<div id="front-case-study" class="row">		
+		<div class="case-study">
+			<h2><i>Case Study:</i> ${ wpRssItem.title }</h2>
+			${ wpRssItem.encoded }
```
daines: Rather than putting the button at the bottom of the blog post, so that it goes in the RSS `<content>` element, could we modify the `<link>` element to be the URL of the collection and have the "Browse the collection" button use that link?

PR #410, opened 2015-11-19
Bagit source ingest

bbpennel:


admin/src/main/java/edu/unc/lib/dl/admin/collect/IngestSourceManager.java
```diff
@@ -0,0 +1,303 @@
...
+		candidate.put("sourceId", source.getId());
+		candidate.put("base", base);
+		candidate.put("patternMatched", Paths.get(base).relativize(filePath).toString());
+		
+		candidate.put("version", version);
+
+		if (version != null) {
+			// Add payload stats for bags
+			addBagInfo(candidate, filePath);
+		}/* else if (file.isDirectory()) {
```
daines: Are we supporting these other types? If not, addDirectoryStats is dead code, right?

bbpennel: Yes, its dead code currently, but we have a ticket for adding deposit from plain directories so I figured we might as well keep it until we get to that, although perhaps it'd be better to move it to a branch related to that purpose

admin/src/main/java/edu/unc/lib/dl/admin/controller/IngestSourceController.java
```diff
@@ -0,0 +1,188 @@
...
+			
+			Map<String, String> deposit = new HashMap<>();
+			deposit.put(DepositField.sourcePath.name(), source.getBase() + packageDetails.getPackagePath());
+			deposit.put(DepositField.fileName.name(), packageDetails.getLabel());
+			deposit.put(DepositField.packagingType.name(), packageDetails.getPackagingType());
+			deposit.put(DepositField.uuid.name(), depositPID.getUUID());
+			deposit.put(DepositField.submitTime.name(), String.valueOf(System.currentTimeMillis()));
+			deposit.put(DepositField.depositorName.name(), GroupsThreadStore.getUsername());
+			deposit.put(DepositField.permissionGroups.name(), GroupsThreadStore.getGroupString());
+			deposit.put(DepositField.depositorEmail.name(), GroupsThreadStore.getUsername() + "@email.unc.edu");
```
daines: We get email addresses in two ways across the CDR codebase: by appending "@email.unc.edu" to the onyen and by looking at the "mail" request header. Can we settle on one of these?

deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
```diff
@@ -0,0 +1,222 @@
...
+		UUID uuid = UUID.randomUUID();
+		PID pid = new PID("uuid:" + uuid.toString());
+
+		Resource fileResource = top.getModel().createResource(pid.getURI());
+		folderBag.add(fileResource);
+
+		return fileResource;
+	}
+	
+	private com.hp.hpl.jena.rdf.model.Bag getFolderBag(com.hp.hpl.jena.rdf.model.Bag top, String basepath, String filepath) {
```
daines: No more recursion? :(

deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
```diff
@@ -0,0 +1,222 @@
...
+		segmentLoop: for (int i = 1; i < pathSegments.length - 1; i++) {
+			String segment = pathSegments[i];
+			
+			// Search to see if a folder with the same name as this segment exists as a child
+			NodeIterator nodeIt = currentNode.iterator();
+			while (nodeIt.hasNext()) {
+				Resource child = nodeIt.nextNode().asResource();
+				
+				String label = child.getProperty(labelProp).getString();
+				if (label.equals(segment) && child.hasProperty(hasModelProp, containerResource)) {
```
daines: Why are you checking that the child is a container?

static/js/admin/src/IngestFromSourceForm.js
```diff
@@ -0,0 +1,205 @@
...
+			var $input = $(this);
+			var value = $input.val();
+			
+			// Compute a regular expression for the user's search terms
+			var matchRe = null;
+			if (value && value.trim()) {
+				var pattern = "";
+				var terms = value.trim().split(" ");
+				for (var i = 0; i < terms.length; i++) {
+					var escaped = terms[i].replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&");
```
daines: This regexp could use a comment... It looks symbols are taken out of space-separated terms and terms are matched in order...?

What's the rationale for this particular type of matching?

static/templates/admin/ingestSourceCandidates.html
```diff
@@ -0,0 +1,45 @@
...
+					<i class="fa fa-suitcase"></i>
+				<% } else { %>
+					<i class="fa fa-folder"></i>
+				<% } %>
+			</div>
+			
+			<div class="file_browse_entry_data">
+				<h2><%= candidate.patternMatched %></h2>
+				<p><%= candidate.type %> <%= candidate.version? candidate.version : "" %>
+					- <%= candidate.files %> item<%= candidate.files == 1? "" : "s" %>
```
daines: I'd prefer to use &ndash; or &mdash; rather than a hyphen here.

static/js/admin/src/IngestFromSourceForm.js
bbpennel: The regex is a standard replacement regex for escaping regex characters.  The matching is just (*term1*|*term2*) so that users could search for different parts of the path if they can't remember the whole thing.

deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
bbpennel: Yeah, that made me slightly sad, but when I was merging our two different versions (lets avoid having to do that in the future) I mostly went with the version I already knew if they looked to do basically the same thing.

deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
bbpennel: You're right, that's probably not necessary

static/js/admin/src/IngestFromSourceForm.js
daines: Given a list of space-separated terms t1, t2, ... doesn't the final pattern look like this?

```
".*" + escape(t1) + ".*" + escape(t2) + ".*" + ... ".*"
```

So if the filter is "b a", it won't match the path "apple/butter"?

Couldn't we just split the filter on spaces and use indexOf on those terms?

admin/src/main/java/edu/unc/lib/dl/admin/controller/IngestSourceController.java
bbpennel: I'll resolve this after we've merged https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/412

PR #412, opened 2015-11-20
Email cleanup

bbpennel:
Using the same method of getting email addresses everywhere and fixing a template bug related to deposit collectors.


sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/deposit/AbstractDepositHandler.java
```diff
@@ -113,8 +113,9 @@ protected void registerDeposit(PID depositPid, PID destination, Deposit deposit,
 		status.put(DepositField.submitTime.name(), String.valueOf(System.currentTimeMillis()));
 		status.put(DepositField.fileName.name(), deposit.getFilename());
 		status.put(DepositField.fileMimetype.name(), deposit.getMimeType());
-		String email = SwordConfigurationImpl.getUserEmailAddress();
+		String email = GroupsThreadStore.getEmail();
 		status.put(DepositField.depositorName.name(), owner);
+		// Fall back to a generated email address using the owner if there was no email address on the request
 		status.put(DepositField.depositorEmail.name(), email != null ? email : owner+"@email.unc.edu");
```
daines: When do we actually need to construct the email this way? From what I can tell, `On-Behalf-Of` and therefore `owner` is set when proxying requests via RESTProxyController. It doesn't appear to be set from the forms app.

We could pass on the `mail` header in RESTProxyController when we pass along the Shibboleth groups.

PR #433, opened 2015-12-08
Bulk ingest files from a directory

lfarrell:


bbpennel:
It looks like IngestFromSourceForm.js was based off an older version, so you'llneed to update to use the version from master with your changes copied over.  Looks like it should be pretty easy to put them together though.  That should fix the problems with selection, which was busted for me.

I'd also still like there to either be an AbstractFileServerToBagJob or some other abstraction with the methods you moved up to AbstractDepositJob in it.

Otherwise this seems to be working well!

PR #434, opened 2015-12-08
Full record page navigation

bbpennel:
* Removed the previous and next buttons (which weren't working) from the full record page
* Fixed the ordering of the related items list


access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -338,152 +340,63 @@ public SearchResultResponse getFacetList(SearchRequest searchRequest) {
...
-			solrQuery.setSort(solrSettings.getFieldName(SearchFieldKeys.DISPLAY_ORDER.name()), SolrQuery.ORDER.desc);
-
+		
+		solrQuery.setQuery("*:*" + accessRestrictionClause);
+		
+		addSort(solrQuery, "default", true);
+		solrQuery.setRows(NEIGHBOR_SEEK_PAGE_SIZE);
+		solrQuery.setFields("id");
+		
+		long total = -1;
```
daines: This isn't super complicated, but a short note about what's going on here would be helpful. eg, "Query for ids in this container in groups of NEIGHBOR_SEEK_PAGE_SIZE until we find the offset of the object. Then query for windowSize objects with the start adjusted so that the object is in the middle."

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -338,152 +340,63 @@ public SearchResultResponse getFacetList(SearchRequest searchRequest) {
...
-				LOG.debug("Picking up slack from left list");
-
-				int extra = halfWindow - rightList.size();
-
-				if (halfWindow + extra < leftList.size())
-					leftList.subList(halfWindow + extra, leftList.size()).clear();
-
+		} while (start < total);
+		
+		// Wasn't found, no neighbors shall be forthcoming
```
daines: When will this happen?

daines:
Ideally getNeighboringItems should have a note about its complexity in its documentation.

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
bbpennel: It shouldn't, but could happen in unlikely cases such as an item having been moved, deleted, hidden during the request, or some other general error state.

PR #441, opened 2016-01-06
Remove mp4 video player. Upgrading to latest version doesn't fix video not playing in Chrome

lfarrell:


lfarrell:
Pulling in whole files that aren't really changed. Closing & opening new pull request

PR #450, opened 2016-02-12
If ingested file comes in with a mime-type of (x)-download favor the fits mime-type over the fedora mime-type.

lfarrell:


access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
```diff
@@ -176,9 +176,11 @@ private void streamData(String simplepid, Datastream datastream, String filename
 					String mimeType;
 					try {
 						mimeType = method.getResponseHeader("content-type").getValue();
-						if (mimeType == null || "application/octet-stream".equals(mimeType)) {
+						if (mimeType == null || "application/octet-stream".equals(mimeType) || "application/download".equals(mimeType) || "application/x-download".equals(mimeType)) {
```
bbpennel: Did you find that the overrides in this class were necessary with the changes down in TechnicalMetadataEnhancement in place?

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
```diff
@@ -176,9 +176,11 @@ private void streamData(String simplepid, Datastream datastream, String filename
...
 							if ("mp3".equals(datastream.getExtension())) {
 								mimeType = "audio/mpeg";
+							} else if ("pdf".equals(datastream.getExtension())) {
```
lfarrell: I'm not sure it's strictly necessary, but pdfs with taht mime-type will never hit this line if it's not there.

bbpennel: Yes, but it sounds like our intention is to correct all the existing records with "download" mimetypes, and prevent the mimetype from showing up in the future, so I am thinking we may not need this mimetype overriding at streaming time.  The same may be true for this file extension bit as well.  What do you think?

lfarrell:
Sounds reasonable. I'll go ahead and remove that bit.

PR #452, opened 2016-02-22
Add record uri to api output. Base uri set as a server property.

lfarrell:


access/src/main/webapp/WEB-INF/access-fedora-context.xml
```diff
@@ -96,6 +96,7 @@
 		<property name="solrPath"
 			value="${solr.protocol}://${solr.host}${solr.port}/${solr.context}" />
 		<property name="djatokaPath" value="${djatoka.base.url}" />
+		<property name="apiPath" value="${repository.apiUrl}" />
```
daines: This could use a more specific name. I assume apiPath is supposed to be something like "https://cdr.lib.unc.edu/record/"?

Also, do we want this defined under the "repository" namespace, or something else, like "services"?

lfarrell:
Yes, the apiPath would need something like this https://cdr.lib.unc.edu/record/. I'm open to putting it in wherever it seems reasonable.

PR #454, opened 2016-03-11
Directory index from file server fix 

lfarrell:
Adds nested files, checksums and correctly includes file data.


deposit/src/main/java/edu/unc/lib/deposit/normalize/DirectoryToBagJob.java
```diff
@@ -77,16 +92,29 @@ public void runJob() {
...
 		// Add all of the payload objects into the bag folder
 		for (File file : listOfFiles) {
-			String filePath = file.getName();
+			String fullPath = file.toString();
+			String checksum = null;
+			
+			try {
+				checksum = DigestUtils.md5Hex(new FileInputStream(fullPath));
+			} catch (FileNotFoundException e) {
+				log.debug("Unable to compute checksum {}", checksum, e);
```
bbpennel: Failing to compute the checksum should fail the ingest, since it probably means the file wasn't accessible.  For the failure you'll want to include fullPath rather than checksum, since the latter will be null.  You can also combine the two catches into one if they have the same result, by doing (FileNotFoundException | IOException e).

deposit/src/main/java/edu/unc/lib/deposit/normalize/DirectoryToBagJob.java
```diff
@@ -43,6 +55,8 @@
  * @author lfarrell
  */
 public class DirectoryToBagJob extends AbstractFileServerToBagJob {
+	private static final Logger log = LoggerFactory.getLogger(IngestDeposit.class);
```
bbpennel: classname should match the name of the class you are logging for (DirectoryToBagJob.class), otherwise the logging statements will be listed as coming from the wrong class.

PR #457, opened 2016-03-23
Include empty folders in directory ingests

lfarrell:


bbpennel:
This seems reasonable.  I am going to take a look at changing the implementation of getFolderBag to address memory usage, but do so in another PR

PR #462, opened 2016-04-05
Add new performance metrics

lfarrell:


deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
```diff
@@ -613,6 +611,18 @@ private void queueNextJob(Job job, String depositUUID, Map<String, String> statu
 			depositStatusFactory.setState(depositUUID, DepositState.finished);
 			metricsClient.incrFinishedDeposit();
 			
+			Map<String, String> deposit = depositStatusFactory.get(depositUUID);
```
bbpennel: I believe you don't need to retrieve the deposit status again here as it is already being passed into this method in the "status" parameter

deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
```diff
@@ -613,6 +611,18 @@ private void queueNextJob(Job job, String depositUUID, Map<String, String> statu
...
+			String strDepositStartTime = deposit.get(DepositField.startTime.name());
+			Long depositStartTime = Long.parseLong(strDepositStartTime, 10);
```
bbpennel: The "10" radix parameter isn't needed

metadata/src/main/java/edu/unc/lib/dl/reporting/ActivityMetricsClient.java
```diff
@@ -86,11 +86,23 @@ public void incrFinishedDeposit() {
 		}
 	}
 	
-	public void incrDepositFileThroughput(long bytes) {
+	public void incrDepositFileThroughput(String uuid, long bytes) {
 		try (Jedis jedis = getJedisPool().getResource()) {
 			String date = metricsDateFormat.format(new Date());
 			jedis.hincrBy(DEPOSIT_METRICS_PREFIX + date, "throughput-files", 1);
```
bbpennel: How do you feel about continuing to track the cumulative throughput for the day when we are also recording per deposit/file statistics?  With the new setup its a nice side effect that ingests that run overnight will end up with metrics on the correct day rather than the whole deposit going onto whatever day it finishes on.  So if it is just as useful for your visualization purposes, I don't see a reason that we would need to keep the cumulative as well.

lfarrell:
Weird merging issues from code I didn't touch. Will open new request.

PR #466, opened 2016-05-06
CDR Metrics Visualizations

lfarrell:


access/src/main/webapp/WEB-INF/jsp/fullRecord/fileRecord.jsp
```diff
@@ -84,7 +84,7 @@
 			<c:choose>
 				<c:when test="${cdr:permitDatastreamAccess(requestScope.accessGroupSet, 'IMAGE_JP2000', briefObject)}">
 					<div class="clear_space"></div>
-					<div id="jp2_viewer" class="jp2_imageviewer_window djatokalayers_window" data-url='${briefObject.id}'></div>
```
bbpennel: Not sure why this change is showing up here, master doesn't have the djatoka class anymore

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+	public Set<String> getDepositMetrics() {
+		try (Jedis jedis = getJedisPool().getResource()){
+			return jedis.keys("deposit-metrics:*");
+		}
+	}
+	
+	private Boolean buildFile(String path) {
+		File csvFile = new File(path);
+		
+		if(csvFile.exists() && !csvFile.isDirectory()) {
```
bbpennel: space between if and (

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+	/**
+	 * Gets totals for metrics that are only aggregated daily such as moves & enhancements
+	 * It also grabs throughput totals for older data, before these fields were set to measure performance at the individual deposit level
+	 * @return
+	 */
+	public String getOperationsData() {
+		FileWriter fileWriter = null;
+		CSVPrinter csvFilePrinter = null;
+		Set<String> operations = null;
+		String filePath = "/opt/data/ingest-times-daily.csv";
```
bbpennel: Avoid hard coding the path, I'd suggest switching over to injecting the data.dir server property and using that

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+				csvFilePrinter = new CSVPrinter(fileWriter, csvFileFormat);
+				csvFilePrinter.printRecord(FILE_HEADERS);
+
+				Boolean matchingDate = false;
+				for (String deposit : deposits) {
+					depositJob = jedis.hgetAll(deposit);
+					depositKeys = deposit.split(":");
+					
+					// Ignore data for individual deposits by uuid. Only need the daily ones in this instance
+					if (depositKeys.length > 2) {
```
bbpennel: Make this check before the hgetall on line 124 so that it doesn't have to retrieve the data that its not going to use

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+	}
+	
+	private Boolean buildFile(String path) {
+		File csvFile = new File(path);
+		
+		if(csvFile.exists() && !csvFile.isDirectory()) { 
+			Long currentTime = System.currentTimeMillis();
+			Long fileCreationTime = csvFile.lastModified();
+			
+			if ((currentTime - fileCreationTime ) > 3600000) {
```
bbpennel: I'd suggest making this value a constant at the top of the file, and it can be helpful to set it to 60 * 60 * 1000 to make it a little clearer how long the period is

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+		FileWriter fileWriter = null;
+		CSVPrinter csvFilePrinter = null;
+		Set<String> operations = null;
+		String filePath = "/opt/data/ingest-times-daily.csv";
+		Map<String, String> depositJob = null;
+		Map<String, String> operationJob = null;
+		String[] depositKeys = null;
+
+		if (buildFile(filePath)) {
+			Set<String> deposits = getDepositMetrics();
```
bbpennel: There is a lot of looping over deposit metrics in here considering this is the OperationMetrics method.  Why is this needed here?  Is it somehow related to showing the deposit stats when hovering in the operations graph?

access/src/main/webapp/WEB-INF/jsp/fullRecord/fileRecord.jsp
lfarrell: Yeah, that one shouldn't be in there.

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+					if (depositKeys.length > 2) {
+						continue;
+					}
+
+					String jobDate = depositKeys[1];
+					String throughputFiles = depositJob.get("throughput-files");
+					String throughputBytes = depositJob.get("throughput-bytes");
+					String finished = depositJob.get("finished");
+					String failed = depositJob.get("failed");
+					String failedDepositJob = depositJob.get("failed-job:edu.unc.lib.dl.cdr.services.techmd.TechnicalMetadataEnhancementService");
```
bbpennel: The depositJob shouldn't have anything about the TechnicalMetadataEnhancementService in it, that is an enhancement.  I'm guessing you don't actually need to try to pull this here since you are populating enhancement stats here, and judging by ActivityMetricsClient, it doesn't look like we are recording the failed job per deposit currently (but probably should), just per day

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+
+					for (String operation : operations) {
```
bbpennel: I might be misunderstanding, but does this need to be nested inside a loop over deposit jobs?  It seems to be independent.

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+						String operationDate = operation.split(":")[1];
+						
+						if (operationDate.equals(jobDate)) {
+							operationJob = jedis.hgetAll(operation);
+
+							String moves = (operationJob.get("moves"));
+							String finishedEnhancements = operationJob.get("finished-enh:edu.unc.lib.dl.cdr.services.techmd.TechnicalMetadataEnhancementService");
```
bbpennel: TechMD is one of four or five jobs that should be getting tracked

admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,299 @@
...
+		Map<String, String> depositJob = null;
+		Map<String, String> operationJob = null;
+		String[] depositKeys = null;
+
+		if (buildFile(filePath)) {
+			Set<String> deposits = getDepositMetrics();
+			try (Jedis jedis = getJedisPool().getResource()) {
+				operations = jedis.keys("operation-metrics:*");
+
+				fileWriter = new FileWriter(filePath);
```
bbpennel: Generally its good to use the try() to create and autoclose resources when you can.  In this case, you could put all your writers and the jedis into the try together (and maybe end up with a new star wars movie), and not worry about closing them later, like this:
```java
try (
	Jedis jedis = getJedisPool().getResource();
	FileWriter fileWriter = new FileWriter(filePath);
	CSVPrinter csvFilePrinter = new CSVPrinter(fileWriter, csvFileFormat);
) { ```

File: admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java [hunk 11]
```diff
@@ -0,0 +1,299 @@
...
+						data.add("0");
+						data.add(failed);
+						data.add(failedDepositJob);
+	
+						csvFilePrinter.printRecord(data);
+					}
+				}
+			} catch (Exception e) {
+				log.error("Failed to write data to {}", filePath, e);
+			} finally {
```
bbpennel: If you switch over to opening everything in the try() section you shouldn't need this finally

static/js/admin/performance_visualizations/assets/ramda/unc-custom-build.js
```diff
@@ -0,0 +1,162 @@
+//  Ramda v0.21.0
```
bbpennel: What is custom about UNC in this file?

static/js/admin/performance_visualizations/helpers.js
```diff
@@ -0,0 +1,584 @@
...
+            "<li>" + "Files Ingested: " + this.numFormat(d.throughput_files) + "</li>" +
+            "<li>" + "Total MB Ingested: " + this.numFormat(d.throughput_bytes) + "</li>" +
+            "<li>" + "Avg Filesize (MB): " + this.numFormat(d.avg_filesize) + "</li>" +
+        "</ul>" +
+
+        "<p class='text-center'>Operations Metrics</p>" +
+
+        "<ul class='list-unstyled smaller'>" +
+        "<li>" + "Moves: " + this.numFormat(d.moves) + "</li>" +
+        "<li>" + "Finished Enh: " + this.numFormat(d.finished_enhancements) + "</li>" +
```
bbpennel: For this kind of total, we definitely want to make sure it is encompassing all enhancements, not just TechnicalMetadata

static/js/admin/performance_visualizations/assets/ramda/unc-custom-build.js
lfarrell: It is a build of just the bits of ramda.js used, instead of the whole library. Might be better renamed to something like ramda-merge-with-key.js, which describes what's in it.

PR #469, opened 2016-05-16
Visuals clean

lfarrell:


admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -0,0 +1,334 @@
...
+					String failedDepositJob = depositJob.get("failed-job:edu.unc.lib.dl.cdr.services.techmd.TechnicalMetadataEnhancementService");
+
+					for (String operation : operations) {
+						String operationDate = operation.split(":")[1];
+						
+						if (operationDate.equals(jobDate)) {
+							operationJob = jedis.hgetAll(operation);
+
+							String moves = (operationJob.get("moves"));
+							String imageEnhancements = (operationJob.get("finished-enh:edu.unc.lib.dl.cdr.services.imaging.ImageEnhancementService"));
```
bbpennel: I would suggest making an array of enhancement class names, and then looping through them to when adding to the data instead of hardcoding a series of variables like this.  It'll be a pain when we add more classes to have to keep counting all the zeroes to add

PR #472, opened 2016-05-20
add css for visited articles

hannahlwang:


bbpennel:
If the objective is to make the title for search results change color after visiting, and no other links, it might be better to make the rule:
```css
.search_item h2 a:visited {
	color: #629BB8;
}
```

Which limits it to the class for a search result, rather than the layout element which a white background.  Are these the only links we need to worry about?

The color looks nice, my only concern might be that it is very similar to the deleted item color, but I'm not sure if that is a problem.  You can see a deleted item here ("singleETD" with 0 children):
https://cdr-qa.lib.unc.edu/list/uuid:ad44255f-4263-4d26-84f7-8d3de9de4773

What do you think?

hannahlwang:
@bbpennel That's a good idea. What do you think of this lighter gray?

bbpennel:
Hmm, it stands out, but grey to me indicates that something is not active or clickable.  

Tried to find some examples, a surprising number of sites don't change colors it turns out.  Google search results are a fairly minor shift from #190BB8 to #8B0099, so from a bluer tone to a more purplish one.  Stackoverflow also goes from a blue with green in it to a darker blue (#0077CC => #0059A1), wikipedia goes from a blue with red and green components to one that is mostly blue, a little red, and darker.  

So the trend seems to be to shift to a cooler and darker tone.  Our links are currently a dark blue-green, which is a little unfortunate since we can't darken much.  It'd be nicer if we had different base colors for header links (lighter and more blue).  Unless we want to try some larger change though, we might want to do something in the blue/purple range?

bbpennel:
This was a pretty interesting article, although I don't think we want to try adopting those kinds of standards currently: http://webaim.org/blog/wcag-2-0-and-link-colors/

hannahlwang:
@bbpennel updated the link color

PR #476, opened 2016-06-20
Previous/Next record navigation option

lfarrell:
Allows users to navigate backwards and forwards through records in the same folder.

### DIFF
--- a/access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
+++ b/access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
@@ -212,7 +212,6 @@ public class FullRecordController extends AbstractSolrSearchController {
 			
 			int selectedRecord = 0;
 			int index = 0;
-			
 			for (BriefObjectMetadataBean neighbor : neighbors) {
 				if (neighbor.getId().equals(briefObject.getId())) {
 					selectedRecord = index;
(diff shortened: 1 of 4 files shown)


access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
```diff
@@ -206,6 +206,32 @@ public String getFullRecord(String pid, Model model, HttpServletRequest request)
 			List<BriefObjectMetadataBean> neighbors = queryLayer.getNeighboringItems(briefObject,
 					searchSettings.maxNeighborResults, accessGroups);
 			model.addAttribute("neighborList", neighbors);
+			
+			// Get previous and next record in the same folder if there are any
+			List<BriefObjectMetadataBean> previousNext = new ArrayList<>();
+			
+			int selectedRecord = 0;
+			int index = 0;
+			for (BriefObjectMetadataBean neighbor : neighbors) {
```
daines: Can we use List#indexOf here? https://docs.oracle.com/javase/8/docs/api/java/util/List.html#indexOf-java.lang.Object-

access/src/main/webapp/WEB-INF/jsp/fullRecord/navigationBar.jsp
```diff
@@ -28,7 +30,31 @@
 					<c:param name="linkLast">true</c:param>
 					<c:param name="queryPath">list</c:param>
 				</c:import>
+				
+				<ul id="prevNext">
+				<c:forEach items="${previousNext}" var="neighbor" varStatus="status">
+					<c:url var="fullRecordUrl" scope="page" value="record/${neighbor.id}">
+					</c:url>
+					<c:if test="${not empty neighbor}">
+						<c:set var="hasListAccessOnly" value="${cdr:hasListAccessOnly(requestScope.accessGroupSet, neighbor)}"/>
```
daines: When is this variable used?

PR #478, opened 2016-06-22
Metrics requests correctly pull data for all job types from redis

lfarrell:


admin/src/main/java/edu/unc/lib/dl/admin/controller/PerformanceMonitorController.java
```diff
@@ -215,9 +215,9 @@ public String getOperationsData() {
 						
 						csvFilePrinter.printRecord(data);
 					}
-
-					csvFilePrinter.close();
 				}
+				
+				csvFilePrinter.close();
```
daines: CSVPrinter is Closeable, so you can declare it in a try-with-resources statement (as with the Jedis instance) and avoid calling close() explicitly. https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

PR #484, opened 2016-08-02
Fcrepo4 mapping

lfarrell:
Update deposit pipeline to use RDF Premis events.

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/validate/ValidateMODS.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/validate/ValidateMODS.java
@@ -63,6 +63,7 @@ public class ValidateMODS extends AbstractDepositJob {
 	public void runJob() {
 		Model premisNormalizationEvent;
 		String message;
+		String detailNote;
 		int count = 0;
 		int invalidXSD = 0;
 		int invalidVocab = 0;
@@ -82,14 +83,14 @@ public class ValidateMODS extends AbstractDepositJob {
 			PID p = DepositConstants.getPIDForTagFile(f.getPath());
 			String xsdMessage = "Validation of Controlled Vocabularies in Descriptive Metadata (MODS)";
 			
-			PremisLogger premisLogger = createOrAppendToEventsFile(p);
-			PremisEventBuilder premisValidationBuilder = premisLogger.buildEvent(Premis.Validation, null);
+			PremisLogger premisLogger = getPremisLogger(p);
+			PremisEventBuilder premisValidationBuilder = premisLogger.buildEvent(Premis.Validation);
 			Model premisEvent = premisValidationBuilder
 					.addEventDetail(xsdMessage)
 					.create();
 			premisLogger.writeEvent(premisEvent); 
 			
-			PremisEventBuilder premisNormalizationBuilder = premisLogger.buildEvent(Premis.Normalization, null);
+			PremisEventBuilder premisNormalizationBuilder = premisLogger.buildEvent(Premis.Normalization);
 			try {
 				// XSD validation
 				getModsSchema().newValidator().validate(new StreamSource(f));
@@ -99,8 +100,10 @@ public class ValidateMODS extends AbstractDepositJob {
 				premisLogger.writeEvent(premisNormalizationEvent);
 			} catch (SAXException e) {
 				invalidXSD++;
+				
 				premisNormalizationEvent = premisNormalizationBuilder
-						.addEventDetail("MODS is not valid with respect to the schema (XSD). " + e.getMessage())
+						.addEventDetail("MODS is not valid with respect to the schema (XSD)")
+						.addEventDetailOutcomeNote(e.getMessage())
 						.create();
 				premisLogger.writeEvent(premisNormalizationEvent);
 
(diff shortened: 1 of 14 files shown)


deposit/src/main/java/edu/unc/lib/deposit/work/AbstractDepositJob.java
```diff
@@ -232,68 +208,24 @@ protected void verifyRunning() {
...
 	 * Appends an event to the PREMIS document for the given PID, creating the document if it does not already exist or
 	 * is corrupt.
 	 *
 	 * @param pid
-	 * @param event
 	 * @return the premis document file
 	 */
-	private File createOrAppendToEventsFile(PID pid, Element event) {
-		File file = new File(depositDirectory, DepositConstants.EVENTS_DIR + "/" + pid.getUUID() + ".xml");
+	public PremisLogger createOrAppendToEventsFile(PID pid) {
```
bbpennel: This method doesn't do any appending anymore, how about renaming it to "getPremisLogger"?

metadata/src/main/java/edu/unc/lib/dl/event/PremisLogger.java
```diff
@@ -0,0 +1,90 @@
...
+	public String eventId;
+	public String cdrEventURI = "http://cdr.lib.unc.edu/event/";
+	
+	public PremisLogger(PID pid, File file) {
+		this.pid = pid;
+		this.file = file;
+		this.eventId = generateUUID();
+	}
+	
+	public PremisEventBuilder buildEvent(Resource eventType, Date date) {
```
bbpennel: Since date is being passed in as null every time so far, I think we should add an extra version of this method
public PremisEventBuilder buildEvent(Resource eventType)

Use that everywhere, and directly mention in the javadoc that the event will be assigned the current time as a timestamp

deposit/src/main/java/edu/unc/lib/deposit/work/AbstractDepositJob.java
bbpennel: Also, the javadoc needs to be updated

metadata/src/main/java/edu/unc/lib/dl/util/PremisEventBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+	private Date date;
+	private Model model;
+	private PremisLogger premisLogger;
+	private Resource premisObjResc;
+	
+	public PremisEventBuilder(String eventId, Resource eventType, Date date,
+			PremisLogger premisLogger) {
+		this.eventId = eventId;
+		this.eventType = eventType;
+		this.date = date;
```
bbpennel: Doesn't seem like there is a reason to hold date and eventType as instance variables on the PremisEventBuilder class, they don't need to live long enough for the call to addEvent, so maybe just pass them in as parameters?

metadata/src/main/java/edu/unc/lib/dl/util/PremisEventBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+		premisObjResc.addProperty(Premis.hasAgentName, sourceDataStream);
+		premisObjResc.addProperty(Premis.hasAgentType, "Source Data");
+		premisObjResc.addProperty(Premis.hasAgentName, destDataStream);
+		premisObjResc.addProperty(Premis.hasAgentType, "Derived Data");
+		
+		return this;
+	}
+	
+	public Model create() {
+		return getModel();
```
bbpennel: Since the event is really a resource rather than a model, it would make sense to me for this to return the event resource instead.  You can get back to the model in other places if needed by calling .getModel() on the resource.

deposit/src/main/java/edu/unc/lib/deposit/validate/ValidateMODS.java
```diff
@@ -77,57 +81,72 @@ public boolean accept(File f) {
...
+						.addEventDetail("MODS is valid with respect to the schema (XSD)")
+						.create();
+				premisLogger.writeEvent(premisNormalizationEvent);
 			} catch (SAXException e) {
 				invalidXSD++;
-				PremisEventLogger.addDetailedOutcome(
-						xsdEvent,
-						"MODS is not valid with respect to the schema (XSD)",
-						e.getMessage(), null);
+				premisNormalizationEvent = premisNormalizationBuilder
```
bbpennel: Are these equivalent?  addDetailedOutcome was creating a number of sub-elements

metadata/src/main/java/edu/unc/lib/dl/util/PremisEventBuilder.java
lfarrell: Not entirely sure what you mean here?

deposit/src/main/java/edu/unc/lib/deposit/validate/VirusScanJob.java
```diff
@@ -134,17 +135,15 @@ public void runJob() {
...
-							PremisEventLogger.Type.VIRUS_CHECK,
-							"File passed pre-ingest scan for viruses.", href.getKey(),
-							ContentModelHelper.Datastream.DATA_FILE.getName());
-					PremisEventLogger.addSoftwareAgent(ev2, "ClamAV", version);
-					PremisEventLogger.addDetailedOutcome(ev2, "success", null,
-							null);
-					appendDepositEvent(href.getKey(), ev2);
-						scannedObjects++;
+					premisEvent = premisEventBuilder.addSoftwareAgent("ClamAV", version)
+						.addEventDetail("success")
```
bbpennel: I believe the message here should be "File passed pre-ingest scan for viruses."?

deposit/src/main/java/edu/unc/lib/deposit/work/AbstractDepositJob.java
```diff
@@ -232,68 +208,23 @@ protected void verifyRunning() {
...
-			if (event != null) {
-				dom.getRootElement().addContent(event.detach());
-			}
-
-			try (FileOutputStream out = new FileOutputStream(file, false)) {
-				new XMLOutputter(Format.getPrettyFormat()).output(dom, out);
-			}
+				
+			} 
+			premisLogger = createNewEventsFile(pid, file);
```
bbpennel: I think we can drop createNewEventsFile, all the method is doing is calling the constructor of PremisLogger which is doing the actual file creation

metadata/src/main/java/edu/unc/lib/dl/util/PremisEventBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+		Resource premisObjResc = getResource();
+
+		premisObjResc.addProperty(Premis.hasEventType, eventType);
+		premisObjResc.addProperty(Premis.hasEventDateTime, dateFormat.format(date));
+		
+		return this;
+	}
+	
+	public PremisEventBuilder addEventDetail(String message, Object... args) {
+		if(args != null) {
```
bbpennel: Might as well not persist the squished "if(" into a new class.  There are a couple in here

PR #488, opened 2016-08-22
Add in premis checksum events.

lfarrell:


deposit/src/main/java/edu/unc/lib/deposit/normalize/Simple2N3BagJob.java
```diff
@@ -126,6 +130,27 @@ private void populateSimple(Model model, Resource primaryResource, String alabel
...
 		}
+		
+		String checksum = null;
+		String fullPath = contentFile.toString();
+		
+		try {
+			checksum = DigestUtils.md5Hex(new FileInputStream(fullPath));
+			
+			PremisLogger premisDepositLogger = getPremisLogger(getDepositPID());
+			PremisEventBuilder premisDepositEventBuilder = premisDepositLogger.buildEvent(Premis.MessageDigestCalculation);
```
bbpennel: This is just a style and readability thing, but the eventBuilder generally doesn't need to be visible as a local variable if you aren't going to need to use it past a single chain of methods, like so:

```java
Resource event = premisDepositLogger.buildEvent(Premis.MessageDigestCalculation)
        .addEventDetail("Checksum for file is {0}", checksum)
        .addSoftwareAgent(SoftwareAgent.depositService.getFullname())
        .create();
```

deposit/src/main/java/edu/unc/lib/deposit/normalize/DirectoryToBagJob.java
```diff
@@ -113,10 +117,19 @@ public void runJob() {
 				
 				try {
 					checksum = DigestUtils.md5Hex(new FileInputStream(fullPath));
+					
+					PremisLogger premisDepositLogger = getPremisLogger(getDepositPID());
```
bbpennel: Both of these events should be events on the object checksummed rather than the deposit

deposit/src/main/java/edu/unc/lib/deposit/normalize/Simple2N3BagJob.java
```diff
@@ -126,6 +128,26 @@ private void populateSimple(Model model, Resource primaryResource, String alabel
...
 					contentFile.getAbsolutePath());
 		}
+		
+		String checksum = null;
+		String fullPath = contentFile.toString();
+		
+		try {
+			checksum = DigestUtils.md5Hex(new FileInputStream(fullPath));
+			
+			PremisLogger premisDepositLogger = getPremisLogger(getDepositPID());
```
bbpennel: This should be an event of the file being uploaded, not the deposit.  You can get that in this case by using primaryResource.toString()

PR #500, opened 2016-09-07
Make menu text dynamic

harringj:


static/js/EditFilenameForm.js
```diff
@@ -0,0 +1,53 @@
+define('editFilenameForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStateChangeMonitor', 'tpl!../templates/admin/editFilenameForm',
```
bbpennel: This file should be located in the same directory as other admin js classes, so static/js/admin/src/

static/js/EditFilenameForm.js
```diff
@@ -0,0 +1,53 @@
...
+		'ModalLoadingOverlay', 'AbstractFileUploadForm', 'AlertHandler'], 
+		function($, ui, _, RemoteStateChangeMonitor, editFilenameForm, ModalLoadingOverlay, AbstractFileUploadForm) {
+	
+	var defaultOptions = {
+			title : 'Edit Filename',
+			createFormTemplate : editFilenameForm,
+			showUploadProgress : false
+	};
+	
+	function editFilenameForm(options) {
```
bbpennel: Since this is a class, the first letter should be capitalized, so EditFilenameForm

static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -130,7 +132,11 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'E
 			items["editAccess"] = {name : 'Edit Access'};
 		
 		if ($.inArray('editDescription', metadata.permissions) != -1) {
-			items["editLabel"] = {name : 'Edit Label'};
+			if(isContainerFlag) {
```
bbpennel: missing a space after "if"

static/js/admin/src/EditFilenameForm.js
```diff
@@ -0,0 +1,53 @@
+define('EditFilenameForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStateChangeMonitor', 'tpl!../templates/admin/editFilenameForm', 
+		'ModalLoadingOverlay', 'AbstractFileUploadForm', 'AlertHandler'], 
+		function($, ui, _, RemoteStateChangeMonitor, editFilenameForm, ModalLoadingOverlay, AbstractFileUploadForm) {
```
bbpennel: I would suggest changing editFilenameForm to editFilenameTemplate here in this parameter list to avoid a conflict with the name of the class being defined here.

static/js/admin/src/EditFilenameForm.js
```diff
@@ -0,0 +1,53 @@
...
+	
+	var defaultOptions = {
+			title : 'Edit Filename',
+			createFormTemplate : editFilenameForm,
+			showUploadProgress : false
+	};
+	
+	function editFilenameForm(options) {
```
bbpennel: editFilenameForm should be EditFilenameForm, along with all the other places where its prototype is being established

static/js/admin/src/EditFilenameForm.js
harringj: I left this ref alone for now, since I patterned this class after EditLabelForm. Just thought maybe we should keep those consistent.

static/templates/admin/editFilenameForm.html
```diff
@@ -0,0 +1,18 @@
+<iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
+<form id="edit_filename_form" method="post" action="editfilename/<%= pid %>" enctype="multipart/form-data" target="upload_file_container" class="edit_form edit_filename_form">
```
bbpennel: You will need to change the action here back to "editlabel/<%= pid %>" since that is the mapping being used by the controller server side
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/master/admin/src/main/java/edu/unc/lib/dl/admin/controller/EditLabelController.java#L27

harringj: Done! Good call -- that seems to have fixed the problem :-)

PR #503, opened 2016-09-12
Fcrepo4 PID creation

bbpennel:


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/PIDs.java
```diff
@@ -0,0 +1,119 @@
...
+		String repositoryPath;
+
+		if (value.startsWith(repository.getFedoraBase())) {
+			// Given value was a fedora path. Remove the base and decompose
+			repositoryPath = value;
+			String path = value.substring(repository.getFedoraBase().length());
+
+			Matcher matcher = RepositoryPathConstants.repositoryPathPattern.matcher(path);
+			if (matcher.matches()) {
+				qualifier = matcher.group(1);
```
harringj: What determines the values of these numbers? What do you think about storing them as class constants for readability/maintainabilty?

bbpennel: The group numbers?  They are indexes referencing groups over in this regular expression pattern https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/1aa1a2d8cfaaa7bcfa698745463a3515ba5bc3b3/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPathConstants.java#L41 (the other regex after it is used below similarly).  

These particular indexes/values are not likely to be used elsewhere.  I wonder if that should be taken as a sign that the patterns should be in this class rather than the constants one, but I was trying to put the pieces used to understand paths together in one place (RepositoryPathConstants.java)

PR #504, opened 2016-09-13
Add mods column to csv export

harringj:
I have deployed to QA and the exported .csv file now has a column for content status that is populated with appropriate values.


bbpennel:
You will want to check with Julie about if this satisfies the use case.  The content status field does not exclusively keep track of whether an object has MODS, it's more like a bucket of tags.  So the CSV could end up with "Described", "Not Described", "Default Access Object Assigned", "Invalid Term" in that field (occasionally multiple at once).  What I'm not sure about is if Julie was looking to have a "Is Described" column, or a "Content Status" column

harringj:
Yeah, I agree, it would be important to check with her about that. I'll add a comment to the JIRA ticket.

admin/src/main/java/edu/unc/lib/dl/admin/controller/ExportController.java
```diff
@@ -50,6 +50,9 @@
 @RequestMapping("export")
 public class ExportController extends AbstractSolrSearchController {
 	
+	private static final String DESCRIBED = "Described";
```
bbpennel: My only suggestion would be to use the constants that already exist over here https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/master/solr-search/src/main/java/edu/unc/lib/dl/search/solr/util/FacetConstants.java#L20 since this is referencing a specific facet value that is generated using these constants

PR #505, opened 2016-09-14
Bugfix delete extra paren

harringj:
Somehow an extra parenthesis managed to sneak into my code. I just noticed it as I went to close the Eclipse window with the ExportController class. I pushed the bugfix out to QA and all seems to look good. Not sure why it worked earlier, and why I didn't see any issues with the build.


harringj:
I'm going to start over. My local repo is a mess :-)

PR #507, opened 2016-09-19
Create deposit record and supporting clients

bbpennel:
Implements fcrepo4 clients for creating deposit records with manifests.

Adds a client for BinaryObjects, in the context of adding manifests.

Updates fcrepo-java-client version to the most recent snapshot for bug fixes.


deposit/src/test/resources/ingest-bags/fcrepo4/proquest-bag/everything.n3
```diff
@@ -1,5 +1,5 @@
-<info:fedora/uuid:55c262bf-9f15-4184-9979-3d8816d40103> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/1999/02/22-rdf-syntax-ns#Bag> .
-<info:fedora/uuid:55c262bf-9f15-4184-9979-3d8816d40103> <http://www.w3.org/1999/02/22-rdf-syntax-ns#_1> <info:fedora/uuid:fc39ce7d-b81b-4c2b-bf38-e2f18ebfd4dd> .
+<http://examle.com/deposit/55/c2/62/bf/55c262bf-9f15-4184-9979-3d8816d40103> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/1999/02/22-rdf-syntax-ns#Bag> .
```
harringj: "example" instead of "examle"? "deposits" rather than "deposit"?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
```diff
@@ -0,0 +1,97 @@
...
+	 * @param path
+	 * @param model
+	 * @return
+	 * @throws FedoraException
+	 */
+	public URI createDepositRecord(URI path, Model model) throws FedoraException {
+		URI depositRecordUri;
+
+		try (FcrepoResponse response = getClient().put(path)
+				.body(RDFModelUtil.streamModel(model), "text/turtle")
```
harringj: turtle?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
```diff
@@ -0,0 +1,97 @@
...
+			throw new FedoraException("Unable to create deposit record at " + path, e);
+		} catch (FcrepoOperationFailedException e) {
+			throw ClientFaultResolver.resolve(e);
+		}
+
+		return depositRecordUri;
+	}
+
+	public URI createPremisEvent(URI objectUri, Model model) throws FedoraException {
+		return null;
```
harringj: Why is null being returned here?

bbpennel: Sorry, its a stub, that method will be implemented in the next PR about premis events

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
bbpennel: Turtle is a particular encoding for triples which is the default in fedora 4.  Perhaps the mimetype should be a constant somewhere rather than hardcoded?

harringj: Yeah, I think that could be helpful as an instance of self-documenting code.

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactoryIT.java
```diff
@@ -0,0 +1,88 @@
...
+
+import edu.unc.lib.dl.rdf.Cdr;
+import edu.unc.lib.dl.rdf.Ldp;
+import edu.unc.lib.dl.util.RDFModelUtil;
+
+/**
+ * 
+ * @author bbpennel
+ *
+ */
```
harringj: Should this class name end with "Test", just to make it clearer that it is a test?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
```diff
@@ -0,0 +1,196 @@
...
+		return model;
+	}
+
+	/**
+	 * Set the model
+	 * 
+	 * @param model
+	 * @return
+	 */
+	public RepositoryObject setModel(Model model) {
```
harringj: Out of curiosity, why is the return type of this setter not void?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectDataLoader.java
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		return this;
+	}
+
+	public RepositoryObjectDataLoader loadHeaders(RepositoryObject obj) throws FedoraException {
+		PID pid = obj.getPid();
+
+		try (FcrepoResponse response = getClient().head(pid.getRepositoryUri()).perform()) {
+			if (response.getStatusCode() != 200) {
```
harringj: I tend to prefer using constants, like Apache's SC_OK, rather than hardcoded numbers, but that's just a matter of personal taste.

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactoryIT.java
bbpennel: A little weird, but the convention for integration tests is to end the class name with IT and end unit tests with Test.  I believe this comes from the maven-surefire plugin

harringj: I did not know that. Thanks for the explanation!

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
bbpennel: Geneally it would be to allow for method chaining (ie, obj.setModel(...).setBlah(...).doSomething()), but I haven't been consistent about it here and it doesn't seem like that useful of a pattern for RepositoryObjects and their descendents, so I am going to switch to void for these methods

PR #509, opened 2016-09-21
URI path join

bbpennel:


metadata/src/test/java/edu/unc/lib/dl/util/URIUtilTest.java
```diff
@@ -0,0 +1,39 @@
...
+				URIUtil.join("http://example.com/base", "path"));
+
+		assertEquals("http://example.com/base/path/to",
+				URIUtil.join("http://example.com/", "base/", "/path", "to/"));
+
+		assertEquals("path/to", URIUtil.join("", "path", "to"));
+
+		assertEquals("http://example.com/base/path",
+				URIUtil.join(URI.create("http://example.com/base"), "path"));
+	}
```
harringj: My one suggestion would be to cover a few more test secnarios:
(1)  a null base String
(2)  an empty base String
(3) a null array of segments
(4) an empty array of segments

PR #512, opened 2016-09-27
PREMIS event clients

bbpennel:
* Implemented clients for adding premis events to the repository.  
* Added integration tests, and migrated shared dependency setup between integration tests to context file.  
* Refactoring of exceptions in new api to be unchecked to support cases where different implementations do not unnecessarily require the same exceptions and to clean up code where bubbling is expected


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
```diff
@@ -133,6 +137,31 @@ public BinaryObject createBinary(URI path, String slug, InputStream content, Str
...
+	 * @param model
+	 *            Model containing properties of this event. Must only contain
+	 *            the properties for one event.
+	 * @return URI of the event created
+	 * @throws FedoraException
+	 */
+	public URI createPremisEvent(PID eventPid, Model model) throws FedoraException {
+
+		try (FcrepoResponse response = getClient().put(eventPid.getRepositoryUri())
+				.body(RDFModelUtil.streamModel(model), "text/turtle")
```
harringj: What do you think about storing the literal "text/turtle" in a constant?

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/PremisLogger.java
```diff
@@ -83,37 +87,35 @@ public PremisEventBuilder buildEvent(Resource eventType) {
...
+		if (model != null) {
+			return model;
+		}
+
 		Model model = ModelFactory.createDefaultModel();
 
-		if (premisFile.exists()) {
+		if (premisFile != null && premisFile.exists()) {
 			InputStream in = FileManager.get().open(premisFile.getAbsolutePath());
 			model.read(in, null, "TURTLE");
```
harringj: Again, for purposes of maintainability, maybe "TURTLE" should go in a constant.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
```diff
@@ -133,6 +143,75 @@ public BinaryObject createBinary(URI path, String slug, InputStream content, Str
...
+	 * @param model
+	 *            Model containing properties of this event. Must only contain
+	 *            the properties for one event.
+	 * @return URI of the event created
+	 * @throws FedoraException
+	 */
+	public PremisEventObject createPremisEvent(PID eventPid, Model model) throws FedoraException {
+
+		try (FcrepoResponse response = getClient().put(eventPid.getRepositoryUri())
+				.body(RDFModelUtil.streamModel(model), "text/turtle")
```
harringj: Change literal to a constant?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
```diff
@@ -133,6 +143,75 @@ public BinaryObject createBinary(URI path, String slug, InputStream content, Str
...
+				.perform()) {
+
+			return new PremisEventObject(PIDs.get(response.getLocation()), this, repositoryObjectDataLoader);
+		} catch (IOException e) {
+			throw new FedoraException("Unable to create premis event for " + eventPid, e);
+		} catch (FcrepoOperationFailedException e) {
+			throw ClientFaultResolver.resolve(e);
```
harringj: What's going on behind the scenes here? I'm guessing resolve() returns a throwable, but is it doing anything else?

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/PremisLogger.java
bbpennel: This is fixed in later commits

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
bbpennel: ClientFaultResolver.resolve inspects the exception that comes back from the fcrepoClient request and rethrows it as a FedoraException (our local exception for stuff that goes wrong in fedora) or preferably, one of its subclasses.  At the moment this just involves looking at the status code of the response, but in the past it has been a bit more involved since you had to inspect fedora 3's response bodies to figure out what went wrong.

This lets us keep the FcrepoClient library from percolating up into the rest of the repository and allows us to react to exceptions that indicate that, for example, an authentication error occurred instead of a generic exception.

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/FilePremisLogger.java
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	@Override
+	public List<PremisEventObject> getEvents() {
+		List<PremisEventObject> events = new ArrayList<>();
+
+		// Find all of the events and construct a list of PremisEventObjects from them.
+		ResIterator eventIt = model.listResourcesWithProperty(Premis.hasEventType);
+		while (eventIt.hasNext()) {
+			Resource eventResc = eventIt.nextResource();
```
harringj: I may be reading this incorrectly, but won't the code as written lead to the first Resource being ignored? In other words, wouldn't we want to get the eventPid from the Resource eventIt is currently pointing to when it's first created, and then call nextResource() afterwards?

ResIterator eventIt = model.listResourcesWithProperty(Premis.hasEventType);
Resource eventResc = eventIt.getResource(); // or whatever the appropriate method is for this action
 +		while (eventIt.hasNext()) {
 +		      PID eventPid = PIDs.get(eventResc.getURI());	
                      eventResc = eventIt.nextResource();

fcrepo-clients/src/test/java/edu/unc/lib/dl/util/FilePremisLoggerTest.java
```diff
@@ -42,9 +63,16 @@ public void setup() throws Exception {
 		eventType = Premis.VirusCheck;
 		premisFile = File.createTempFile(depositUUID, ".ttl");
 		premisFile.deleteOnExit();
-		premis = new PremisLogger(pid, premisFile);
+		premis = new FilePremisLogger(pid, premisFile, repository);
 		date = new Date();
 		SoftwareAgentConstants.setCdrVersion("4.0-SNAPSHOT");
```
harringj: Why is the SNAPSHOT version hardcoded here?

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/FilePremisLogger.java
bbpennel: Iterators a little weird, but this is the pattern for them.  Right after getting an iterator, eventIt.hasNext() will evaluate to true if there is at least one item in the iterator, but it doesn't have any sort of "get" method for getting the current item, instead you have to call eventIt.nextResource() to get the first item.  Many iterators are a "read once" sort of deal, so you can only get each entry once by calling the .next method.

It actually appears to be standard practice these days to use a for loop instead of a while loop for these things, so perhaps I should switch over in the future:
`for (Iterator<?> it = c.iterator(); it.hasNext(); ) {
  it.next();
}
`

fcrepo-clients/src/test/java/edu/unc/lib/dl/util/FilePremisLoggerTest.java
bbpennel: That line wasn't necessary for this test, so I've removed it.

PR #514, opened 2016-09-28
Therapy Dog forms in Admin UI (Create Work Option)

lfarrell:


bbpennel:
Going to hold off on reviewing this until CDR-1553 (depositing as the user) is addressed

bbpennel:
To save some potential user frustration later, can the close button on the modal prompt the user with a "Are you sure you wish to close this form?" choice rather than immediately closing?  In case of accidental clicks or escape presses in the middle of inputing a lengthy form.  Ideally, they would only get this prompt after having selected a form (and not after submitting), but that would just be a bonus, no need to do so if it is technically challenging

static/js/admin/src/AddFileForm.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * Implements functionality and UI for the generic Ingest Package form
```
bbpennel: Apparently this comment has made it into a few files.  Maybe give it something that relates to this specifc form?

static/js/admin/src/AddMenu.js
```diff
@@ -9,19 +9,56 @@ define('AddMenu', [ 'jquery', 'jquery-ui', 'underscore', 'CreateContainerForm',
...
+
 		if ($.inArray('addRemoveContents', this.container.permissions) != -1) {
-			items["addContainer"] = {name : "Add Container"};
-			items["ingestPackage"] = {name : "Add Ingest Package"};
-			items["ingestSource"] = {name : "Add from File Server"};
-			items["simpleObject"] = {name : "Add Simple Object"};
-			
+			var self = this;
+
+			items["addContainer"] = {
```
bbpennel: This button is recreated after a page navigation, so it would also be okay to have an if statement that determines if the options gets added rather than a callback function, so you could cut down on the amount of code here.  

Also, the isVisible method is a bit of a misnomer since it is really "isAggregate", which is the what's actually being evaluated for all of the "add" options.

Additionally, I can't tell if this is being handled another way since I can't deploy the application at the moment, but none of these options should be visible if the self.container.type == "File" (or simple or whatever its called)

static/js/admin/src/CreateWorkObjectForm.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * Implements functionality and UI for the generic Ingest Package form
```
bbpennel: Update me!

static/js/admin/src/CreateWorkObjectForm.js
```diff
@@ -0,0 +1,71 @@
...
+		title : 'Add New Work',
+		createFormTemplate : workObjectTemplate
+	};
+
+	function CreateWorkObjectForm(options) {
+		this.options = $.extend({}, AbstractFileUploadForm.prototype.getDefaultOptions(), defaultOptions, options);
+	}
+	
+	CreateWorkObjectForm.prototype.constructor = CreateWorkObjectForm;
+	//CreateWorkObjectForm.prototype = Object.create( AbstractFileUploadForm.prototype );
```
bbpennel: Remove me!

static/js/admin/src/CreateWorkObjectForm.js
```diff
@@ -0,0 +1,71 @@
...
+		var iframe = $("#work_submission_form");
+		$("select[name='name']").on("change", function() {
+			var ajaxIcon = $(".loading-icon");
+
+			var formUrl = location.href.split("/");
+			var protocol = formUrl[0];
+			var collectionId = iframe.attr("title");
+			var host, subDirectory;
+
+			if (/cdr/.test(formUrl[2])) {
```
bbpennel: Hmm, I'd really rather not environmental details like this in our view code.  I would rather if we added a property to server.properties that specifies the base uri for the forms app in this environment, which would get passed down to this form.  forms.site.baseUri or something along those lines.

static/js/admin/src/CreateWorkObjectForm.js
```diff
@@ -0,0 +1,71 @@
...
+
+			iframe.on("load", function () {
+				$(this).removeClass("addwork");
+				ajaxIcon.addClass("in-admin-iframe");
+			});
+
+		});
+
+		containingDialog.on("dialogclose", function(e, ui) {
+			$(this).dialog("destroy");
```
bbpennel: See comment about prompting for confirmation before closing

static/templates/admin/addFileForm.html
```diff
@@ -0,0 +1,23 @@
+<iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
+<form id="create_simple_object_form" method="post" action="ingest/<%= pid %>" enctype="multipart/form-data" target="upload_file_container" class="edit_form create_simple_object_form">
+    <p>Add a single file as a new object.  If no label is provided, the filename will be used.</p>
```
bbpennel: Should probably read "Add a file to this aggregate object" or "Add a file to this work" if you want to future proof it

static/templates/admin/resultTableHeader.html
```diff
@@ -20,9 +20,11 @@ <h2 class="container_title">
 					<% } %>
 				</h2>
 			</span>
+
 			<%
-			if (container && !containerDeleted && (_.indexOf(container.permissions, "addRemoveContents") != -1
+			if (container && !containerDeleted && container.type !== "File" && (_.indexOf(container.permissions, "addRemoveContents") != -1
```
bbpennel: Hmm, I guess this is removing the button for File objects

lfarrell: Correct.

static/js/admin/src/AddMenu.js
bbpennel: Are you sure that is the case?  Because if so, if you switched pages between a folder you didn't have permission on to one you did, you wouldn't get any entries in the add menu because of the preceeding if statement.  If so, then that needs to be fixed too, but I would like to verify first.

The isVisible method is one that you defined further down as a member function of the AddMenu prototype, so you should be able to name it something else.  I understand that the `visible` property of the menu entry is reserved though

PR #515, opened 2016-09-30
File Objects

bbpennel:
* Implemented clients for cdr:FileObjects
* Implemented creation of pcdm:FileSets
* Added supporting helper methods for adding relations
* Fixes and missing bits


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
```diff
@@ -213,6 +245,27 @@ public Model getObjectModel(URI metadataUri) throws FedoraException {
...
 	}
 
+	/**
+	 * 
+	 * @param subject
+	 * @param property
+	 * @param object
+	 */
+	public void createRelationship(PID subject, Property property, Resource object) {
+		String sparqlUpdate = RDFModelUtil.createSparqlInsert(subject.getRepositoryPath(), property, object);
```
harringj: Just out of curiosity, how doe getRepositoryPath() and getRepositoryURI(), which is used a few lines down, differ?

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/FileObjectIT.java
```diff
@@ -0,0 +1,155 @@
...
+		String thumbFilename = "thumb.png";
+		String thumbMimetype = "image/png";
+		InputStream thumbContentStream = new ByteArrayInputStream(thumbBodyString.getBytes());
+		BinaryObject bObj2 = fileObj.addDerivative("thumb", thumbContentStream, thumbFilename, thumbMimetype, PcdmUse.ThumbnailImage);
+		assertNotNull(bObj1);
+
+		// Retrieve the binary objects directly
+		List<BinaryObject> binaries = fileObj.getBinaryObjects();
+
+		assertEquals("Incorrect number of binaries add", 3, binaries.size());
```
harringj: "...of binaries added"

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPathConstants.java
```diff
@@ -25,25 +25,41 @@
...
 	public static final int HASHED_PATH_DEPTH = 4;
-	
+
+	// Number of characters to use per level of hierarchy in a hashed path
 	public static final int HASHED_PATH_SIZE = 2;
-	
+
 	// Regex pattern for decomposing a repository URI for an object or component of an object
 	public static final Pattern repositoryPathPattern = Pattern
 			.compile("/?([a-zA-Z]+)/([a-f0-9]{" + HASHED_PATH_SIZE + "}/){" + HASHED_PATH_DEPTH + "}"
```
harringj: This obviously wasn't changed in this PR, but I wonder whether we might want the regex strings to be stored with other constants for maintainability, in case our requirements for paths/identifiers should change at some later date.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
bbpennel: They return more or less the same thing, getRepositoryPath returns the absolute path to object, while getRepositoryURI returns the same thing but as a URI object.  The reason for having both is just that there are two apis involved, one that demands strings and the other demands URIs

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPathConstants.java
bbpennel: It's in a file with other constants currently, or do you mean the straight that is passed into Pattern.compile would be a constant on its own?

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/FileObjectIT.java
bbpennel: Fixed in fcrepo4

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPathConstants.java
harringj: Yes, I meant what gets passed to Pattern.complie.

PR #516, opened 2016-10-07
Implement work object client

bbpennel:
Adds client and factory methods for creating Work objects.  Implemented general ContentObject ability to retrieve members for use by work objects.

### DIFF
--- a/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
+++ b/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
@@ -16,10 +16,17 @@
 package edu.unc.lib.dl.fcrepo4;
 
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+import com.hp.hpl.jena.rdf.model.Resource;
+import com.hp.hpl.jena.rdf.model.StmtIterator;
 
 import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.rdf.PcdmModels;
 
 /**
+ * Represents a generic repository object within the main content tree.
  * 
  * @author bbpennel
  *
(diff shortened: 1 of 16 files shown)


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -28,18 +35,35 @@
...
+	/**
+	 * Retrieve a list of member content objects for this object.
+	 * 
+	 * @return
+	 */
+	public List<ContentObject> getMembers() {
+		List<ContentObject> members = new ArrayList<>();
+		Resource resc = getResource();
+
+		for (StmtIterator it = resc.listProperties(PcdmModels.hasMember); it.hasNext(); ) {
```
harringj: This isn't a criticism of what's written here, but I find while-loop syntax with an iterator to be cleaner/easier to read than the for-loop.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
```diff
@@ -74,21 +74,29 @@ public URI getUri() {
...
+
+	/**
 	 * Set the model
 	 * 
 	 * @param model
 	 * @return
 	 */
 	public void setModel(Model model) {
 		this.model = model;
+		// Clear the cached types list
```
harringj: I'm not clear why this line is in the setter.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
bbpennel: Oh, didn't realize that.  I don't have a strong opinion on it, I switched to for-loops after a previous PR you commented on.  Most examples I saw at the time were using for-loops, but the only real reason I've seen to prefer one over the other is the first answer here:
http://stackoverflow.com/questions/99164/iterate-with-for-loop-or-while-loop

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
bbpennel: The list of types needs to be cleared whenever the model is changed to avoid becoming stale since it is just a cached subset of values from the model.  I could shift that logic to the data loader though if that'd make more sense

harringj:
No, it’s fine to keep it as is. I see some value in the argument for limiting the scope of the iterator. I guess I don’t like the fact that the update happens inside the body of the loop rather than in the header, where I would expect it to be in a for-loop.

harringj:
Could you re-name the method changeModel()? I just don’t expect extra logic inside a traditional setter.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
harringj: Could you re-name the method changeModel()? I just don’t expect extra logic inside a traditional setter.

bbpennel: I switched it to storeModel() to make it clear that the method wouldn't behave like a bean property, but also so it wouldn't feel a little weird when called before a value had been set.  (Un)happy compromise?

harringj: :-)

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
harringj: No, it’s fine to keep it as is. I see some value in the argument for limiting the scope of the iterator. I guess I don’t like the fact that the update happens inside the body of the loop rather than in the header, where I would expect it to be in a for-loop.

harringj:
I read through the rest of this, @bbpennel , and I think it looks good. Merge at will :-)

PR #521, opened 2016-10-13
Upload encoding

lfarrell:


admin/src/main/java/edu/unc/lib/dl/admin/controller/IngestController.java
```diff
@@ -82,9 +85,17 @@
...
-		method.addHeader("Content-Disposition", "attachment; filename=" + ingestFile.getOriginalFilename());
 		if (name != null && name.trim().length() > 0)
-			method.addHeader("Slug", name);
+			slugValue = name;
+
+		try {
+			method.setHeader("Content-Disposition", "attachment; filename=" + URLEncoder.encode(ingestFile.getOriginalFilename(), "UTF-8"));
+			method.addHeader("Slug", URLEncoder.encode(slugValue, "UTF-8"));
+		} catch (UnsupportedEncodingException e1) {
+			log.warn("Unable to properly encode value to UTF-8");
```
bbpennel: Could you add a ", e1" to this log statement so that it will print out the trace?  That may also get it to show what failed to encode.  Although I wouldn't be surprised if we never actually saw this

PR #522, opened 2016-10-13
Repository client unit test

bbpennel:
Adds a unit test for the Repository class
Cleans up the API for the class a bit for consistency.

### DIFF
--- a/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
+++ b/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
@@ -307,8 +307,27 @@ public class RepositoryObjectFactory {
 				parentUri, memberUri);
 	}
 
-	public URI createPremisEvent(URI objectUri, Model model) throws FedoraException {
-		return null;
+	/**
+	 * Creates a fedora object at the given location with the provided
+	 * properties
+	 * 
+	 * @param uri
+	 * @param model
+	 * @return
+	 * @throws FedoraException
+	 */
+	public URI createObject(URI uri, Model model) throws FedoraException {
+
+		try (FcrepoResponse response = getClient().put(uri)
+				.body(RDFModelUtil.streamModel(model), TURTLE_MIMETYPE)
+				.perform()) {
+
+			return response.getLocation();
+		} catch (IOException e) {
+			throw new FedoraException("Unable to create object at " + uri, e);
+		} catch (FcrepoOperationFailedException e) {
+			throw ClientFaultResolver.resolve(e);
+		}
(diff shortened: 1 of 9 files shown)


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
```diff
@@ -307,8 +307,27 @@ public URI createMemberLink(URI parentUri, URI memberUri) throws FedoraException
...
+
+		try (FcrepoResponse response = getClient().put(uri)
+				.body(RDFModelUtil.streamModel(model), TURTLE_MIMETYPE)
+				.perform()) {
+
+			return response.getLocation();
+		} catch (IOException e) {
+			throw new FedoraException("Unable to create object at " + uri, e);
+		} catch (FcrepoOperationFailedException e) {
+			throw ClientFaultResolver.resolve(e);
```
harringj: Is this exception thrown when response.getLocation() fails? Just trying to understand which conditions trigger each of these two exceptions.

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/RepositoryTest.java
```diff
@@ -0,0 +1,416 @@
...
+		repository.setClient(fcrepoClient);
+
+		PIDs.setRepository(repository);
+	}
+
+	@Test
+	public void mintDepositRecordPidTest() {
+		PID pid = repository.mintDepositRecordPid();
+
+		assertTrue(pid.getRepositoryPath().contains(RepositoryPathConstants.DEPOSIT_RECORD_BASE));
```
harringj: What do you think about asserting something more specific here, say, .startsWith(RepositoryPathConstants.DEPOSIT_RECORD_BASE) ?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
bbpennel: I believe the IOException comes from closing the response, and the FcrepoOperationFailedException is thrown by any call to the fcrepoClient.put/get/post/etc methods, and is thrown whenever a 400 or 500 status code is returned I believe.

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/RepositoryTest.java
bbpennel: Updated this to be a little bit more exact about where the DEPOSIT_RECORD_BASE is in the returned pid

PR #524, opened 2016-10-17
Update normalization job namespaces

harringj:


fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/BinaryObjectTest.java
```diff
@@ -0,0 +1,128 @@
+package edu.unc.lib.dl.fcrepo4;
```
harringj: You are correct. I've been working this morning on pulling out these files. I accidentally branched from my local fcrepo4 instead of the remote branch.

harringj:
Ok, @bbpennel , should be ready to look at now :-)

bbpennel:
Fcrepo4 works primarily with SHA1 checksums instead of MD5, so there should be another block similar to https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/524/files#diff-dac387d3f4d43a9d580cc40448eb2805R127 but for SHA1s.  I believe there is already a property for that

bbpennel:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/c86c8ce49a956ed2c0f5aba9cb7fe3512a5e4aca/deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java#L103

we should switch over from use any fedora 3 models to using RDF types.  This ties in to the question of whether to update the normalization jobs to reflect the work only model as well

harringj:
But switch over to what, exactly? I don't see a corresponding property in Cdr.java. Should I just pause on this until a decision is made about whether to update the normalization jobs?

PR #525, opened 2016-10-19
updates validation jobs to use new namespaces

harringj:
I updated the two jobs that were still using the old properties.

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/validate/ValidateFileAvailabilityJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/validate/ValidateFileAvailabilityJob.java
@@ -19,7 +19,7 @@ import com.hp.hpl.jena.rdf.model.StmtIterator;
 
 import edu.unc.lib.deposit.work.AbstractDepositJob;
 import edu.unc.lib.dl.fedora.PID;
-import edu.unc.lib.dl.util.ContentModelHelper;
+import edu.unc.lib.dl.rdf.CdrDeposit;
 import edu.unc.lib.staging.Stages;
 import edu.unc.lib.staging.StagingException;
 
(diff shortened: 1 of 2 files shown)


bbpennel:
Could you also get PackageIntegrityCheckJobTest passing?  It is just missing the repository object and its setup, and it might benefit from a parent class that sets that up

deposit/src/test/java/edu/unc/lib/deposit/CleanupDepositJobTest.java
```diff
@@ -105,6 +109,9 @@ public void setup() throws Exception {
 
 		// save mappings
 		FileUtils.writeStringToFile(modifiedStagesConfig, stages.getLocalConfig());
+		
+		PIDs.setRepository(repo);
+		when(repo.getFedoraBase()).thenReturn("http://www.fcrepo.com");
```
bbpennel: Could you add a trailing / to uri?  Also, it might be helpful to make an AbstractDepositJobTest which sets this sort of thing up rather than having to do it in each test.  AbstractNormalizationJobTest could inherit from that

deposit/src/test/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJobTest.java
```diff
@@ -67,14 +68,20 @@
...
 	@Before
 	public void setup() throws Exception {
+		initMocks(this);
 		stages = mock(Stages.class);
 		
 		status = new HashMap<String, String>();
 
 		when(depositStatusFactory.get(anyString())).thenReturn(status);
+		PIDs.setRepository(repo);
+		when(repo.getFedoraBase()).thenReturn("http://www.fcrepo.com");
```
bbpennel: Same comment here about / and parent class

harringj:
Good suggestion about the AbstractDepositJobTest. I created that and updated the other tests.

PR #529, opened 2016-10-25
Fcrepo4 ingest deposit job

bbpennel:
Adds the ingest deposit job along with clients (and changes to clients) to support it, as well as adding unit and integration tests for the job.

This PR does not address updating of the non-test spring context files so that this can be run as a service, which will be addressed in another PR.  It also doesn't remove the superseded fcrepo3 clients.


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -0,0 +1,449 @@
...
+	/**
+	 * Determines how many operations will be required to complete this deposit.
+	 * 
+	 * @param depositBag
+	 */
+	private void calculateWorkRemaining(Bag depositBag) {
+		List<String> contentPaths = new ArrayList<>();
+		DepositGraphUtils.walkChildrenDepthFirst(depositBag, contentPaths, true);
+		// Number of actions is the number of ingest objects plus deposit record
+		setTotalClicks(contentPaths.size());
```
harringj: Do we need a + 1 here to account for the deposit record?

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -0,0 +1,449 @@
...
+	private void populatePreviouslyIngested(List<String> contentPaths) {
+		// Generate a list of object paths from this deposit already in fedora
+		previouslyIngestedSet = new HashSet<>();
+		for (String path : contentPaths) {
+			if (repository.objectExists(PIDs.get(path))) {
+				previouslyIngestedSet.add(path);
+			}
+		}
+		// Update the ingested object count to reflect the number in the repo
+		getJobStatusFactory().setCompletion(getJobUUID(), previouslyIngestedSet.size());
```
harringj: Does this subtract the count of previously ingested objects from the total object count?

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
bbpennel: Currently the deposit record is added via a separate job, although i had been considering merging them together.  Either way though, I've found the inclusion of the deposit record in the object count to be rather confusing, particularly when depositing a single file, so I was leaning towards not reporting it that way.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -0,0 +1,449 @@
...
+		Statement stmt = resc.getProperty(property);
+		if (stmt == null) {
+			return null;
+		}
+		return stmt.getString();
+	}
+
+	/**
+	 * Return an iterator for the children of the given resource, base on what
+	 * type of container is is.
```
harringj: "...it is"

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
bbpennel: No, it is instead starting from the number of already ingested items and keeping the total number of objects the same.  My hope is that this will result in a more consistent count in the status monitor, although might need to see it in action to be sure

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -0,0 +1,449 @@
...
+	 * @throws DepositException
+	 */
+	private void ingestFileObject(ContentObject parent, Resource parentResc, Resource childResc)
+			throws DepositException {
+
+		if (skipResumed(childResc)) {
+			return;
+		}
+
+		// TODO add ACLs
```
bbpennel: Access control properties

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobTest.java
```diff
@@ -0,0 +1,446 @@
...
+		when(depositStatusFactory.isResumedDeposit(anyString())).thenReturn(true);
+
+		ingestFileObjectAsWorkTest();
+
+		// No preprocessing ticks
+		verify(jobStatusFactory).setCompletion(eq(jobUUID), eq(0));
+	}
+
+	@Test
+	public void permTest() {
```
harringj: What is this testing?

bbpennel: Whoops, didn't mean to leave that in, I was just testing to see how two different ways of checking for RDF.types performed to see if I could remove some code from the ingest job.  I'll take this out

deposit/src/test/resources/spring-test/cdr-client-container.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
harringj: What are all of the following spring-related xml files doing?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FileObject.java
```diff
@@ -68,6 +68,12 @@ public FileObject validateType() throws FedoraException {
 		return this;
 	}
 
+	@Override
```
bbpennel: Because it is inherited from the ContentObject parent class.  FileObjects don't get to have any children other than binary files (which are not presently ContentObjects), unlike the other ContentObject subclasses, but they are otherwise usable in the same ways as the others.  I could introduce another parent class where the addMember method comes in, something like ContentContainerObject, and that might be better, particularly if there is more shared functionality exclusive to container types (mainly access control comes to mind):

RepositoryObject > ContentObject > ContentContainerObject > FolderObject
RepositoryObject > ContentObject > FileObject

PR #532, opened 2016-11-01
Update cdr mets normalization job

harringj:


services-worker/src/test/resources/edu/unc/lib/dl/cdr/services/techmd/fitsOutputMultipleLineBreaks.xml
```diff
@@ -1,5 +1,3 @@
-2015-06-13 06:19:59,410 ERROR [main] ToolBelt:79 - Thread 1 error initializing edu.harvard.hul.ois.fits.tools.exiftool.Exiftool: edu.harvard.hul.ois.fits.exceptions.FitsToolCLIException  Message: Error calling external command line routine (Cannot run program "which": java.io.IOException: error=12, Cannot allocate memory)
```
bbpennel: I think we should retain this line while the test that relies on it still exists https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/a7f3c69d0ab4fe71d3a013b3874a47d13d198ab8/services-worker/src/test/java/edu/unc/lib/dl/cdr/services/techmd/TechnicalMetadataEnhancementTest.java#L75

This was here to make sure that the TechMD enhancement could handle dirty output from FITS, which unfortunately has historically been a thing.  The TechMD service class will need to be replaced, since it deals with FOXML, but the part that interacts with the output from FITS will likely need to be the same.

harringj: I reverted this file, @bbpennel . Sorry, I didn't realize it was important; Eclipse was complaining that the line wasn't allowed in prolog.

bbpennel:
We'll need to address SHA1 checksums and get a new METS normalization test before too long, preferably one that actually tests something :)

harringj:
Agreed!

PR #533, opened 2016-11-02
Collection object client

harringj:
Adds a new class and unit test for the fcrepo CollectionObject client. Updates Repository and RepositoryObjectFactory to handle the new type.

### DIFF
--- a/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/CollectionObject.java
+++ b/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/CollectionObject.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright 2016 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */package edu.unc.lib.dl.fcrepo4;
+
+import edu.unc.lib.dl.fedora.FedoraException;
+import edu.unc.lib.dl.fedora.ObjectTypeMismatchException;
+import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.rdf.Cdr;
+import edu.unc.lib.dl.rdf.PcdmModels;
+
+public class CollectionObject extends ContentContainerObject {
+
+	public CollectionObject(PID pid, Repository repository, RepositoryObjectDataLoader dataLoader) {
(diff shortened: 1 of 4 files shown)


bbpennel:
Change PR destination to fcrepo4

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/CollectionObject.java
```diff
@@ -0,0 +1,51 @@
...
+	public CollectionObject(PID pid, Repository repository, RepositoryObjectDataLoader dataLoader) {
+		super(pid, repository, dataLoader);
+	}
+
+	@Override
+	public CollectionObject validateType() throws FedoraException {
+		if (!isType(Cdr.Collection.getURI())) {
+			throw new ObjectTypeMismatchException("Object " + pid + " is not a Collection.");
+		}
+		if(!isType(PcdmModels.Object.getURI())) {
```
bbpennel: Add space after if

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/CollectionObjectTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+		pid = PIDs.get(UUID.randomUUID().toString());
+
+		collection = new CollectionObject(pid, repository, dataLoader);
+
+		folderChildPid = PIDs.get(UUID.randomUUID().toString());
+		when(repository.mintContentPid()).thenReturn(folderChildPid);
+		
+		workChildPid = PIDs.get(UUID.randomUUID().toString());
+		when(repository.mintContentPid()).thenReturn(workChildPid);
```
bbpennel: I don't think you need both of these `when(repository.mintContentPid()).thenReturn(...);` the second one will overwrite the first one.  If you want to repository.mindContentPid return two different specific pids in the same test, then you can chain like so: `when(repository.mintContentPid()).thenReturn(folderChildPid).thenReturn(workChildPid);` where you have one thenReturn per call (and I believe the last value will repeat for every call after the second))

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/CollectionObjectTest.java
```diff
@@ -0,0 +1,140 @@
...
+		assertTrue("Incorrect type of child added", child instanceof FolderObject);
+		assertEquals("Child did not have the expected pid", folderChildPid, child.getPid());
+	}
+
+	// should not be able to add a Collection object as a member
+	@Test(expected = ObjectTypeMismatchException.class)
+	public void addCollectionObjectMemberTest() {
+		CollectionObject childObj = new CollectionObject(workChildPid, repository, dataLoader);
+
+		when(repository.createCollectionObject(any(PID.class), any(Model.class)))
```
bbpennel: I can't run the test at the moment, but I don't think this line is doing anything in this test since repository.createCollectionObject is not used in collection.addMember.  The repository create methods would only get called from the CollectionObject class if we had addFolder and addWork methods like with FolderObject, but I'm not clear yet on if we need those, so I'm okay that you left them out of CollectionObject.

harringj:
Thanks for the feedback, @bbpennel ! I made the changes you suggested, and this should be ready for final review/merging.

PR #534, opened 2016-11-04
updates properties in VocabularyEnforcementJob and related unit tests

harringj:
Updates properties with those in the new namespace.

### DIFF
--- a/deposit/src/test/java/edu/unc/lib/deposit/normalize/AbstractNormalizationJobTest.java
+++ b/deposit/src/test/java/edu/unc/lib/deposit/normalize/AbstractNormalizationJobTest.java
@@ -15,8 +15,6 @@
  */
 package edu.unc.lib.deposit.normalize;
 
-import static edu.unc.lib.deposit.work.DepositGraphUtils.cdrprop;
-import static edu.unc.lib.dl.util.ContentModelHelper.CDRProperty.sourceMetadata;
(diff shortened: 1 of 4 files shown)


deposit/src/test/java/edu/unc/lib/deposit/normalize/AbstractNormalizationJobTest.java
```diff
@@ -100,10 +97,10 @@ protected File verifyStagingLocationExists(Resource resource, Property stagingLo
 
 	protected void verifyMetadataSourceAssigned(Model model, Resource primaryResource, File depositDirectory,
 			String sourceType, String fileSuffix) {
-		Property stagingLoc = dprop(model, stagingLocation);
-		Property hasSourceMetadata = cdrprop(model, hasSourceMetadataProfile);
-		Property sourceMD = cdrprop(model, sourceMetadata);
-		Property hasDS = dprop(model, hasDatastream);
+		Property stagingLoc = CdrDeposit.stagingLocation;
+		Property hasSourceMetadata = Cdr.hasSourceMetadataProfile;
+		Property sourceMD = RDF.type;
```
bbpennel: At line 109 all that is being tested now is that the object has an RDF.type, not that it has the specific type we want, Cdr.SourceMetadata.  Also, i think it's a bit confusing to be assigning RDF.type to a variable named sourceMD, I would just remove sourceMD and use RDF.type directly

PR #535, opened 2016-11-04
Admin UI header

bbpennel:


harringj:
One suggestion I have for improving the state of our CSS code base is that we store color hex-values in variables. Doing so would make our code more readable and maintainable.  I'm not insisting that it happen for this PR, but perhaps we should create a technical debt ticket for our backlog?

bbpennel:
That's a good idea to keep in mind for the future, although I don't think all browsers (IE, Edge and some slightly old versions of Firefox http://caniuse.com/#feat=css-variables ) support it yet, so I think we may have to hold off for the moment.

PR #536, opened 2016-11-09
Add mods to content objects

harringj:


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -26,16 +29,50 @@
...
+		repository.createRelationship(pid, IanaRelation.describedby, mods.getResource());
+		
+		return fileObj;
+	}
+	
+	public FileObject addDescription(InputStream sourceMdStream, String sourceProfile,
+			InputStream modsStream) {
+		fileObj = createFileObject();
+		
+		BinaryObject orig = fileObj.addOriginalFile(sourceMdStream, null, "text/plain", null);
```
bbpennel: In this case, the original file is a source metadata file, so in addition to its normal types, it should also be assigned Cdr.SourceMetadata as an RDF.type

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -26,16 +29,50 @@
...
 	protected ContentObject(PID pid, Repository repository, RepositoryObjectDataLoader dataLoader) {
 		super(pid, repository, dataLoader);
 	}
 
-	public BinaryObject addDescription(InputStream contentStream) {
-		return null;
+	public FileObject addDescription(InputStream modsStream) {
+		fileObj = createFileObject();
+		
+		BinaryObject mods = fileObj.addOriginalFile(modsStream, null, "text/xml", null);
```
bbpennel: Add a pcdm:hasRelatedObject relationship from this contentObject to the description fileObject.  See image_1 in the diagram for an example.  The purpose of the hasRelatedObject is to indicate ownership from the content object to everything in the description object, while the iana:describedby is there to point out specifically which part of the description object we are using to describe the contentObject.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -31,11 +37,55 @@ protected ContentObject(PID pid, Repository repository, RepositoryObjectDataLoad
 		super(pid, repository, dataLoader);
 	}
 
-	public BinaryObject addDescription(InputStream contentStream) {
-		return null;
+	public FileObject addDescription(InputStream modsStream) {
```
bbpennel: We are going to need a way to put new versions of the description in to support the edit description page.  My feel is that this should be done using the versioning api, but i haven't researched how that would work yet.  So given our current focus on the deposit application i think it's alright to just cover adding new descriptions, just something to keep in mind for later.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -31,11 +37,55 @@ protected ContentObject(PID pid, Repository repository, RepositoryObjectDataLoad
...
+		FileObject fileObj = createFileObject();
+		
+		BinaryObject mods = fileObj.addOriginalFile(modsStream, null, "text/xml", null);
+		repository.createRelationship(pid, PcdmModels.hasRelatedObject, fileObj.getResource());
+		repository.createRelationship(pid, Cdr.hasMods, mods.getResource());
+		return fileObj;
+	}
+	
+	public FileObject addDescription(InputStream sourceMdStream, String sourceProfile,
```
bbpennel: Could you add javadoc for the new public methods?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -31,11 +37,55 @@ protected ContentObject(PID pid, Repository repository, RepositoryObjectDataLoad
...
+		BinaryObject mods = fileObj.addDerivative(null, modsStream, null, "text/plain", null);
+		repository.createRelationship(pid, Cdr.hasMods, mods.getResource());
+		
+		return fileObj;
 	}
 
-	public BinaryObject getDescription() {
-		return null;
+	public FileObject getDescription() {
+		Resource sourceMd = this.getResource();
```
bbpennel: I'd suggest renaming this variable since sourceMd implies source metadata in this context, but its actually just the resource for the current contentObject.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -31,11 +37,55 @@ protected ContentObject(PID pid, Repository repository, RepositoryObjectDataLoad
...
+		if (s != null) {
+			PID fileObjPid = PIDs.get(s.getResource().getURI());
+			return repository.getFileObject(fileObjPid);
+		} else {
+			return null;
+		}
+	}
+	
+	public BinaryObject getMODS() {
+		Resource mods = this.getResource();
```
bbpennel: suggest renaming this too, since it is also just the resource for the current contentObject.

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/WorkObjectIT.java
```diff
@@ -138,4 +140,21 @@ public void addPrimaryObjectAndSupplements() throws Exception {
...
+	public void addModsTest() throws Exception {
+		WorkObject obj = repository.createWorkObject(pid);
+		InputStream modsStream = new FileInputStream(new File("src/test/resources/vocabTest.xml"));
+		obj.addDescription(modsStream);
+		assertObjectExists(obj.getMODS().getPid());
+	}
+	@Test
+	public void addSourceMdTest() throws Exception {
+		WorkObject anotherObj = repository.createWorkObject(pid);
+		String sourceProfile = "some source md";
```
bbpennel: should assert that the profile gets set and that the source MD object exists.  I'd also recommend checking that content makes it into the MODS and source md files, similar to how the addDataFileTest is checking the binary content.

metadata/src/main/java/edu/unc/lib/dl/util/URIUtil.java
```diff
@@ -52,14 +52,16 @@ public static String join(String base, String... segments) {
...
 
-		for (String segment : segments) {
-			if (segment.charAt(0) != '/' && pathBuilder.length() > 0) {
-				pathBuilder.append('/');
-			}
-			if (segment.charAt(segment.length() - 1) == '/') {
-				pathBuilder.append(segment.substring(0, segment.length() - 1));
-			} else {
-				pathBuilder.append(segment);
+		if (segments[0] != null) {
```
bbpennel: This is to catch the case were someone calls UriUtil.join(base, null) I take it?  I wonder if instead we should do something like:
```
for (String segment : segments) {
    if (segment[i] == null) {
        continue;
    }
    ...
```

It would be helpful to cover this in the UriUtil unit test as well

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -31,11 +37,55 @@ protected ContentObject(PID pid, Repository repository, RepositoryObjectDataLoad
...
+	}
+	
+	public FileObject addDescription(InputStream sourceMdStream, String sourceProfile,
+			InputStream modsStream) {
+		FileObject fileObj = createFileObject();
+		
+		BinaryObject orig = fileObj.addOriginalFile(sourceMdStream, null, "text/plain", null);
+		repository.createRelationship(orig.getPid(), RDF.type, Cdr.SourceMetadata);
+		repository.createRelationship(pid, PcdmModels.hasRelatedObject, fileObj.getResource());
+		orig.getResource().addProperty(Cdr.hasSourceMetadataProfile, sourceProfile);
```
harringj: Just to clarify, is the behavior changing the local model rather than persisting to Fedora what we want here?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
harringj: Done

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
harringj: Done

metadata/src/main/java/edu/unc/lib/dl/util/URIUtil.java
harringj: You are correct. This was a bug I discovered that was leading to a NPE. The varargs construction takes the null value passed in and stores it in segments[0].

Just to help me understand this better, what are some cases you think would be covered by your coding suggestion versus what I have now?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
bbpennel: No, sorry, we want to add a relationship to the object in fedora, i was trying to point out that the relation wasn't making it to fedora.

metadata/src/main/java/edu/unc/lib/dl/util/URIUtil.java
bbpennel: `UriUtil.join(base, "path", null, "to", "stuff")`

harringj: Thanks! I wrote a test for that case and finally got it to pass :-)

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
harringj: Done! I added a createProperty() method to Repository to handle the source profile.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
harringj: Done!

harringj:
@ben, I believe I addressed all of your concerns. This should be ready for review again :-)

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/ContentObject.java
```diff
@@ -31,11 +40,79 @@ protected ContentObject(PID pid, Repository repository, RepositoryObjectDataLoad
...
+		repository.createRelationship(pid, Cdr.hasMods, mods.getResource());
+		return fileObj;
+	}
+	
+	/**
+	 * Adds source metadata and MODS to the object by creating the necessary relations in Fedora
+	 * @param sourceMdStream
+	 * @param sourceProfile
+	 * @param modsStream
+	 * @return a FileObject containing BinaryObejcts for source metadata and MODS
```
bbpennel: BinaryObjects

Also, for the description of this method, could you make it something like "Adds description information to this object, containing a source metadata and a MODS record derived from it."  

It may be helpful to also state what the sourceProfile is, so something like "identifies the encoding, profile, and/or origins of the sourceMdStream using an identifier defined in edu.unc.lib.dl.util.MetadataProfileConstants".  I just figure the details of how source metadata could be confusing to future generations.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
```diff
@@ -515,6 +515,27 @@ public void createRelationship(PID subject, Property property, Resource object)
 			throw ClientFaultResolver.resolve(e);
 		}
 	}
+	
+	/**
```
bbpennel: Apparently I failed to add a description to createRelationship, oops!  Could you add a short description to createProperty and createRelationship?

Also, since the two methods are identical past the first line, you may want to abstract out a private method for them both.

PR #540, opened 2016-11-18
Deposit records have ingested objects

harringj:


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestDepositRecordJob.java
```diff
@@ -95,7 +99,20 @@ public void runJob() {
...
-			// TODO add references to deposited objects 
+			// Add references to deposited objects 
+			Bag depositBag = dModel.getBag(getDepositPID().getRepositoryPath());
+			List<Resource> children = new ArrayList<Resource>();
+			DepositGraphUtils.walkObjectsDepthFirst(depositBag, children);
+			
+			Model triples = ModelFactory.createDefaultModel();
+			List<Statement> relations = new ArrayList<Statement>();
+			for (Resource child : children) {
+				relations.add(triples.createStatement(depositRecord.getResource(),
```
bbpennel: I don't think you need to go through the intermediate `relations` array here.  Prior to this loop I'd suggest doing a `triples.createResource(depositUri)` and then have the loop directly add to that resource.

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/DepositRecordIT.java
```diff
@@ -172,6 +175,21 @@ public void addPremisEventsTest() throws Exception {
 		// Check that the second event has the right type
 		assertTrue(events.get(1).getResource().hasProperty(Premis.hasEventType, Premis.Ingestion));
 	}
+	
+	public void addObjectsTest() throws Exception {
```
bbpennel: This test isn't annotated as a @Test, so it is not running

bbpennel: The test is throwing an exception with a 400 response from fedora.  It might be because the model.createResource() calls are using weird auto-generated IDs and because those ids don't exist in fedora, but we are trying to reference them.  Or the sparql syntax could be off.  So many options!

.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "staging-areas"]
```
harringj: Hey, @bbpennel, is this what you wanted? I had already rebased my branch off of fcrepo4, and I had added this to the .gitignore file, so I'm not sure why continues to show up uninvited.

deposit/src/main/java/edu/unc/lib/deposit/work/DepositGraphUtils.java
```diff
@@ -7,43 +7,17 @@
...
-import edu.unc.lib.dl.util.ContentModelHelper.FedoraProperty;
-import edu.unc.lib.dl.util.ContentModelHelper.Relationship;
 
 public class DepositGraphUtils {
 
 	public DepositGraphUtils() {
 		// TODO Auto-generated constructor stub
 	}
 
-	public static Property dprop(Model m, DepositRelationship r) {
```
harringj: Ok, sorry! Done!

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/DepositRecord.java
```diff
@@ -95,22 +97,27 @@ public BinaryObject getManifest(PID pid) throws FedoraException {
...
-
-		List<PID> pids = new ArrayList<>();
-		while (containsIt.hasNext()) {
-			String path = containsIt.next().getObject().toString();
-			pids.add(PIDs.get(path));
+	public List<PID> listManifests() throws FedoraException {
+		return addPidsToList(Cdr.hasManifest);
+	}
+	
+	public DepositRecord addIngestedObjects(PID depositPID, List<Resource> children) {
```
harringj: Done!

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/DepositRecordIT.java
harringj: What would you suggest using for IDs? Or can I somehow get those atuo-generated ones into Fedora, so that it will know about them?

PR #549, opened 2017-01-12
adds a PremisLoggerFactory class and refactors client code to use its createPremisLogger() methods

harringj:
Adds a factory class for PremisLogger to facilitate testing

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/normalize/AbstractMETS2N3BagJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/normalize/AbstractMETS2N3BagJob.java
@@ -99,7 +99,7 @@ public abstract class AbstractMETS2N3BagJob extends AbstractDepositJob {
 				div.setAttribute("CONTENTIDS", sb.toString());
 			}
 			
-			PremisLogger premisLogger = getPremisLogger(pid);
+			PremisLogger premisLogger = premisLoggerFactory.createPremisLogger(pid);
 			Resource premisEvent = premisLogger.buildEvent(Premis.Normalization)
 					.addEventDetail("Assigned PID, {0}, to object defined in a METS div", pid)
 					.addSoftwareAgent(SoftwareAgent.depositService.getFullname())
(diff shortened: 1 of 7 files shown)


deposit/src/main/java/edu/unc/lib/deposit/normalize/AbstractMETS2N3BagJob.java
```diff
@@ -99,7 +99,7 @@ protected void assignPIDs(Document mets) {
 				div.setAttribute("CONTENTIDS", sb.toString());
 			}
 			
-			PremisLogger premisLogger = getPremisLogger(pid);
```
bbpennel: I would suggest changing the behavior of `getPremisLogger` to call `premisLoggerFactory.createPremisLogger` rather than replacing its usage.  Doing so would update all of our deposit jobs.  Additionally, `getPremisLogger` actually calls `new FilePremisLogger(pid, file, repository);`, passing in the file is pretty critical in this case

bbpennel: actually it looks like you did update getPremisLogger, so I'd say to just use that and roll back the changes to AbstractMETS2N3BagJob and CDRMETS2N3BagJob

deposit/src/main/java/edu/unc/lib/deposit/work/AbstractDepositJob.java
```diff
@@ -52,7 +52,10 @@
 
 	@Autowired
 	protected Repository repository;
-
+	
+	@Autowired
```
bbpennel: This looks good, I'll just mention that the spring context files will need to be updated to instantiate the PremisLoggerFactory to inject here.

harringj: Yep, sorry, that was definitely on my radar, but forgot to add it to the PR!

deposit/src/main/java/edu/unc/lib/deposit/normalize/AbstractMETS2N3BagJob.java
harringj: Ok, sounds good

fcrepo-clients/src/main/java/edu/unc/lib/dl/event/PremisLoggerFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * A factory class for creating PremisLogger instances
+ * 
+ * @author harring
+ */
+
+public class PremisLoggerFactory {
+	
+	public PremisLogger createPremisLogger(PID pid) {
```
bbpennel: You can probably leave out this method, and we may actually want to remove the `FilePremisLogger(pid)` constructor since its not used and doesn't make a lot of sense for the file logger to be constructed without a file

harringj: Consider it deleted :-)

PR #550, opened 2017-01-12
Jedis try blocks

bbpennel:


metadata/src/main/java/edu/unc/lib/dl/util/DepositStatusFactory.java
```diff
@@ -97,95 +112,91 @@ public void set(String depositUUID, DepositField field, String value) {
...
-		Jedis jedis = getJedisPool().getResource();
-		String state = jedis.hget(DEPOSIT_STATUS_PREFIX+depositUUID, DepositField.state.name());
-		try {
+		
+		try (Jedis jedis = getJedisPool().getResource()) {
+			String state = jedis.hget(DEPOSIT_STATUS_PREFIX + depositUUID, DepositField.state.name());
 			result = DepositState.valueOf(state);
-		} catch(NullPointerException e) {
-		} catch(IllegalArgumentException e) {
+		} catch (IllegalArgumentException | NullPointerException e) {
```
harringj: I know it's how the code was before your changes, but do we really want to be catching a NPE here? Seems like testing for null and then having a path where the code takes appropriate action might be better.

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
```diff
@@ -79,64 +79,59 @@ public void clearStale(String depositUUID) {
...
 	public void completed(String jobUUID) {
-		Jedis jedis = getJedisPool().getResource();
-		jedis.hset(JOB_STATUS_PREFIX + jobUUID,
-				JobField.status.name(), JobStatus.completed.name());
-		jedis.hset(JOB_STATUS_PREFIX + jobUUID,
-				JobField.endtime.name(),
-				String.valueOf(System.currentTimeMillis()));
-		getJedisPool().returnResource(jedis);
+		try (Jedis jedis = getJedisPool().getResource()) {
+			jedis.hset(JOB_STATUS_PREFIX + jobUUID,
```
harringj: Seems like there might be some extra indents here

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
```diff
@@ -79,64 +79,59 @@ public void clearStale(String depositUUID) {
...
 
 	public void killed(String jobUUID) {
-		Jedis jedis = getJedisPool().getResource();
-		jedis.hset(JOB_STATUS_PREFIX + jobUUID,
-				JobField.status.name(), JobStatus.killed.name());
-		jedis.hset(JOB_STATUS_PREFIX + jobUUID,
-				JobField.endtime.name(),
-				String.valueOf(System.currentTimeMillis()));
-		getJedisPool().returnResource(jedis);
+		try (Jedis jedis = getJedisPool().getResource()) {
```
harringj: Ditto here

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
```diff
@@ -79,64 +79,59 @@ public void clearStale(String depositUUID) {
...
+					String.valueOf(System.currentTimeMillis()));
+		}
 	}
 
 	public void incrCompletion(String jobUUID, int amount) {
-		Jedis jedis = getJedisPool().getResource();
-		jedis.hincrBy(JOB_STATUS_PREFIX + jobUUID,
-				JobField.num.name(), amount);
-		getJedisPool().returnResource(jedis);
+		try (Jedis jedis = getJedisPool().getResource()) {
```
harringj: And here

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
```diff
@@ -198,18 +186,14 @@ public String getJobByStatus(String depositUUID, JobStatus status) {
...
 
 		return null;
 	}
 
 	public List<String> getJobsByStatus(String depositUUID, JobStatus status) {
-		Jedis jedis = getJedisPool().getResource();
-
-		List<String> result = new ArrayList<String>();
-		try {
+		List<String> result = new ArrayList<>();
```
harringj: Just curious about why you reverted to using the raw ArrayList type here. I'm sure there's a good reason, :-)

metadata/src/main/java/edu/unc/lib/dl/util/DepositStatusFactory.java
bbpennel: Yeah both of these are rather weird cases to be catching, and we aren't generally handling null cases this way in the rest of the code base, so i'll switch that.  I'm not really sure how the IllegalArgumentExceptions would have occurred.

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
bbpennel: This code was actually indented too far before, I shifted it back to the left by one indent.

bbpennel: Actually sorry I was thinking of DepositStatusFactory.addManifest.  I think the indent is correct here though

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
bbpennel: Because we hates them.

PR #560, opened 2017-02-08
Remove last modfied date of folders from the UI

lfarrell:


bbpennel:
Remove date from being displayed on the full record page for collections and folders, but retain it on works and files.

PR #563, opened 2017-02-10
Log spam

lfarrell:


fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
```diff
@@ -339,6 +340,28 @@ public Datastream getDatastream(PID pid, String dsID, String asOfDateTime) throw
...
+
+		try (CloseableHttpResponse httpResp = httpClient.execute(method)) {
+			int statusCode = httpResp.getStatusLine().getStatusCode();
+
+			if (statusCode == 200) {
+				return true;
+			}
+
+			return false;
+		} catch (Exception e) {
```
harringj: Do we want to log any exceptions that might get thrown when execute() fails?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
```diff
@@ -339,6 +340,28 @@ public Datastream getDatastream(PID pid, String dsID, String asOfDateTime) throw
...
+		builder.pathSegment(pid.getPid());
+		builder.pathSegment("datastreams");
+		builder.pathSegment(dsID);
+
+		HttpHead method = new HttpHead(builder.build().encode().toUriString());
+
+		try (CloseableHttpResponse httpResp = httpClient.execute(method)) {
+			int statusCode = httpResp.getStatusLine().getStatusCode();
+
+			if (statusCode == 200) {
```
harringj: If we just want to know whether a ds exists, could there be cases where another 2xx code is returned that we would want to return true, or is 200 really the only signal of success here?

lfarrell: Seemed like 200 is the only one return, but could check for any 2xx code

fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
lfarrell: I'll leave that to Ben.

bbpennel: Yes that sounds like a good idea

lfarrell: Added logging on error. The most common we got was 500 for a malformed request to fedora.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
lfarrell: Looking at HTTP response codes it doesn't seem like the others would be relevant.

harringj: Cool, thanks for checking on that!

fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ManagementClient.java
harringj: 👍

PR #565, opened 2017-02-21
Cdr 1665 encapsulate obj ingest into txs

harringj:


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -396,40 +409,43 @@ private void ingestWork(ContentContainerObject parent, Resource parentResc, Reso
 		boolean skip = skipResumed(childResc);
 		if (skip) {
 			obj = repository.getWorkObject(childPid);
+			ingestChildren(obj, childResc);
+
+			// Avoid adding primaryObject relation for a resuming deposit if already present
+			if (!obj.getResource().hasProperty(Cdr.primaryObject)) {
+				// Set the primary object for this work if one was specified
+				Statement primaryStmt = childResc.getProperty(Cdr.primaryObject);
```
bbpennel: This refactoring looks good!  You may want to consider making a addPrimaryObject helper method in the test that does lines 417-421 so it doesn't have to be duplicated below, but i'll leave that up to your discretion

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -275,8 +279,10 @@ private void ingestFileObjectAsWork(ContentContainerObject parent, Resource pare
 		}
 		Resource workResc = workModel.createResource(workPid.getRepositoryPath());
 		workResc.addProperty(DC.title, label);
-
-		WorkObject newWork = repository.createWorkObject(workPid, workModel);
+		FedoraTransaction tx = repository.startTransaction();
+		try {
+			WorkObject newWork = repository.createWorkObject(workPid, workModel);
```
harringj: Done!

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
harringj: Good suggestion. Done!

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -391,6 +392,35 @@ public void ingestDeepHierarchy() {
...
+		Model model = job.getWritableModel();
+		Bag depBag = model.createBag(depositPid.getRepositoryPath());
+
+		// Constructing the folder in the deposit model with a title
+		PID workPid = repository.mintContentPid();
+		Bag workBag = model.createBag(workPid.getRepositoryPath());
+		
+		// Add incorrect resource type to bag, which causes a DepositException to be thrown
+		// and the tx to be cancelled
+		workBag.addProperty(RDF.type, Cdr.FileObject);
```
bbpennel: I think this should have type Cdr.WorkObject instead of FileObject.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -391,6 +392,35 @@ public void ingestDeepHierarchy() {
...
+		PID mainPid = addFileObject(workBag, FILE1_LOC, FILE1_MIMETYPE, FILE1_SHA1);
+
+		depBag.add(workBag);
+
+		workBag.asResource().addProperty(Cdr.primaryObject,
+				model.getResource(mainPid.getRepositoryPath()));
+
+		job.closeModel();
+
+		job.run();
```
bbpennel: One other suggestion, it would be a bit reassuring to make sure that the transaction has ended and that the work does not exist in fedora after job.run() has completed, just to ensure it did the work of cancelling the tx

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FedoraTransaction.java
```diff
@@ -73,6 +72,23 @@ public void keepAlive() {
 		repo.keepTransactionAlive(txUri);
 	}
 	
+	public void cancel(Throwable t) {
```
bbpennel: The two cancel methods are basically the same except for the rethrowing bit.  Since cancelling actually involves a fair number of steps, maybe we could refactor this so that cancel() just calls cancel(t), passing it null for the exception?

harringj: Yep, I like it :-)

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
harringj: As we discussed, the "wrong" type is there to provoke the tx cancellation.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
harringj: I added assertions to ease your worried mind

PR #566, opened 2017-02-23
Cdr 1702 remove links

harringj:


bbpennel:
I believe the full record controller is still making a call to solr to retrieve facet results for aggregates (which is how it was getting the format field), which it is no longer displayed.  Could you try disabling that just so we're not doing more calls to solr than are needed?

https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/91581a1df3af3ee9a76c847e971fd41bc5f14a01/access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java#L170

I believe you could change that filter to check if the object in question is a Collection, otherwise don't retrieve facets.

PR #567, opened 2017-02-27
FITS Deposit Job

bbpennel:
Adds a technical metadata extraction job to the deposit pipeline (after virus scanning), along with a unit test.  Premis v3 namespace used for storing FITS report.

Refactors common staging location construction and property lookup functionality into shared methods.

Changes FcrepoClientFactory into a singleton with non-static methods.


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FcrepoClientFactory.java
```diff
@@ -24,11 +24,24 @@
...
 	}
+	
+	public static FcrepoClientFactory factory(String baseUri) {
+		if (instance == null) {
+			instance = new FcrepoClientFactory(baseUri);
+		}
+		return instance;
+	}
+
+	public static FcrepoClientFactory defaultFactory() {
```
harringj: What's the purpose of the defaultFactory() method? How does the instance's baseUri get set? Could the method return null?

bbpennel: I was having a hard time figuring out how to best handle that situation.  In this implementation, when you create a FcrepoClientFactory and provide a baseUri it stores the constructed factory as the singleton instance of that factory.  You can then call FcrepoClientFactory.defaultFactory() elsewhere in the code and not have to provide a baseUri since it is just returning the originally constructed copy.  But it requires FcrepoClientFactory.factory(baseUri) to have been called first, otherwise .defaultFactory would return null, which isn't so great.

PR #568, opened 2017-03-01
removes deposits set and replaces it with a key set

harringj:


metadata/src/main/java/edu/unc/lib/dl/util/DepositStatusFactory.java
```diff
@@ -89,7 +88,6 @@ public void addManifest(String depositUUID, String value) {
 	public void save(String depositUUID, Map<String, String> status) {
 		try (Jedis jedis = getJedisPool().getResource()) {
 			jedis.hmset(DEPOSIT_STATUS_PREFIX + depositUUID, status);
```
bbpennel: Could you also remove the declaration of DEPOSIT_SET from the constants class since it is no longer used anywhere?

metadata/src/main/java/edu/unc/lib/dl/util/DepositStatusFactory.java
```diff
@@ -59,9 +58,9 @@ public DepositStatusFactory() {}
 	public Set<Map<String, String>> getAll() {
 		Set<Map<String, String>> result = new HashSet<>();
 		try (Jedis jedis = getJedisPool().getResource()) {
-			Set<String> deposits = jedis.smembers(DEPOSIT_SET);
-			if (deposits != null) {
-				for(String uuid : deposits) {
+			Set<String> keys = jedis.hkeys(DEPOSIT_STATUS_PREFIX +"*");
```
bbpennel: space after + :)

metadata/src/main/java/edu/unc/lib/dl/util/DepositStatusFactory.java
```diff
@@ -59,9 +58,9 @@ public DepositStatusFactory() {}
...
+			if (keys != null) {
+				for(String uuid : keys) {
```
bbpennel: Might as well add the space after for while I'm being pedantic, although I know it was already this way.

PR #570, opened 2017-03-02
Fcrepo4 fulltext

lfarrell:


services-camel/src/main/java/edu/unc/lib/cdr/FulltextProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+		this.slug = slug;
+		this.fileSuffix = fileSuffix;
+	}
+
+	@Override
+	public void process(Exchange exchange) throws Exception {
+		final Message in = exchange.getIn();
+
+		String binaryUri = (String) in.getHeader(FCREPO_URI);
+		String binaryMimeType = (String) in.getHeader(CdrBinaryMimeType);
```
harringj: Where are these params defined?

services-camel/src/main/java/edu/unc/lib/cdr/fulltext/FulltextRouter.java
```diff
@@ -0,0 +1,46 @@
+package edu.unc.lib.cdr.fulltext;
```
harringj: Again, include license :-)

services-camel/src/main/java/edu/unc/lib/cdr/fulltext/FulltextRouter.java
```diff
@@ -0,0 +1,46 @@
...
+public class FulltextRouter extends RouteBuilder {
+	@BeanInject(value = "binaryMetadataProcessor")
+	private BinaryMetadataProcessor mdProcessor;
+	
+	@BeanInject(value = "fulltextProcessor")
+	private FulltextProcessor ftProcessor;
+	
+	
+	public void configure() throws Exception {
+		//String mimetypePattern = "^(text/|application/pdf|application/msword|application/vnd\\.|application/rtf|application/powerpoint|application/postscript).*)";
```
harringj: Are we leaving this alternate pattern here for a reason?

services-camel/src/main/java/edu/unc/lib/cdr/thumbnail/ThumbnailRouter.java
```diff
@@ -53,11 +53,11 @@ public void configure() throws Exception {
 			.to("fcrepo:{{fcrepo.baseUrl}}?preferInclude=ServerManaged&accept=text/turtle")
 			.process(mdProcessor)
 			.to("direct:images");
-			
+		
 		from("direct:images")
 		.routeId("IsImage")
-		.filter(simple("${headers[MimeType]} regex '^application.*?$'"))
-			//.filter(simple("${headers[MimeType]} regex '^(image.*?$|application.*?(photoshop|psd)$)'"))
+		//.filter(simple("${headers[MimeType]} regex '^application.*?$'"))
```
harringj: Is this being left here for troubleshooting?

services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
```diff
@@ -0,0 +1,85 @@
+package edu.unc.lib.cdr;
```
harringj: License

services-camel/src/main/java/edu/unc/lib/cdr/fulltext/FulltextRouter.java
lfarrell: Can probably be changed now that FITS is included. Before only application/octet-stream made it through

services-camel/src/main/java/edu/unc/lib/cdr/fulltext/FulltextRouter.java
```diff
@@ -0,0 +1,46 @@
...
+import static edu.unc.lib.dl.rdf.Fcrepo4Repository.Binary;
+
+import org.apache.camel.BeanInject;
+import org.apache.camel.LoggingLevel;
+import org.apache.camel.builder.RouteBuilder;
+
+import edu.unc.lib.cdr.BinaryMetadataProcessor;
+import edu.unc.lib.cdr.FulltextProcessor; 
+
+public class FulltextRouter extends RouteBuilder {
```
bbpennel: Also add class level javadoc.

services-camel/src/main/java/edu/unc/lib/cdr/FulltextProcessor.java
```diff
@@ -0,0 +1,83 @@
+package edu.unc.lib.cdr;
```
harringj: No worries, I used to, as well, until Ben set me straight :-)

services-camel/src/main/java/edu/unc/lib/cdr/FulltextProcessor.java
harringj: Wondering whether we should change those header names to be in ALL_CAPS_SEPARATED_BY_UNDERSCORES to follow Java conventions. I had thought they were variable names, not constants :-)

services-camel/src/test/java/edu/unc/lib/cdr/fulltext/FulltextRouterTest.java
```diff
@@ -0,0 +1,173 @@
...
+public class FulltextRouterTest extends CamelSpringTestSupport {
+	private static final String EVENT_NS = "http://fedora.info/definitions/v4/event#";
+	private static final String EVENT_TYPE = "org.fcrepo.jms.eventType";
+	private static final String IDENTIFIER = "org.fcrepo.jms.identifier";
+	private static final String RESOURCE_TYPE = "org.fcrepo.jms.resourceType";
+	private static final long timestamp = 1428360320168L;
+	private static final String userID = "bypassAdmin";
+	private static final String userAgent = "curl/7.37.1";
+	private static final String fileID = "/file1";
+	private final String eventTypes = EVENT_NS + "ResourceCreation";
```
harringj: Why are these not static?

services-camel/src/test/java/edu/unc/lib/cdr/fulltext/FulltextRouterTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+import edu.unc.lib.dl.fcrepo4.PIDs;
+import edu.unc.lib.dl.fcrepo4.Repository;
+
+public class FulltextRouterTest extends CamelSpringTestSupport {
+	private static final String EVENT_NS = "http://fedora.info/definitions/v4/event#";
+	private static final String EVENT_TYPE = "org.fcrepo.jms.eventType";
+	private static final String IDENTIFIER = "org.fcrepo.jms.identifier";
+	private static final String RESOURCE_TYPE = "org.fcrepo.jms.resourceType";
+	private static final long timestamp = 1428360320168L;
```
harringj: What would you think about changing the names of these to be in all caps with underscores, instead of camelCase?

services-camel/src/main/java/edu/unc/lib/cdr/fulltext/FulltextRouter.java
bbpennel: Yes we should turn this on and make sure it works as expected now that we should be getting mimetypes.

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
```diff
@@ -29,7 +28,9 @@
 	public void configure() throws Exception {
 		from("{{fcrepo.stream}}")
 		.routeId("MetaServicesRouter")
-		.routingSlip(new SimpleExpression("direct-vm:index.start,direct-vm:createThumbnail"));
+		.to("direct-vm:index.start")
+		.multicast()
+		.to("direct-vm:createThumbnail","direct-vm:extractFulltext");
```
bbpennel: We don't need to figure it out for this PR, but I would still like to figure out if there's anything we can do about losing the messages midway through.  I may look at that some tomorrow

services-camel/src/main/java/edu/unc/lib/cdr/thumbnail/ThumbnailRouter.java
bbpennel: Yeah think its safe to remove this at this point

services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
```diff
@@ -0,0 +1,85 @@
...
+	@Test
+	public void validTest() throws Exception {
+		BufferedWriter writeFile = new BufferedWriter(new FileWriter(this.file));
+		writeFile.write(testText);
+		writeFile.close();
+		
+		String filePath = this.file.getAbsolutePath().toString();
+		
+		when(message.getHeader(eq(FCREPO_URI)))
+		.thenReturn("http://fedora/test/original_file");
```
bbpennel: double indent the second line here and below

services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
```diff
@@ -0,0 +1,85 @@
...
+		parent = mock(FileObject.class);
+
+		when(repository.getBinary(any(PID.class))).thenReturn(binary);
+		when(binary.getParent()).thenReturn(parent);
+
+		processor.process(exchange);
+
+		verify(message).equals(FileObject.class);
+
+		this.file.deleteOnExit();
```
bbpennel: Move this into the init method rather after the file is created to ensure that it always gets called.  Having it at the end of the method could mean that the file wouldn't get marked for deletion if the test didn't pass.

services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
```diff
@@ -0,0 +1,85 @@
...
+		
+		binary = mock(BinaryObject.class);
+		parent = mock(FileObject.class);
+
+		when(repository.getBinary(any(PID.class))).thenReturn(binary);
+		when(binary.getParent()).thenReturn(parent);
+
+		processor.process(exchange);
+
+		verify(message).equals(FileObject.class);
```
bbpennel: Instead we should test that addDerivative was called on the FileObject class, and that it got the expected text (aka tika did its job and we passed the result to fedora).  Should also verify that the other parameters sent to addDeriv are correct (filename, etc).

services-camel/src/test/java/edu/unc/lib/cdr/fulltext/FulltextRouterTest.java
```diff
@@ -0,0 +1,173 @@
+package edu.unc.lib.cdr.fulltext;
```
bbpennel: License

services-camel/src/test/java/edu/unc/lib/cdr/fulltext/FulltextRouterTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+		assertMockEndpointsSatisfied();
+	}
+	
+	@Test
+	public void testFullTextExtractionFilterValidMimeType() throws Exception {
+		getMockEndpoint("mock:direct:fulltext.extraction").expectedMessageCount(1);
+		createContext(extractionRoute);
+		
+		template.sendBodyAndHeaders("", createEvent(fileID, eventTypes));
```
bbpennel: After you turn on the real mimetype filter you'll want to give it a real valid mimetype here

services-camel/src/main/java/edu/unc/lib/cdr/FulltextProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	@Override
+	public void process(Exchange exchange) throws Exception {
+		final Message in = exchange.getIn();
+
+		String binaryUri = (String) in.getHeader(FCREPO_URI);
+		String binaryMimeType = (String) in.getHeader(CdrBinaryMimeType);
+		String binaryPath = (String) in.getHeader(CdrBinaryPath);
+		String derivativePath = binaryPath + fileSuffix;
```
lfarrell: Yeah, was wondering about that one.

services-camel/src/main/java/edu/unc/lib/cdr/FulltextProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+	}
+	
+	private String extractText(String filepath) throws IOException, SAXException, TikaException {
+		BodyContentHandler handler = new BodyContentHandler();
+
+		AutoDetectParser parser = new AutoDetectParser();
+		Metadata metadata = new Metadata();
+
+		try (InputStream stream = new FileInputStream(new File(filepath))) {
+			parser.parse(stream, handler, metadata);
```
lfarrell: Seems to work though :)

services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
lfarrell: Not sure how you'd actually do this as we only have access  to the message headers and body.

bbpennel: Take a look here:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/deposit/src/test/java/edu/unc/lib/deposit/validate/ExtractTechnicalMetadataJobTest.java#L231

you can use an argument captor to get back the inputstream that was passed to the mock object, then read the string from the inputstream and make sure it matches the original test text.  For the string based parameters you can just use the eq(mimetype) type matcher.  The `verify` method is a wonderful thing

lfarrell: I'm still pretty stumped on this one. I can set up a captor but it apparently never interacts with anything, though it should be mocked.

bbpennel:
Often if the methods you are trying to verify aren't getting called it
means that the code you're resting isn't working as expected.  I could take
a look if you push what you have.  Otherwise I suggest setting break points
in your processor to see if it is hitting the code it should be, and the
mock it should be

On Mar 6, 2017 8:24 AM, "Dean Farrell" <notifications@github.com> wrote:

*@lfarrell* commented on this pull request.
------------------------------

In services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
<https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/570#discussion_r104455515>
:

> +		
+		when(message.getHeader(eq(CdrBinaryPath)))
+				.thenReturn(filePath);
+		
+		when(message.getHeader(eq(CdrBinaryMimeType)))
+		.thenReturn("plain/text");
+		
+		binary = mock(BinaryObject.class);
+		parent = mock(FileObject.class);
+
+		when(repository.getBinary(any(PID.class))).thenReturn(binary);
+		when(binary.getParent()).thenReturn(parent);
+
+		processor.process(exchange);
+
+		verify(message).equals(FileObject.class);

I'm still pretty stumped on this one. I can set up a captor but it
apparently never interacts with anything, though it should be mocked.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/570#discussion_r104455515>,
or mute the thread

lfarrell:
Nope, makes it all the way through the processor. Dying here 
```java
ArgumentCaptor<FulltextProcessor> requestCaptor = ArgumentCaptor.forClass(FulltextProcessor.class);
verify(fulltext).process((Exchange) requestCaptor.capture());
List<FulltextProcessor> request = requestCaptor.getAllValues();
```
When trying to get values. I'll push it up.

services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
```diff
@@ -56,46 +66,70 @@
...
 		parent = mock(FileObject.class);
 
 		when(repository.getBinary(any(PID.class))).thenReturn(binary);
 		when(binary.getParent()).thenReturn(parent);
+		
+		
 
 		processor.process(exchange);
+		
+		ArgumentCaptor<FulltextProcessor> requestCaptor = ArgumentCaptor.forClass(FulltextProcessor.class);
```
bbpennel: fullText isn't a mock (and it shouldn't be since this unit test is testing that class (you are declaring it as a Mock and then assigning a real value in the init method)), so verifying this probably isn't doing anything.  

I can't see my comment anymore, but what you probably should be doing is verify(parent).addDerivative(...) with the argument captor.

services-camel/src/test/java/edu/unc/lib/cdr/FulltextProcessorTest.java
```diff
@@ -0,0 +1,121 @@
...
+		file = File.createTempFile(fileName, "txt");
+		file.deleteOnExit();
+		when(exchange.getIn()).thenReturn(message);
+		PIDs.setRepository(repository);
+		when(repository.getBaseUri()).thenReturn("http://fedora");
+	}
+	
+	@Test
+	public void extractFulltextTest() throws Exception {
+		try (ByteArrayOutputStream bos = new ByteArrayOutputStream()) {
```
bbpennel: I just realized that `bos` isn't used for anything.  You should remove that, and I'd suggest instead moving the declaration of `writeFile` into the try with section so it gets autoclosed

services-camel/src/main/resources/config.properties
```diff
@@ -9,6 +9,7 @@ fcrepo.authPassword=
 fcrepo.authHost=
 
 cdr.enhancement.thumbnail=PNG
+cdr.enhancement.fulltext=_fullText.txt
```
bbpennel: can you rename this property here and in its usage to "cdr.enhancement.fulltext.fileName", and also since its not a prefix anymore it should be formatted like the value you use in the test.  Should also rename `cdr.enhancement.thumbnail` to something that indicates what property it is.

services-camel/src/test/java/edu/unc/lib/cdr/fulltext/FulltextRouterTest.java
```diff
@@ -0,0 +1,189 @@
...
+	}
+	
+	@Test
+	public void testEventTypeFilter() throws Exception {
+		getMockEndpoint("mock:direct:fulltext.filter").expectedMessageCount(0);
+		
+		createContext(ENHANCEMENT_ROUTE);
+		
+		Map<String, Object> headers = createEvent(FILE_ID, EVENT_TYPES);
+		headers.put(EVENT_TYPE, "ResourceDeletion");
```
bbpennel: Might as well just pass this event type to `createEvent` instead of `EVENT_TYPES`.  Also, shouldn't the event type begin with `EVENT_NS`?

bbpennel: Apparently these are two separate headers, weird

services-camel/src/test/resources/config.properties
```diff
@@ -8,6 +8,9 @@ fcrepo.authUsername=
 fcrepo.authPassword=
 fcrepo.authHost=
 
+cdr.enhancement.thumbnail=PNG
+cdr.enhancement.fulltext=_fullText.txt
```
bbpennel: these will need to be updated too (even though they may not be currently used)

services-camel/src/test/java/edu/unc/lib/cdr/fulltext/FulltextRouterTest.java
bbpennel: If we are never changing the value passed to createEvent (EVENT_TYPES) and we aren't testing it in the class, you can probably remove it as a parameter to the createEvent method to avoid confusion like this.  I'm kind of wondering if we're supposed to be checking the other headers?  what do they exist for?

PR #575, opened 2017-03-21
Cdr 1729 ingest mods

harringj:


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -529,4 +536,16 @@ private NodeIterator getChildIterator(Resource resc) {
...
+	
+	private void addDescription(ContentObject obj) {
+		File modsFile = new File(getDescriptionDir(), obj.getPid().getUUID() + ".xml");
+		if (!modsFile.exists()) {
+			return;
+		}
+		try(InputStream modsStream = FileUtils.openInputStream(modsFile)) {
+			obj.addDescription(modsStream);
+		} catch(IOException e) {
+			// just eat the exception, or do we want some exception handling here?
```
bbpennel: Fix spacing on this try block.  Also, we don't want to eat the exception, we'd want the code calling this method to get the exception and figure out what to do with it

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -101,6 +102,12 @@ public void init() throws Exception {
 
 		techmdDir = new File(depositDir, TECHMD_DIR);
 		techmdDir.mkdir();
+		
+		File modsFolder = job.getDescriptionDir();
```
bbpennel: There's a lot of stuff going on in the ingestWorkObjectTest method.  How would you feel about making another test that is dedicated to testing addDescription?  It'd just be a very slimmed down version of the same thing you already did, with or without using a Work, up to you.  

Also explicitly test that another object that shouldn't get mods doesn't get it.

Also, if workPidForMods and the modsFolder are only used for one test you should probably just put them into the test that uses them.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -284,6 +288,7 @@ private void ingestFileObjectAsWork(ContentContainerObject parent, Resource pare
 			WorkObject newWork = repository.createWorkObject(workPid, workModel);
 
 			// TODO add the FileObject's description to the work instead
```
bbpennel: Not a TODO anymore.  Toodle-oo TODOs

harringj: Right, I was just leaving those in untilI was finished, so I could easily find the spots I was workingon

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
harringj: I figured we didn't want to eat it, but wasn't sure what you had in mind. I added a throws IOException to the header

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -171,7 +173,7 @@ public void runJob() {
 		// Ingest objects included in this deposit into the destination object
 		try {
 			ingestChildren((ContentContainerObject) destObj, depositBag);
-		} catch (DepositException | FedoraException e) {
+		} catch (DepositException | FedoraException | IOException e) {
```
bbpennel: One thing I noticed is that it seems like the IOException from addDescription will never make it back to this point.  It's immediately caught and rethrown as a transaction interruption exception everywhere except in addFileObject, which does the same one step up in the addWork* method that calls it.  So you may be able to skip having `throws IOException` in a lot of the cases added.

PR #578, opened 2017-03-30
Cdr 1697 plus sign bug

harringj:


access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -515,7 +516,8 @@ public void getChildrenCounts(List<BriefObjectMetadata> resultList, AccessGroupS
 		solrQuery.add("f." + ancestorPathField + ".facet.limit", Integer.toString(Integer.MAX_VALUE));
 		// Sort by value rather than count so that earlier tiers will come first in case the result gets cut off
 		solrQuery.setFacetSort("index");
-
+		LOG.warn("This is the one you're looking for!" + solrQuery.toString());
```
bbpennel: Remove

access/src/main/resources/log4j.properties
```diff
@@ -22,6 +22,7 @@ log4j.rootLogger=WARN, SERVER, CDRACCESS
 log4j.logger.org.apache.commons.httpclient=ERROR
 log4j.logger.httpclient=ERROR
 log4j.logger.org.springframework=ERROR
+log4j.logger.edu.unc.lib.dl.search.solr=DEBUG
```
bbpennel: Remove

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -515,7 +516,8 @@ public void getChildrenCounts(List<BriefObjectMetadata> resultList, AccessGroupS
...
+
```
harringj: I can remove this logging statement, unless you think we should leave something like this in there

access/src/main/resources/log4j.properties
harringj: Do we want to leave this change in here?

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -710,7 +712,7 @@ public HierarchicalBrowseResultResponse getHierarchicalBrowseResults(Hierarchica
 					this.generateSearch(filteredChildrenRequest));
 
 			this.getChildrenCounts(results.getResultList(), accessGroups, "containers",
-					"contentModel:" + SolrSettings.sanitize(ContentModelHelper.Model.CONTAINER.toString()),
+					"resourceType:" + ("(" + ResourceType.Collection.toString() + " OR " + ResourceType.Folder.toString() + ")"),
```
bbpennel: I find the ('s wrapping the "("'s kind of confusing.  Its totally cosmetic and up to you, but removing the outter ('s shouldn't cause any problems.

PR #579, opened 2017-04-04
Cdr 1740 add ingestion events

harringj:


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -560,12 +576,32 @@ private void addDescription(ContentObject obj) throws IOException {
...
 			return;
 		}
-		FilePremisLogger premisLogger = new FilePremisLogger(obj.getPid(), premisFile, repository);
+		PremisLogger premisLogger = getPremisLogger(obj.getPid());
 		List<PremisEventObject> events = premisLogger.getEvents();
+		// add premis events to fedora
 		obj.addPremisEvents(events);
 	}
+	
+	private void addIngestionEvent(ContentObject obj) throws IOException {
```
bbpennel: This method looks pretty good.  Looking at the spreadsheet a little more though I think there may need to be a some changes to the functionality, which is my fault for not working more through the requirements:

1) I don't think this works correctly for the ingestFolder case because the children are added after premis is uploaded, so obj.getMembers() wouldn't return anything.  There doesn't appear to be an ingestFolder integration test at the moment except the resume case.  But this ties in with another suggestion
2) Based on the spreadsheet, I think when a container is ingested in should have two details: "ingested as PID: <>" referring to itself, and "added child object <>" for the items added to it.  If no children are added, then I don't think it needs the second message
3) The "added child object <>" message should also be added to the existing container where the deposit was ingested to, since it records how many items were added to the existing container.

It may be helpful to split this method into two methods, one which records the ingestion event info for the object itself, and one that records ingestion of children.  These can be two separate events.

Also, in the case of adding to an existing container, obj.getMembers() wouldn't be accurate since there may already be existing objects in the container.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -197,23 +205,29 @@ private void ingestChildren(ContentContainerObject destObj, Resource parentResc)
...
+						obj = ingestFileObject(destObj, parentResc, childResc);
 					} else {
 						// File object is a standalone, so construct a Work around it
-						ingestFileObjectAsWork(destObj, parentResc, childResc);
+						obj= ingestFileObjectAsWork(destObj, parentResc, childResc);
 					}
 				}
+				if (obj != null) {
+					// add premis event for ingestion to each child object
+					addIngestionEventForChild(obj);
```
bbpennel: I appreciate the centralized code of this approach, but i think it is going to result in all the premis events being added twice for most objects since addPremisEvents is called twice and it always reads from the same file.  It may be hard to go with this approach without a larger code change

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -560,12 +586,47 @@ private void addDescription(ContentObject obj) throws IOException {
...
+		PremisLogger premisLogger = getPremisLogger(obj.getPid());
 		List<PremisEventObject> events = premisLogger.getEvents();
+		// add premis events to fedora
 		obj.addPremisEvents(events);
 	}
+	
+	private void addIngestionEventForContainer(ContentContainerObject obj, Model model) throws IOException {
+		PremisLogger premisLogger = getPremisLogger(obj.getPid());
+		PremisEventBuilder builder = premisLogger.buildEvent(Premis.Ingestion);
+		builder.addEventDetail("ingested as PID: {0}", obj.getPid().toString());
```
bbpennel: The container should only ever get one event with this message, the first time it is ingested.  I believe at the moment the destination container would get a new one of these everytime an ingest happens in it.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -560,12 +586,47 @@ private void addDescription(ContentObject obj) throws IOException {
...
+		ResIterator childrenIt = model.listSubjects();
```
bbpennel: Traversing the hierarchy of the deposit should follow the jena bag structure

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -231,24 +235,28 @@ private void ingestChildren(ContentContainerObject destObj, Resource parentResc)
...
 		// TODO add ACLs
 		WorkObject work = (WorkObject) parent;
 		FileObject obj = addFileToWork(work, childResc);
-		addDescription(work);
-		// add premis events to the file object
+		// Add ingestion event for file object
+		addIngestionEventForChild(obj);
 		addPremisEvents(obj);
+		// add MODS
+		addDescription(work);
```
bbpennel: I can't recall why this needed to be moved, but either way I think the description should be getting added to the fileObject rather than the work here (since it is not ingestFileObjectAsWork)

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -288,22 +296,24 @@ private void ingestFileObjectAsWork(ContentContainerObject parent, Resource pare
...
 		try {
-			WorkObject newWork = repository.createWorkObject(workPid, workModel);
+			newWork = repository.createWorkObject(workPid, workModel);
 
 			addDescription(newWork);
-
-			addFileToWork(newWork, childResc);
+			FileObject fileObj = addFileToWork(newWork, childResc);
+			// add ingestion event for work object
+			addIngestionEventForContainer(newWork, parentResc);
```
bbpennel: Mainly for readability and truth in reporting, but could you move this to after `parent.addMember`?

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -405,11 +416,12 @@ private void ingestFolder(ContentContainerObject parent, Resource parentResc, Re
...
 			try {
 				obj = repository.createFolderObject(childPid, model);
+				addIngestionEventForChild(obj);
 				parent.addMember(obj);
 				
 				addDescription(obj);
-				// add premis events to the folder object
-				addPremisEvents(obj);
+				// write premis events for the folder object to fedora
+				//addPremisEvents(obj);
```
bbpennel: Can remove this commented out code

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -211,7 +217,27 @@ public void ingestWorkObjectTest() throws Exception {
 
 		// Verify the properties and content of the supplemental file
 		assertBinaryProperties(supObj, FILE2_LOC, FILE2_MIMETYPE, null, FILE2_SIZE);
-
+		// Verify that ingestion event gets added for work
+		List<PremisEventObject> eventsWork = mWork.getPremisLog().getEvents();
+		assertEquals(2, eventsWork.size());
+		
+		assertTrue(eventsWork.contains(findPremisEventByEventDetail(eventsWork, "ingested as PID: " + mWork.getPid())));
```
bbpennel: if you switch this to `assertNotNull` then you won't need to go through `eventsWork.contains`

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -211,7 +217,27 @@ public void ingestWorkObjectTest() throws Exception {
...
+		assertTrue(eventsWork.contains(findPremisEventByEventDetail(eventsWork, "added 2 child objects to this container")));
+		
+		// Verify that ingestion event gets added for primary object
+		List<PremisEventObject> eventsPrimObj = primaryObj.getPremisLog().getEvents();
+		assertEquals(1, eventsPrimObj.size());
+		assertEquals("ingested as PID: " + primaryObj.getPid().toString() + "\n ingested as filename: "
```
bbpennel: this is a meaty assertion :)  I believe you could cut down a bit on the length by using mainPid and FILE1_LOC instead of making requests against the primaryObj since that's not what's being tested here.  I wonder if it might be a bit more readable to use your awesome lambda helper method too?

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -211,7 +217,27 @@ public void ingestWorkObjectTest() throws Exception {
...
+				+ primaryObj.getOriginalFile().getFilename(), eventsPrimObj.get(0).getResource()
+				.getProperty(Premis.hasEventDetail).getString());
+		
+		// Verify that ingestion event gets added for supplementary object
+		List<PremisEventObject> eventsSupObj = supObj.getPremisLog().getEvents();
+		assertEquals(1, eventsSupObj.size());
+		assertEquals("ingested as PID: " + supObj.getPid().toString() + "\n ingested as filename: "
```
bbpennel: Same as above except supPid and FILE2_LOC

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -554,13 +580,14 @@ public void addPremisEventsTest() throws IOException {
...
-		assertEquals(2, events.size());
+		// there should be three events total: the ingestion event, plus the two added in the test
+		assertEquals(3, events.size());
 		assertTrue(events.contains(findPremisEventByType(events, Premis.Normalization)));
 		assertTrue(events.contains(findPremisEventByType(events, Premis.VirusCheck)));
 	}
 	
 	@Test
-	public void noPremisEventsAddedTest() throws IOException {
+	public void onlyIngestionEventAddedTest() throws IOException {
```
bbpennel: 👍

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -231,24 +235,28 @@ private void ingestChildren(ContentContainerObject destObj, Resource parentResc)
 	 * @throws DepositException
 	 * @throws IOException 
 	 */
-	private void ingestFileObject(ContentObject parent, Resource parentResc, Resource childResc)
+	private FileObject ingestFileObject(ContentObject parent, Resource parentResc, Resource childResc)
 			throws DepositException, IOException {
 
 		if (skipResumed(childResc)) {
-			return;
+			return null;
```
harringj: Done!

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
harringj: I'm fairly certain this has gone back and forth between those two options at least three or four times now :-) I originally had it being added to the file object, but then we discussed it and decided the work should get it. I've changed it back for the time being.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
harringj: Done!

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
harringj: Done!

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
harringj: Good suggestion

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -560,12 +576,50 @@ private void addDescription(ContentObject obj) throws IOException {
...
+			// we need the pid only if there is exactly one child
+			if (numChildren < 1) {
+				childPid = PIDs.get(resc.getURI());
+			}
+			numChildren++;
+		}
+		// don't add event unless at least one child is ingested
+		if (numChildren > 0) {
+			PremisLogger premisLogger = getPremisLogger(obj.getPid());
+			PremisEventBuilder builder = premisLogger.buildEvent(Premis.Ingestion);
```
bbpennel: Sorry to add one more thing, but I was reminded that these events need to record what agents were responsible for the event.  This should be a copypasta job from IngestDepositRecordJob at lines 84 and 85:

```
.addSoftwareAgent(SoftwareAgent.depositService.getFullname())
.addAuthorizingAgent(DepositField.depositorName.name())
```

That same is needed in `addIngestionEventForChild`

harringj: Copypasta?! You're making me hungry :-)

PR #580, opened 2017-04-11
Accessibility

hannahlwang:
Improve accessibility on CDR front page and student papers landing page


static/front/peek.js
```diff
@@ -528,6 +569,13 @@ $(function() {
 
     $(window).on("hashchange", function() {
       $(document.body).toggleClass("peek", window.location.hash == "#p");
```
lfarrell: I'd suggest just selecting document.body once, something like `var docBody = $(document.body);` then using that variable in place of the `$(document.body)` selections.

lfarrell: Looks good overall, just the one small fix. There are also some merge conflicts that will need to be worked out before it can be merged.

PR #581, opened 2017-04-20
Cdr 1741 premis file logger

harringj:


fcrepo-clients/src/main/java/edu/unc/lib/dl/event/FilePremisLogger.java
```diff
@@ -153,26 +154,49 @@ public Model getModel() {
...
+				stmtIt = eventResc.listProperties();
+				while (stmtIt.hasNext()) {
+					RDFNode objNode = stmtIt.next().getObject();
+					if (objNode.isResource()) {
+						StmtIterator objIt = objNode.asResource().listProperties();
+						// Add statements to the event's model for any objects that have properties
+						if (objIt != null) {
+							eventModel.add(objIt);
+						}
+						objIt.close();
```
bbpennel: Minor thing, but calling objIt.close here would cause an NPE if objIt were null.  Only pointing it out since there is an if statement right above it that tests for null, i'm not sure if the iterator can actually come back null.

PR #584, opened 2017-04-24
ACL Processing clients

bbpennel:
Adds the fcrepo4 implementation of the AccessControlService, which provides endpoints for enforcing the updated access control scheme.

Adds numerous clients and tests which support this processing.

Performs ACL enforcement in IngestContentJob.

Updates UserRoles and Permissions.

Changes the default behavior of FedoraPid.getString to return the repositoryPath instead of the abbreviated path since it is much more useful when shown in log statements.


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -158,13 +162,11 @@ public void runJob() {
...
 		String groups = depositStatus.get(DepositField.permissionGroups.name());
-		AccessGroupSet groupSet = new AccessGroupSet(groups);
+		
 		// Verify that the depositor is allow to ingest to the given destination
-		if (!destObj.getAccessControls().hasPermission(groupSet, Permission.ingest)) {
-			failJob("Cannot add children to destination",
-					"Depositor does not have permissions to ingest to destination "
-							+ destObj.getPid().getRepositoryPath());
-		}
+		aclService.assertHasAccess(
```
harringj: A small thing, but maybe we can store destObj.getPid() in a variable to avoid the duplicate calls

deposit/src/main/resources/service-context.xml
```diff
@@ -175,4 +175,55 @@
...
+	<bean id="sparqlQueryService" class="edu.unc.lib.dl.sparql.FusekiSparqlQueryServiceImpl">
+		<property name="fusekiQueryURL" value="${fuseki.baseUri}" />
+	</bean>
+	
+	<!-- Initialize access control dependencies -->
+	<bean id="contentPathFactory" class="edu.unc.lib.dl.fedora.ContentPathFactory"
+			init-method="init">
+		<property name="cacheMaxSize" value="100" />
+		<property name="cacheTimeToLive" value="100" />
+		<property name="queryService" ref="sparqlQueryService" />
```
harringj: I have been unable to find any examples in the Spring documentation of the use of the <ref> tag without the bean attribute. Is there a reason for the use of the bare tag here and in some of the following?

fcrepo-clients/pom.xml
```diff
@@ -144,18 +144,17 @@
...
-			<scope>provided</scope>
-		</dependency>
-		<dependency>
 			<groupId>javax.xml.bind</groupId>
 			<artifactId>jaxb-api</artifactId>
 			<version>2.1</version>
 			<scope>provided</scope>
 		</dependency>
 		<dependency>
+			<groupId>com.google.guava</groupId>
```
harringj: I know you mentioned this yesterday, but what is Guava used for?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/AccessControlServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class AccessControlServiceImpl implements AccessControlService {
+
+	private InheritedPermissionEvaluator permissionEvaluator;
+
+	private GlobalPermissionEvaluator globalPermissionEvaluator;
+
+	@Override
+	public ObjectAccessControlsBean getObjectAccessControls(PID pid) {
+		// TODO stub
```
harringj: Can you remind me what is supposed to go here eventually?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/AclFactory.java
```diff
@@ -0,0 +1,45 @@
+package edu.unc.lib.dl.acl.fcrepo4;
```
harringj: Needs license statement

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/GlobalPermissionEvaluator.java
```diff
@@ -0,0 +1,83 @@
...
+		// A principal cannot be assigned to more than one global role
+		globalPrincipalToPermissions = properties.entrySet().stream()
+				.filter(p -> p.getKey().toString().startsWith(GLOBAL_PROP_PREFIX))
+				.map(p -> new SimpleEntry<>((String) p.getValue(), (String) p.getKey()))
+				// Transform into map of principal to permissions
+				.collect(Collectors.toMap(SimpleEntry::getKey, p -> {
+					String roleString = p.getValue().substring(GLOBAL_PROP_PREFIX.length());
+					UserRole role = UserRole.valueOf(roleString);
+					// Role must be valid and not a patron
+					if (role == null) {
```
harringj: Do we need to test explicitly that the role is valid and not a patron, or does null get returned in those cases?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
```diff
@@ -0,0 +1,189 @@
...
+	public Map<String, Set<String>> getPrincipalRoles(PID target) {
+
+		// Retrieve the path of objects up to and including the target
+		List<PID> path = getPidPath(target);
+
+		Map<String, Set<String>> inheritedPrincRoles = new HashMap<>();
+		Set<String> patronPrincipals = null;
+
+		// Iterate through each step in the path except for the root content node
+		int depth = 0;
```
harringj: I still might like to see these int values stored as meaningful constants for better code readability. So, e.g., int depth = ADMIN_UNIT or if (depth >= COLLECTION). I'd be interested in knowing others' opinions on this, though. @lfarrell, do you have a preference?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
```diff
@@ -0,0 +1,189 @@
...
+		for (; depth < path.size(); depth++) {
+			PID pathPid = path.get(depth);
+
+			// For the first two objects (unit, collection), staff roles should be considered
+			if (depth < 2) {
+
+				Map<String, Set<String>> objectPrincipalRoles = objectAclFactory.getPrincipalRoles(pathPid);
+
+				// Add this object's principals/roles to the result
+				mergePrincipalRoles(inheritedPrincRoles, objectPrincipalRoles);
```
harringj: This is a random spot for this comment, but I'm thinking it could be helpful to have a basic definition somewhere in the code of what principals are, since the terms is being used with a specific, technical meaning that may not be immediately clear to future devs.

deposit/src/main/resources/service-context.xml
bbpennel: Sorry I'm not sure which tag you're referring to?  Although it looks like I need to set the cache properties with injected values rather than hard coded ones

fcrepo-clients/pom.xml
bbpennel: It is providing the cache implementation being used in the ObjectAclFactory and the ContentPathFactory

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/AccessControlServiceImpl.java
bbpennel: It is probably going to be removed, and i will go ahead and deprecate it.  It was used in f3 code, but there was only one or two places that used it for anything more than calling hasPermission.

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/GlobalPermissionEvaluator.java
bbpennel: Good catch, my if did not live up to its commentary.  Added plus a test

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
bbpennel: I totally agree, its a lot more readable with the constants, it is update now

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
bbpennel: True, i'm not sure where i would put that, and it is also googleable (as long as you make sure google knows you want principal and not principle) https://en.wikipedia.org/wiki/Principal_(computer_security)

deposit/src/main/resources/service-context.xml
harringj: Sorry, the <ref> tag. I have seen examples like this <ref bean="anotherExampleBean"/> , but none like this <ref="anotherExampleBean"/> .

bbpennel: Hmm, if I'm understanding the question correctly you are wondering about `<property ref="sparqlQueryService" />`?  You can see some examples of that pattern in the spring docs, like:
https://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#beans-factory-instantiation

There shouldn't be any cases of straight up `<ref="anotherExampleBean"/>` though since that's not valid xml

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
```diff
@@ -0,0 +1,201 @@
...
+				roles.clear();
+				roles.add(UserRole.canViewMetadata.getPropertyString());
+			}
+		}
+	}
+
+	@Override
+	public PatronAccess getPatronAccess(PID target) {
+		List<PID> pidPath = getPidPath(target);
+		if (pidPath.size() < 3) {
```
harringj: Can you remind me why 3 is the magic number here?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
```diff
@@ -0,0 +1,201 @@
...
+
+	@Override
+	public boolean isMarkedForDeletion(PID target) {
+		return getPidPath(target).stream()
+				.anyMatch(p -> objectAclFactory.isMarkedForDeletion(p));
+	}
+
+	private List<PID> getPidPath(PID pid) {
+		List<PID> path = pathFactory.getAncestorPids(pid);
+		path = new ArrayList<>(path.subList(1, path.size()));
```
harringj: Why does the sublist start from an index of 1? What's being trimmed off here?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedPermissionEvaluator.java
```diff
@@ -0,0 +1,135 @@
...
+import edu.unc.lib.dl.fedora.PID;
+
+/**
+ * Evaluates the permissions of agents on specific objects, taking into account
+ * inherited access control information
+ * 
+ * @author bbpennel
+ *
+ */
+public class InheritedPermissionEvaluator {
```
harringj: I'm wondering whether it might make sense to break this class into two or three classes: one each for evaluating staff perms and patron perms, and a utility class with the responsibility for classifying principals. It might simplify some of the logic, as well as make the code more maintainable if we ever want to tweak patron or staff access without touching the other. Thoughts?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
```diff
@@ -0,0 +1,218 @@
...
+	private LoadingCache<String, List<Entry<String, String>>> objAclCache;
+	private long cacheTimeToLive;
+	private long cacheMaxSize;
+
+	private SparqlQueryService queryService;
+
+	private final List<String> roleUris;
+
+	public ObjectAclFactory() {
+		List<String> roleUris = new ArrayList<>(UserRole.values().length);
```
harringj: Store UserRole.values() in local var to avoid duplicate calls?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
```diff
@@ -0,0 +1,218 @@
...
+
+	@Override
+	public PatronAccess getPatronAccess(PID pid) {
+		String pidString = pid.getRepositoryPath();
+		List<Entry<String, String>> objAcls = objAclCache.getUnchecked(pidString);
+
+		String patronPropertyUri = CdrAcl.patronAccess.getURI();
+		return objAcls.stream()
+				.filter(p -> patronPropertyUri.equals(p.getKey()))
+				.findFirst()
```
harringj: What is the use case for calling findFirst() here? I understand the motivation for its use below for the embargo.

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
```diff
@@ -0,0 +1,218 @@
...
+	public void init() {
+		objAclCache = CacheBuilder.newBuilder()
+				.maximumSize(cacheMaxSize)
+				.expireAfterWrite(cacheTimeToLive, TimeUnit.MILLISECONDS)
+				.build(new ObjectAclCacheLoader());
+	}
+
+	@Override
+	public Map<String, Set<String>> getPrincipalRoles(PID pid) {
+		String pidString = pid.getRepositoryPath();
```
harringj: These two lines get called repeatedly in (i.e., 87-88) several of the methods that follow. Would it be worth creating a  private method that returns the objAcls?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectPermissionEvaluator.java
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Returns true if the provided agents have patron level access to the
+	 * object represented by the given pid, and false if access is revoked
+	 * 
+	 * @param pid
+	 * @param agentPrincipals
+	 * @param permission
+	 * @return
+	 */
+	public boolean hasPatronAccess(PID pid, Collection<String> agentPrincipals,
```
harringj: Why are there two public methods for hasPatronAccess? Couldn't we simply enforce the use of Collection<String> as a param, rather than allowing a simple String param?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ContentPathFactory.java
```diff
@@ -0,0 +1,116 @@
...
+
+	public void setCacheMaxSize(long cacheMaxSize) {
+		this.cacheMaxSize = cacheMaxSize;
+	}
+
+	public void setQueryService(SparqlQueryService queryService) {
+		this.queryService = queryService;
+	}
+
+	private class AncestorCacheLoader extends CacheLoader<String, List<PID>> {
```
harringj: It almost seems like we could abstract out the common functionality from this and the ObjAclCacheLoader, but that might be more trouble than it's worth.

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedPermissionEvaluator.java
bbpennel: I agree that it would be easier to read/modify if the staff and patron parts were processed separately, and I initially went back and forth on doing it that way.

I ended up implementing it as one uber loop so as to minimize the number of iterations/lookups that would be necessary under the assumption that this method was going to be called on every request (and every object in the request).  So my thinking was to prioritize efficiency over readability a bit in this case.  However, I haven't actually measured how significant a difference there is between the approaches, which is probably not huge since the query results are cached.  So it may not be worth the added complexity.

Principal classification is actually in a separate utility though, its just being referenced statically here.

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
lfarrell: Yeah, I'd go with the constants.

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
lfarrell: Could always add the link to the code.

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
bbpennel: The patron access modification property doesn't apply to units and collections.  So I will switch this conditional to use the constants

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
bbpennel: The collections root object, which is the object that exists to be the starting point for the content tree (where you can start following pcdm:hasMember relations).  All content object's should have paths that start with this object otherwise they are orphaned.  Wonder if the object should be renamed though, since it should actually contain Unit objects rather than Collection objects.  I could probably use a constant here too.  I will add a comment here, and need to write a document describing the repo structure

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
bbpennel: This class is basically a singleton, so this should only get called once, so i think it's okay?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
bbpennel: Sure, that would also abstract out direct interaction with the cache

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
bbpennel: It only makes sense to define one patron access setting on an object, and there was still work to be done with the response before returning so it couldn't be a terminal operator like matchAny, so I'm not sure if there are other ways to handle this?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectPermissionEvaluator.java
bbpennel: We could, yes.  I just did it this way to make it more readable in the code where there was only one principal to check.  It could be easily removed if you think that would make sense

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
harringj: sounds reasonable, i was just curious

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
harringj: 👍

fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ContentPathFactory.java
bbpennel: Do you mean the iteration over a query result from the triple store?  I'd thought about adding a generic method to the SparqlQueryService interface that would take a query and give back a list of rows, but hadn't totally figured out how to do that and if it would be useful.  Could be worth considering though

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectPermissionEvaluator.java
harringj: Hmm, I suppose there is a trade-off between client code readability and economy in this code. I trust your best judgement here. If you're completely on the fence and can't decide, thenI might lean slightly toward having just one method header with the Collection<String> param, since it seems more abstract/general.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fedora/ContentPathFactory.java
harringj: Yeah, that's what I was thinking of. Definitely not something we need to prioritize right now; I had simply noticed some parallel behaviors in the code, which got the old gears spinning.

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/InheritedAclFactory.java
bbpennel: Perhaps it would make the most sense to write up a document in confluence describing our access control model and include the link to what a principal is there?

PR #585, opened 2017-04-26
Binary Replication

lfarrell:


services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+	
+	private void replicate(String binaryPath, String originalFileChecksum, String[] replicationLocations) {
+		try {
+			for (String location : replicationLocations) {
+				String fullPath = Paths.get(createRemoteSubDirectory(location, originalFileChecksum), getFilename(binaryPath)).toString();
+				String[] cmd = new String[]{"rsync", "--update", "--whole-file", "--times", "--verbose", binaryPath, fullPath};
+				Process returnCode = Runtime.getRuntime().exec(cmd);
+				
+				if (returnCode.exitValue() != 0) {
+					throw new ReplicationException("Error replicating {} to {}", binaryPath, fullPath);
```
bbpennel: Would be helpful to include the the exit value in the error

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+				}
+				
+				verifyChecksums(originalFileChecksum, fullPath);
+			}
+		} catch (IOException e) {
+			log.error("Unable to find file to replicate {}", e);
```
bbpennel: Rethrow the exception wrapped as a ReplicationException.  Also, I don't think the exception being caught here necessarily means that the file can't be found.

services-camel/src/main/java/edu/unc/lib/cdr/replication/ReplicationRouter.java
```diff
@@ -0,0 +1,28 @@
...
+
+import edu.unc.lib.cdr.ReplicationProcessor;
+
+public class ReplicationRouter extends RouteBuilder {
+	@BeanInject(value = "replicationProcessor")
+	private ReplicationProcessor replicationProcessor;
+	
+	public void configure() throws Exception {
+		onException(Exception.class)
+		.redeliveryDelay("{{error.retryDelay}}")
```
bbpennel: Indent like the other routers

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
```diff
@@ -60,7 +60,7 @@ public void configure() throws Exception {
 			.to("fcrepo:{{fcrepo.baseUrl}}?preferInclude=ServerManaged&accept=text/turtle")
 			.process(mdProcessor)
 			.multicast()
-				.to("direct-vm:imageEnhancements","direct-vm:extractFulltext");
+				.to("direct-vm:imageEnhancements","direct-vm:extractFulltext", "direct-vm:replication");
```
bbpennel: This won't replicate FITS extracts since it only applies to original_files, you may need to break replication into a separate route from the original binary processing.  I'd suggest kicking it off after indexing and before enhancements.

I'm also realizing that MODS files may get processed as an original file...  trying to decide if that's an issue.  Doesn't need to be addressed right now though.

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,154 @@
+package edu.unc.lib.cdr;
```
bbpennel: Add license headers to all the new classes

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.camel.Message;
+import org.apache.camel.Processor;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
+
+public class ReplicationProcessor implements Processor {
```
bbpennel: Add javadoc describing what this class does

services-camel/src/main/java/edu/unc/lib/cdr/replication/ReplicationRouter.java
```diff
@@ -0,0 +1,28 @@
...
+		.maximumRedeliveries("{{error.maxRedeliveries}}")
+		.backOffMultiplier(2)
+		.retryAttemptedLogLevel(LoggingLevel.WARN);
+		
+		from("direct-vm:replication")
+		.routeId("CdrReplicationRoute")
+		.log(LoggingLevel.INFO, "Calling replication route for ${headers[org.fcrepo.jms.identifier]}")
+		.to("direct:file.replication");
+		
+		from("direct:file.replication")
```
bbpennel: I wonder if this should be a part of the previous route?

services-camel/src/test/java/edu/unc/lib/cdr/replication/ReplicationRouterTest.java
```diff
@@ -0,0 +1,106 @@
...
+	private static final String userID = "bypassAdmin";
+	private static final String userAgent = "curl/7.37.1";
+	private static final String fileID = "/file1";
+	private final String eventTypes = EVENT_NS + "ResourceCreation";
+	private final String replicationRoute = "CdrReplicationRoute";
+	
+	@PropertyInject(value = "fcrepo.baseUrl")
+	private static String baseUri;
+	
+	@EndpointInject(uri = "mock:fcrepo")
```
bbpennel: I think some of these injected variables aren't used in this test, try removing some and seeing if it makes a difference?

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationDestinationUnavailableException.java
```diff
@@ -0,0 +1,10 @@
+package edu.unc.lib.cdr;
+
+@SuppressWarnings("serial")
```
bbpennel: It would also make sense for this exception to be a subclass of ReplicationException

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+	
+	private String createFilePath(String basePath, String originalFileChecksum) {
+		String[] binaryFcrepoChecksumSplit = originalFileChecksum.split(":");
+		String checksum = binaryFcrepoChecksumSplit[2];
+		
+		String[] tokens = Iterables.toArray
+				(Splitter.fixedLength(2).split(checksum), 
+						String.class);
+		
+		String remotePath = new StringJoiner("/")
```
lfarrell: The files in my VM have this sort of path /var/lib/tomcat7/fcrepo4-data/fcrepo.binary.directory/02/f1/c2/02f1c2f26b4adc0e23bef86ae481b5e81c2d99afp

services-camel/src/main/java/edu/unc/lib/cdr/replication/ReplicationRouter.java
lfarrell: It makes it easier test to set it up like this.

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationDestinationUnavailableException.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.cdr;
+
+public class ReplicationDestinationUnavailableException extends ReplicationException {
+	private static final long serialVersionUID = 8L;
+
```
harringj: Does anyone think we may want to overload the constructor to allow this exception to be thrown without a message? (Genuine question, I don't have an opinion on this.)

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationException.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.cdr;
+
+public class ReplicationException extends RuntimeException {
+	private static final long serialVersionUID = 7L; 
+
```
harringj: Ditto here for the non-param constructor

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
```diff
@@ -23,7 +23,7 @@
 import org.apache.camel.builder.RouteBuilder;
 
 import edu.unc.lib.cdr.BinaryMetadataProcessor;
-
+//techmd_fits
```
harringj: What's this little guy doing here?

services-camel/src/main/resources/config.properties
```diff
@@ -17,6 +17,8 @@ cdr.enhancement.jp2.mimetype=image/jp2
 cdr.enhancement.postIndexingDelay=5000
 cdr.enhancement.processingThreads=2
 
+cdr.replication.locations=
```
harringj: Did we ever decide whether we need this variable and, if so, what it's values should be?

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+		
+		if (!Files.exists(fullPath)) {
+			new File(replicationPath).mkdirs();
+		}
+
+		return replicationPath;
+	}
+	
+	private String dbFilePath(String uri, String BinaryChecksum) {
+		String filePath = "/tmp/" + BinaryChecksum;
```
bbpennel: The filename doesn't actually matter, so you don't really need to pass in the checksum here.  Also, we don't really want to hardcode stuff into /tmp/, particularly for local development, so use `File.createTempFile` so that it can put the file into wherever java thinks they should go.

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+		InputStream response = null;
+		
+		try {
+			URLConnection connection = new URL(uri).openConnection();
```
bbpennel: You'll probably run into authentication issues with a vanilla java client like this eventually.  Please use FcrepoClient instead to make a GET request, there is probably already one initialized in services-camel's context files that you can pass in.  Using it would be something along the lines of https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectDataLoader.java#L131

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+	
+	private void replicate(String binaryPath, String originalFileChecksum, String[] replicationLocations, String binaryMimeType, String binaryUri) throws InterruptedException {
+		String remoteFile = null;
+		String localBinary = null;
+		
+		try {
+			for (String location : replicationLocations) {
+				String fullPath = createRemoteSubDirectory(location, originalFileChecksum);
+				
+				if (Files.exists(Paths.get(binaryPath))) {
```
bbpennel: Could you add a comment here to explain what this conditional is doing?

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+		Path fullPath = Paths.get(replicationPath);
+		
+		if (!Files.exists(fullPath)) {
+			new File(replicationPath).mkdirs();
+		}
+
+		return replicationPath;
+	}
+	
+	private String dbFilePath(String uri, String BinaryChecksum) {
```
bbpennel: Could you rename this method to reflect what it does?  My understanding is that it is downloading a file from a fedora URI, storing it to a temp file, and then returning that temp file's path

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+		String localBinary = null;
+		
+		try {
+			for (String location : replicationLocations) {
+				String fullPath = createRemoteSubDirectory(location, originalFileChecksum);
+				
+				if (Files.exists(Paths.get(binaryPath))) {
+					localBinary = binaryPath;
+				} else {
+					localBinary = dbFilePath(binaryUri, originalFileChecksum);
```
bbpennel: The temp file from this is going to need to be cleaned up once it is finished being replicated, which we absolutely don't want to do for the other case.

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
bbpennel: Also, you don't need to calculate the localBinary path once per call to the `replication` method, not for every destination location

services-camel/src/main/java/edu/unc/lib/cdr/headers/CdrFcrepoHeaders.java
```diff
@@ -27,4 +27,6 @@
 	
 	public static final String CdrBinaryChecksum = "Checksum";
 	
+	public static final String CdrBinaryUri = "BinaryUri";
```
bbpennel: Not strictly a part of this PR, but would you be willing to namespace all of these headers to the cdr?  So basically change to "CdrBinaryUri" for the value

services-camel/src/main/java/edu/unc/lib/cdr/replication/ReplicationRouter.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package edu.unc.lib.cdr.replication;
+
+import org.apache.camel.BeanInject;
+import org.apache.camel.LoggingLevel;
+import org.apache.camel.builder.RouteBuilder;
+
+import edu.unc.lib.cdr.ReplicationProcessor;
+
+public class ReplicationRouter extends RouteBuilder {
```
bbpennel: Class javadoc

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
bbpennel: yeah, death to the little guy

services-camel/src/main/java/edu/unc/lib/cdr/ReplicationProcessor.java
lfarrell: Wouldn't this replicate it with the temp file's name?

bbpennel: Specify the destination filename rather than just path in the rsync command

services-camel/src/test/resources/web.xml
```diff
@@ -0,0 +1,46 @@
...
+
+    <load-on-startup>1</load-on-startup>
+  </servlet>
+
+  <servlet-mapping>
+    <servlet-name>jersey-servlet</servlet-name>
+    <url-pattern>/rest/*</url-pattern>
+  </servlet-mapping>
+
+  <!-- filter to add test auth to grizzly -->
```
lfarrell: @harringj Do we still need the commented out bit?

harringj: Good question, @lfarrell. @ben, any thoughts? Can we delete this bit, or do we need it for future reference?

bbpennel: I would say we can remove it, since we aren't testing fedora with authentication here at the moment and can find this reference in the future in the fedora codebase.

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+            }
+            // Ignore duplicate creation of base container
+            return baseUri;
+        }
+    }
+    
+    private URI determineRepositoryPath(URI baseUri) throws IOException, FcrepoOperationFailedException {
+    	    PID parentPid = null;
+        try (FcrepoResponse response = client.post(baseUri).perform()) {
+            parentPid = PIDs.get(response.getLocation());
```
bbpennel: Seems like the PID bit isn't needed since its returning the uri (as a string), which should be the same thing as the response.getLocation()

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+        // Create a parent object to put the binary into
+        URI contentBase = createBaseContainer(RepositoryPathConstants.CONTENT_BASE);
+        URI binaryUri = determineRepositoryPath(contentBase);
+
+        String bodyString = "Test text";
+        String filename = "test.txt";
+        String checksum = "82022e1782b92dce5461ee636a6c5bea8509ffee";
+        InputStream contentStream = new ByteArrayInputStream(bodyString.getBytes());
+
+        BinaryObject internalObj = repository.createBinary(binaryUri, "binary_test", contentStream, filename, MIMETYPE, checksum, null);
```
bbpennel: There's a number of lines that are longer than 120.  I'm not sure what our policy should be on this since we have checkstyle disabled for tests (somewhat annoyingly)

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+
+        when(message.getHeader(CdrBinaryChecksum)).thenReturn(checksum);
+        when(message.getHeader(CdrBinaryUri)).thenReturn(internalObj.getUri().toString());
+
+        processor.process(exchange);
```
bbpennel: Please verify that the file is replicated to the expected location after running the processor for these tests.

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+        BinaryObject externalObj = repository.createBinary(binaryUri, "external_binary_test", contentStream, filename, MIMETYPE, null, null);
+
+        when(message.getHeader(CdrBinaryChecksum)).thenReturn(checksum);
+        when(message.getHeader(CdrBinaryUri)).thenReturn(externalObj.getUri().toString());
+
+        processor.process(exchange);
+    }
+    
+    @Test (expected = ReplicationException.class)
+    public void checksumMismatchTest() throws Exception {
```
bbpennel: 👍

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+
+        when(message.getHeader(CdrBinaryChecksum)).thenReturn(badChecksum);
+        when(message.getHeader(CdrBinaryUri)).thenReturn(externalObj.getUri().toString());
+
+        processor.process(exchange);
+    }
+    
+    @Test (expected = ReplicationDestinationUnavailableException.class)
+    public void badReplicationLocationTest() throws Exception {
+    	    processor = new ReplicationProcessor(repository, "/some/bad/location", 3, 100L);
```
bbpennel: Only this first line is executed since the ReplicationDestinationUnavailableException is immediate.  For clarity it would probably be best to remove the rest of the method.

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+        BinaryObject internalObj = repository.createBinary(binaryUri, "binary_test", contentStream, filename, MIMETYPE, checksum, null);
+
+        when(message.getHeader(CdrBinaryChecksum)).thenReturn(checksum);
+        when(message.getHeader(CdrBinaryUri)).thenReturn(internalObj.getUri().toString());
+
+        processor.process(exchange);
+    }
+    
+    @Test
+    public void replicateExternalFileTest() throws Exception {
```
bbpennel: In the fedora parlance this isn't actually an external file (that would be if the binary just contained a uri to the content), so I'd suggest removing the word from the method name to avoid confusion.

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorTest.java
```diff
@@ -0,0 +1,166 @@
...
+import org.mockito.Mock;
+
+import edu.unc.lib.dl.fcrepo4.BinaryObject;
+import edu.unc.lib.dl.fcrepo4.PIDs;
+import edu.unc.lib.dl.fcrepo4.Repository;
+import edu.unc.lib.dl.fedora.PID;
+
+public class ReplicationProcessorTest {
+    private ReplicationProcessor processor; 
+    private final String replicationLocations = "/tmp";
```
bbpennel: Please use a generated temp folder rather than the system `/tmp`.  See the previous comment about this.

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        processor.process(exchange);
+
+        String remoteChecksum = DigestUtils.sha1Hex(new ByteArrayInputStream(testText.getBytes()));
+        assertEquals(localChecksum, remoteChecksum);
+    }
+
+    @Test(expected = ReplicationDestinationUnavailableException.class)
+    public void testBadReplicationLocations() throws Exception {
+        processor = new ReplicationProcessor(repository, badReplicationLocations, maxRetries, retryDelay);
```
bbpennel: This method ends on this line.  It would be clearer to remove the rest of the method

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        try {
+            processor.process(exchange);
+        } finally {
+            verify(processor).equals(ReplicationDestinationUnavailableException.class);
+        }
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testReplicationtFail() throws Exception {
```
bbpennel: typo in method name

bbpennel:
One other suggestion here: if it doesn't cause issues, I think it would be better to move the ReplicationProcessor into the replication package

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorTest.java
harringj: What's the typo?

harringj: Fixed!

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorTest.java
harringj: Ahh, good call. I somehow failed to notice the check inside of the constructor.

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
harringj: I fixed them

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
harringj: Done! I'd suggest the Fedora community come up with less confusing terminology ;-)

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
harringj: Done!

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
harringj: Ok, took that out.

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+            if (e.getStatusCode() != HttpStatus.SC_CONFLICT) {
+                throw e;
+            }
+            // Ignore duplicate creation of base container
+            return baseUri;
+        }
+    }
+    
+    private URI determineRepositoryPath(URI baseUri) throws IOException, FcrepoOperationFailedException {
+    	    PID parentPid = null;
```
harringj: got em

services-camel/src/test/java/edu/unc/lib/cdr/ReplicationProcessorIT.java
```diff
@@ -0,0 +1,193 @@
...
+
+    @Mock
+    Exchange exchange;
+    @Mock
+    Message message;
+
+    @Before
+    public void init() throws IOException {
+        PIDs.setRepository(repository);
+        processor = new ReplicationProcessor(repository, "/tmp", 3, 100L);
```
harringj: done

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
```diff
@@ -41,26 +41,40 @@ public void configure() throws Exception {
...
+        from("direct:process.creation")
+            .routeId("ProcessCreationEvent")
             .log(LoggingLevel.DEBUG, "Processing binary metadata for ${headers[org.fcrepo.jms.identifier]}")
             .removeHeaders("CamelHttp*")
             .to("fcrepo:{{fcrepo.baseUrl}}?preferInclude=ServerManaged&accept=text/turtle")
             .process(mdProcessor)
+            .choice()
+                .when(simple("${headers[org.fcrepo.jms.identifier]} regex '.*original_file'"))
+                    .to("direct-vm:replication")
+                    .log(LoggingLevel.INFO, "Replication completed for ${headers[org.fcrepo.jms.identifier]}")
```
harringj: Ha, you're listed as the author of this class :-)

harringj: I moved the logging statement into the ReplicationRouter as follows
`from("direct-vm:replication")
            .routeId("CdrReplicationRoute")
            .log(LoggingLevel.INFO, "Calling replication route for ${headers[org.fcrepo.jms.identifier]}")
            .log(LoggingLevel.INFO, "Replication completed for ${headers[org.fcrepo.jms.identifier]}")
            .to("direct:file.replication");`

Not sure whether this is where you wanted it or not.

PR #588, opened 2017-05-19
Check styles

harringj:


pom.xml
```diff
@@ -232,6 +232,27 @@
...
+						<phase>verify</phase>
+						<configuration>
+							<configLocation>checkstyle.xml</configLocation>
+							<encoding>UTF-8</encoding>
+							<consoleOutput>true</consoleOutput>
+							<failsOnError>true</failsOnError>
+							<linkXRef>false</linkXRef>
+						</configuration>
+						<goals>
+							<goal>check</goal>
```
harringj: Would you mind if we called this goal 'checkstyle' instead, just to make it clearer what we're checking?

lfarrell: Fine by me. I just took the fcrepo version.

bbpennel: I don't have especially strong feelings about this, but `check` does appear to be the start goal from what I can tell from looking at the checkstyle plugin docs https://maven.apache.org/plugins/maven-checkstyle-plugin/usage.html and from other examples I've seen when googling.

PR #596, opened 2017-06-12
Increased cap for number of files to display in a structure browse

bbpennel:


bbpennel:
@harringj no unfortunately not, there isn't a unit test for either of the two classes impacted by this PR

PR #598, opened 2017-06-13
Cdr 1842 content type filter

harringj:


solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilter.java
```diff
@@ -0,0 +1,139 @@
...
+			contentTypeProperties.load(new InputStreamReader(this.getClass().getResourceAsStream(
+					"toContentType.properties")));
+		} catch (IOException e) {
+			log.error("Failed to load mimetype mappings", e);
+		}
+	}
+
+	@Override
+	public void filter(DocumentIndexingPackage dip) throws IndexingException {
+		BinaryObject binObj = getFileObject(dip).getOriginalFile();
```
bbpennel: I think you'll need to check to make sure that the thing being indexed is either a work or a fileObject, and if its not then return immediately.  Otherwise there's a good chance this will throw an exception.  Be good to have a unit test to make sure that bit functions correctly too.

bbpennel: Also, I'd suggest a null check before getting the original file, because its possible that a Work wouldn't have a primary object, even though we generally don't want that to be the case.  That'd also make a good simple unit test case.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilter.java
```diff
@@ -0,0 +1,139 @@
...
+		String filepath = binObj.getFilename();
+		String mimetype = binObj.getMimetype();
+		log.debug("The binary has filepath " + filepath + " and mimetype " + mimetype);
+		List<String> contentTypes = new ArrayList<String>();
```
bbpennel: If you're cleaning up things might as well switch to `ArrayList<>()`.  Also, the debug statement could be switched to using {}'s instead of string concatenation

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilter.java
```diff
@@ -0,0 +1,139 @@
...
+				return ContentCategory.image;
+			if (mimetypeType.equals("video"))
+				return ContentCategory.video;
+			if (mimetypeType.equals("audio"))
+				return ContentCategory.audio;
+		}
+
+		String contentCategory = (String) contentTypeProperties.get("mime." + mimetype);
+		if (contentCategory == null)
+			contentCategory = (String) contentTypeProperties.get("ext." + extension);
```
bbpennel: Since there's a lot happening here with working its way down the list of possible sources, could you have a test case that hits this "get content type by extension" bit successfully?  We haven't really been going for 100% code coverage with our unit tests, but it'd be good to make sure all the major cases work.

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilterTest.java
```diff
@@ -0,0 +1,114 @@
...
+import edu.unc.lib.dl.fcrepo4.WorkObject;
+import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.search.solr.model.IndexDocumentBean;
+
+/**
+ * 
+ * @author harring
+ *
+ */
+public class SetContentTypeFilterTest {
```
bbpennel: The tests here so far look good!

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilter.java
```diff
@@ -0,0 +1,139 @@
...
+public class SetContentTypeFilter implements IndexDocumentFilter {
+	
+	private static final Logger log = LoggerFactory.getLogger(SetContentTypeFilter.class);
+	private static final Pattern EXTENSION_REGEX = Pattern.compile("^[^\\n]*[^.]\\.(\\d*[a-zA-Z][a-zA-Z0-9]*)[\"'{}, _\\-`()*]*$");
+	private static final int EXTENSION_LIMIT = 8;
+	
+	private Properties mimetypeToExtensionMap;
+	private Properties contentTypeProperties;
+	
+	public SetContentTypeFilter() {
```
bbpennel: Looking at this (and i realize this is my own crappy code), its really a configuration error for the files to fail to load, and we really don't want to be eating that and just putting something in the log.  Could you either remove the try block and make this constructor throw the IOException, or rethrow an exception with a little extra context?  Up to you which you prefer.

harringj: Fixed!

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilter.java
harringj: would we want to return immediately in this case, too?

bbpennel: Yes that makes sense to me.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilter.java
harringj: Done!

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentTypeFilter.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+        // add string with content category name + extension to list of content types
+        contentTypes.add(contentType.toString());
+    }
+
+    private ContentCategory getContentCategory(String mimetype, String extension) {
+        if (mimetype == null) {
+            return ContentCategory.unknown;
+        }
+        int index = mimetype.indexOf('/');
```
harringj: @ben I'm wondering whether we can discard lines 126-139 altogether, and also get rid of the null check on 141, and just handle all cases the same way. Is there any reason to treat the image/audio/video cases as special ones?

PR #602, opened 2017-06-16
Fixed bug which caused Exiftool results to be skipped

bbpennel:


deposit/src/main/java/edu/unc/lib/deposit/validate/ExtractTechnicalMetadataJob.java
```diff
@@ -274,8 +276,8 @@ private Element getFitsIdentificationInformation(Document fitsDoc) {
             // Trust any answer agreed on by multiple tools
             for (Element el : identification.getChildren("identity", FITS_NS)) {
                 if (el.getChildren("tool", FITS_NS).size() > 1
-                        || (!"Exiftool".equals(el.getChild("tool", FITS_NS).getAttributeValue("toolname"))
-                                && !"application/x-symlink".equals(el.getAttributeValue("mimetype")))) {
+                        || !("Exiftool".equals(el.getChild("tool", FITS_NS).getAttributeValue("toolname"))
```
harringj: It looks like we may have a logic problem here, since ~p & ~q is not equivalent to ~(p & q). Maybe the logic was wrong to begin with, but just thought I'd call this out :-)

bbpennel: You're right, the logic was wrong to begin with, this pull request is partially fixing it :) (its a repetition of a PR Dean merged last week to master, for fcrepo4 instead).  This logic error was what was preventing some item's of Jessica's from getting mimetypes, since they were generated by Exiftool but not symlinks.  Its somewhat moot now though since in the newer versions of FITS this particular situation is handled by the FITS_SINGLE_STATUS logic above, but I think its still a relevant change.

PR #605, opened 2017-06-28
Cdr 1858 update set relations filter

harringj:
@ben let me know if you see any improvements i can make. i’ll be back in the office on 7/6 and can make tweaks before the sprint finishes.

in particular, i’m not clear if we want to throw an indexing exception when there is no primary object (e.g., if somehow a folder object were to be passed to the filter, which seems unlikely) or simply return, as i have it now.


harringj:
@lfarrell just a heads up to hold off on merging this until ben has had a chance to weigh in. i think it's in good shape overall, but there are a couple of small edge cases i want to be sure i've handled correctly.

bbpennel:
We only need to record a primary object for a work object.  At this point we are considering it to be valid for a work not to have a primary object.

The part about invalid terms is independent of the primary object bit and could apply to folders, works and files.  So probably don't want to quit the whole filter if its not a work, just the part about setting the primary object

harringj:
Thanks for clarifying, @ben! I've changed the filter to allow objects without primary objects. Any other changes needed?

PR #609, opened 2017-07-14
Status tags

harringj:
@ben, this is ready for review!

### DIFF
--- a/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
+++ b/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
@@ -15,8 +15,19 @@
  */
 package edu.unc.lib.dl.data.ingest.solr.filter;
 
+import java.util.ArrayList;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import edu.unc.lib.dl.acl.fcrepo4.InheritedAclFactory;
+import edu.unc.lib.dl.acl.service.PatronAccess;
 import edu.unc.lib.dl.data.ingest.solr.exception.IndexingException;
 import edu.unc.lib.dl.data.ingest.solr.indexing.DocumentIndexingPackage;
+import edu.unc.lib.dl.fcrepo4.ContentObject;
+import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.search.solr.util.FacetConstants;
 /**
  * Sets access-related status tags
  *
@@ -24,11 +35,50 @@ import edu.unc.lib.dl.data.ingest.solr.indexing.DocumentIndexingPackage;
  *
  */
 public class SetAccessStatusFilter implements IndexDocumentFilter {
+    private static final Logger log = LoggerFactory.getLogger(SetAccessStatusFilter.class);
+    private InheritedAclFactory aclFactory;
 
-	@Override
-	public void filter(DocumentIndexingPackage dip) throws IndexingException {
-		// TODO Auto-generated method stub
+    @Override
+    public void filter(DocumentIndexingPackage dip) throws IndexingException {
+        log.debug("Performing set access status filter on {}", dip.getPid());
+        dip.getDocument().setStatus(determineAccessStatus(dip));
+    }
 
-	}
+    /**
+     * Sets acl factory
+     *
+     * @param aclFactory an inherited acl factory
+     */
+    public void setAclFactory(InheritedAclFactory aclFactory) {
+        this.aclFactory = aclFactory;
+    }
+
+    private List<String> determineAccessStatus(DocumentIndexingPackage dip)
+            throws IndexingException {
+
+        ContentObject obj = dip.getContentObject();
+        PID pid = obj.getPid();
+        List<String> status = new ArrayList<>();
+
+        if (aclFactory.isMarkedForDeletion(pid)) {
+            status.add(FacetConstants.MARKED_FOR_DELETION);
+        }
+
+        if (aclFactory.getEmbargoUntil(pid) != null) {
+            status.add(FacetConstants.EMBARGOED);
+        }
+
+        if (aclFactory.getPatronAccess(pid).equals(PatronAccess.everyone)) {
+            status.add(FacetConstants.PUBLIC_ACCESS);
+        } else if (aclFactory.getPatronAccess(pid).equals(PatronAccess.authenticated)) {
+            status.add(FacetConstants.STAFF_ONLY_ACCESS);
+        }
+
+        if (aclFactory.getPatronAccess(obj.getParent().getPid()).equals(PatronAccess.authenticated)) {
+            status.add(FacetConstants.PARENT_HAS_STAFF_ONLY_ACCESS);
+        }
+
+        return status;
+    }
 
 }
--- a/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
+++ b/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
@@ -22,7 +22,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import edu.unc.lib.dl.acl.fcrepo4.InheritedAclFactory;
+import edu.unc.lib.dl.acl.fcrepo4.ObjectAclFactory;
 import edu.unc.lib.dl.acl.service.PatronAccess;
+import edu.unc.lib.dl.acl.util.AccessPrincipalConstants;
 import edu.unc.lib.dl.data.ingest.solr.exception.IndexingException;
 import edu.unc.lib.dl.data.ingest.solr.indexing.DocumentIndexingPackage;
 import edu.unc.lib.dl.fcrepo4.ContentObject;
@@ -36,7 +38,8 @@ import edu.unc.lib.dl.search.solr.util.FacetConstants;
  */
 public class SetAccessStatusFilter implements IndexDocumentFilter {
     private static final Logger log = LoggerFactory.getLogger(SetAccessStatusFilter.class);
-    private InheritedAclFactory aclFactory;
+    private InheritedAclFactory inheritedAclFactory;
+    private ObjectAclFactory objAclFactory;
 
     @Override
     public void filter(DocumentIndexingPackage dip) throws IndexingException {
@@ -45,12 +48,21 @@ public class SetAccessStatusFilter implements IndexDocumentFilter {
     }
 
     /**
-     * Sets acl factory
+     * Sets inherited acl factory
      *
      * @param aclFactory an inherited acl factory
      */
-    public void setAclFactory(InheritedAclFactory aclFactory) {
-        this.aclFactory = aclFactory;
+    public void setInheritedAclFactory(InheritedAclFactory iaf) {
+        this.inheritedAclFactory = iaf;
+    }
+
+    /**
+     * Sets non-inherited acl factory
+     *
+     * @param aclFactory an inherited acl factory
+     */
+    public void setObjectAclFactory(ObjectAclFactory oaf) {
+        this.objAclFactory = oaf;
     }
 
     private List<String> determineAccessStatus(DocumentIndexingPackage dip)
@@ -60,21 +72,24 @@ public class SetAccessStatusFilter implements IndexDocumentFilter {
         PID pid = obj.getPid();
         List<String> status = new ArrayList<>();
 
-        if (aclFactory.isMarkedForDeletion(pid)) {
+        PatronAccess inheritedAccess = inheritedAclFactory.getPatronAccess(pid);
+        PatronAccess objAccess = objAclFactory.getPatronAccess(pid);
+
+        if (inheritedAclFactory.isMarkedForDeletion(pid)) {
             status.add(FacetConstants.MARKED_FOR_DELETION);
         }
 
-        if (aclFactory.getEmbargoUntil(pid) != null) {
+        if (inheritedAclFactory.getEmbargoUntil(pid) != null) {
             status.add(FacetConstants.EMBARGOED);
         }
 
-        if (aclFactory.getPatronAccess(pid).equals(PatronAccess.everyone)) {
-            status.add(FacetConstants.PUBLIC_ACCESS);
-        } else if (aclFactory.getPatronAccess(pid).equals(PatronAccess.authenticated)) {
+        if (inheritedAccess.equals(PatronAccess.none)) {
             status.add(FacetConstants.STAFF_ONLY_ACCESS);
+        } else if (inheritedAclFactory.getPrincipalRoles(pid).containsKey(AccessPrincipalConstants.PUBLIC_PRINC)) {
+            status.add(FacetConstants.PUBLIC_ACCESS);
         }
 
-        if (aclFactory.getPatronAccess(obj.getParent().getPid()).equals(PatronAccess.authenticated)) {
+        if (inheritedAccess.equals(PatronAccess.none) && !objAccess.equals(PatronAccess.none)) {
             status.add(FacetConstants.PARENT_HAS_STAFF_ONLY_ACCESS);
         }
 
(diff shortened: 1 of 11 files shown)


solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetContentStatusFilter.java
```diff
@@ -17,79 +17,61 @@
...
-        } else {
-            // Check to see if this object is the DefaultWebObject for its parent
-            DocumentIndexingPackage parentDip = dip.getParentDocument();
-            String parentsDWO = parentDip.getFirstTriple(CDRProperty.defaultWebObject.toString());
-            if (parentsDWO != null) {
-                if (dip.getPid().equals(new PID(parentsDWO))) {
-                    status.add(FacetConstants.CONTENT_IS_DEFAULT_OBJECT);
-                }
+        if (obj instanceof WorkObject) {
+            if (((WorkObject) obj).getPrimaryObject() == null) {
```
bbpennel: I'm trying to decide if it'd be better to do a `resc.hasProperty(Cdr.primaryObject)` or stick with this.  Its nice to be able to use the api to make this kind of determination rather than put the implementation details here.  On the other hand, getPrimaryObject is actually retrieving the primary object from fedora.  That probably won't matter because of the planned caching changes since that object will already have been cached from the other filters that actually need the primary object.  So this is probably fine, but what do you think?

harringj: You raise a good point, and if the caching weren't going to obviate the need to make a fresh call to fedora for the primary object, I would say let's change it. What I do like about the current implementation is that it shows very cleanly that the property and status only apply to work objects. It also provides a small efficiency by quickly short-circuiting the logic whenever any other object type enters the filter. So, I would be inclined to leave it as is unless you feel strongly that it should be changed.

bbpennel: The `instanceof` bit would still communicate that it only applies to works, and could be kept either way.  Its mainly a readability/maintainability decision versus minor performance hit (once caching is in place).  Another approach would be to add a "hasPrimaryObject" method to the WorkObject class.  But I'm okay with keeping things the way they are

harringj: True, I guess I was thinking the instanceof check would go away under your previous suggestion, but you're right that there's no reason it has to. I went ahead and changed it to use the resource :-)

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -0,0 +1,99 @@
...
+
+        PatronAccess inheritedAccess = inheritedAclFactory.getPatronAccess(pid);
+        PatronAccess objAccess = objAclFactory.getPatronAccess(pid);
+
+        if (inheritedAclFactory.isMarkedForDeletion(pid)) {
+            status.add(FacetConstants.MARKED_FOR_DELETION);
+        }
+
+        if (inheritedAclFactory.getEmbargoUntil(pid) != null) {
+            status.add(FacetConstants.EMBARGOED);
```
bbpennel: Could you check with julie to see if she would want this facet to only say if something is directly embargoed, or also include things that are embargoed because their parent is embargoed?

Was wondering the same thing about marked for delete, but I feel like that one makes sense to use the inherited value.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        if (inheritedAccess.equals(PatronAccess.none)) {
+            status.add(FacetConstants.STAFF_ONLY_ACCESS);
+        } else if (inheritedAclFactory.getPrincipalRoles(pid).containsKey(AccessPrincipalConstants.PUBLIC_PRINC)) {
+            status.add(FacetConstants.PUBLIC_ACCESS);
+        }
+
+        if (inheritedAccess.equals(PatronAccess.none) && !objAccess.equals(PatronAccess.none)) {
```
bbpennel: 👍 good that something can be marked STAFF_ONLY_ACCESS and PARENT_HAS_STAFF_ONLY_ACCESS

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param aclFactory an inherited acl factory
+     */
+    public void setObjectAclFactory(ObjectAclFactory oaf) {
+        this.objAclFactory = oaf;
+    }
+
+    private List<String> determineAccessStatus(DocumentIndexingPackage dip)
+            throws IndexingException {
+
+        ContentObject obj = dip.getContentObject();
```
bbpennel: The ContentObject is only being used to get the PID, which you can get from the dip directly, so I think you can remove this

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
```diff
@@ -0,0 +1,152 @@
...
+        initMocks(this);
+
+        when(pid.toString()).thenReturn(PID_STRING);
+
+        when(dip.getDocument()).thenReturn(idb);
+        when(dip.getPid()).thenReturn(pid);
+        when(dip.getContentObject()).thenReturn(contentObj);
+        when(contentObj.getPid()).thenReturn(pid);
+
+        when(inheritedAclFactory.isMarkedForDeletion(any(PID.class))).thenReturn(true);
```
bbpennel: its kind of weird having this happen for every test.  I'd suggest moving it down into testIsMarkedForDeletion.  It would be good to test the negative case as well to ensure that everything isn't getting marked as deleted

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+        principalRoles = new HashMap<>();
+        principalRoles.put(AccessPrincipalConstants.PUBLIC_PRINC, new HashSet<String>());
+        when(inheritedAclFactory.getPrincipalRoles(any(PID.class))).thenReturn(principalRoles);
+        // default for test suite is staff-only access; changes made within test cases as required
+        when(inheritedAclFactory.getPatronAccess(any(PID.class))).thenReturn(PatronAccess.none);
```
bbpennel: I'd suggest PatronAccess.parent as the default.  Its basically the neutral setting and it is the default of the two acl factories.

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
bbpennel: Actually, maybe it would be good to have a single negative case test to make sure that all of the status can NOT get set

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+        filter.filter(dip);
+
+        verify(idb).setStatus(listCaptor.capture());
+        assertTrue(listCaptor.getValue().contains(FacetConstants.STAFF_ONLY_ACCESS));
+    }
+
+    @Test
+    public void testHasPublicAccess() throws Exception {
+        when(inheritedAclFactory.getPatronAccess(any(PID.class))).thenReturn(PatronAccess.everyone);
```
bbpennel: I don't think this is necessary, and isn't a valid result from getPatronAccess for a folder/work/fileObj.  If you have PatronAccess.parent as the default for the tests that'll take care of this case.  I'd also suggest moving 
`principalRoles.put(AccessPrincipalConstants.PUBLIC_PRINC, new HashSet<String>());`
from the init method down to this test so that you can be sure it's not affecting other unrelated tests.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
harringj: Three lines of code...gone!

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
harringj: got it

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
harringj: done!

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
harringj: good idea. done!

harringj:
@bbpennel, this PR is complete and ready for review :-) The one outstanding question I have concerns an 'else' vs 'if-else' at line 51 of SetStatusTagsFilter. I have an 'if-else' currently, but think we might want to have an 'else' there, since content will always either be described or not described.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetStatusTagsFilter.java
```diff
@@ -0,0 +1,70 @@
...
+
+        if (contentStatus.contains(FacetConstants.IS_PRIMARY_OBJECT)) {
+            statusTags.add(FacetConstants.IS_PRIMARY_OBJECT);
+        }
+
+        if (accessStatus.contains(FacetConstants.EMBARGOED)) {
+            statusTags.add(FacetConstants.EMBARGOED);
+        }
+
+        if (accessStatus.contains(FacetConstants.EMBARGOED_PARENT)) {
```
bbpennel: These tags are mainly intended to be notifications of status where action can be taken, so I would rather not have EMBARGOED_PARENT included since you don't do anything about it here.  I figure we will make adjustments to these tags once users can get to the admin interface

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
```diff
@@ -0,0 +1,198 @@
...
+        initMocks(this);
+
+        when(pid.toString()).thenReturn(PID_STRING);
+
+        when(dip.getDocument()).thenReturn(idb);
+        when(dip.getPid()).thenReturn(pid);
+
+        principalRoles = new HashMap<>();
+
+        // default for test suite is staff-only access; changes made within test cases as required
```
bbpennel: comment is out of date since the default has changed

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -0,0 +1,105 @@
...
+            status.add(FacetConstants.EMBARGOED_PARENT);
+            status.add(FacetConstants.EMBARGOED);
+        } else if (parentEmbargo != null) {
+            status.add(FacetConstants.EMBARGOED_PARENT);
+        } else if (objEmbargo != null) {
+                status.add(FacetConstants.EMBARGOED);
+        }
+
+        if (inheritedAccess.equals(PatronAccess.none)) {
+            status.add(FacetConstants.STAFF_ONLY_ACCESS);
```
bbpennel: One of the tests was weirding me out (testParentHasStaffOnlyAccess).  Then I realized it was because STAFF_ONLY_ACCESS (as compared to PARENT_HAS_STAFF_ONLY_ACCESS) should only be added if the object itself is marked as PatronAccess.none.  The inherited value shouldn't matter here.

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetStatusTagsFilterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        filter.filter(dip);
+
+        verify(idb).setStatusTags(setCaptor.capture());
+        assertTrue(setCaptor.getValue().contains(FacetConstants.EMBARGOED_PARENT));
+        assertTrue(setCaptor.getValue().contains(FacetConstants.EMBARGOED));
+    }
+
+    @Test
+    public void testNoStatusTags() throws Exception {
+        // NB: only one status tag is set in the default case, CONTENT_NOT_DESCRIBED
```
bbpennel: Perhaps you should test to make sure its present?

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
harringj: good catch!

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetStatusTagsFilter.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.HashSet;
+import java.util.List;
+
+import edu.unc.lib.dl.data.ingest.solr.exception.IndexingException;
+import edu.unc.lib.dl.data.ingest.solr.indexing.DocumentIndexingPackage;
+import edu.unc.lib.dl.search.solr.model.IndexDocumentBean;
+import edu.unc.lib.dl.search.solr.util.FacetConstants;
+
+/**
+ *
```
harringj: whoops, done!

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetStatusTagsFilter.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public void filter(DocumentIndexingPackage dip) throws IndexingException {
+        IndexDocumentBean idb = dip.getDocument();
+        idb.setStatusTags(determineStatusTags(idb));
+    }
+
+    private HashSet<String> determineStatusTags(IndexDocumentBean idb) throws IndexingException {
+
+        HashSet<String> statusTags = new HashSet<String>();
```
harringj: t-rex!

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetStatusTagsFilter.java
```diff
@@ -0,0 +1,70 @@
...
+
+        if (accessStatus.contains(FacetConstants.EMBARGOED)) {
+            statusTags.add(FacetConstants.EMBARGOED);
+        }
+
+        if (accessStatus.contains(FacetConstants.EMBARGOED_PARENT)) {
+            statusTags.add(FacetConstants.EMBARGOED_PARENT);
+        }
+
+            return statusTags;
```
harringj: fixed

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetStatusTagsFilterTest.java
harringj: good idea :-)

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
harringj: (face-palm)

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetStatusTagsFilter.java
harringj: I've read this a few times, and I still can't quite understand what you mean by "since you don't do anything about it here." Is there something I should be doing with it?

bbpennel: It was a long way of saying I'd rather not have EMBARGOED_PARENT recorded in the status tags :)  These tags will show up next to result entries rather than as facets so you can easily spot something that you might want to change or is important to expose about the item.

PR #610, opened 2017-07-17
Added maven license plugin and updated license headers in .java files

smithjp:


bbpennel:
One thing I'm wondering is if some the exclusions are necessary since the plugin is only set to include java source folders and jsp files?  Just seems like it would affect xml files and such if none of the locations where they are located are included

PR #611, opened 2017-07-17
Solr routes

lfarrell:
Updates solr-routing with the myriad changes in fcrepo4


services-camel/src/main/java/edu/unc/lib/cdr/SolrIngestProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+        this.factory = factory;
+        this.pipeline = pipeline;
+        this.solrUpdateDriver = solrUpdateDriver;
+    }
+
+    @Override
+    public void process(Exchange exchange) throws Exception {
+        final Message in = exchange.getIn();
+        String fcrepoBinaryUri = (String) in.getHeader(FCREPO_URI);
+        PIDs.setRepository(repository);
```
bbpennel: This shouldn't be happening here, it should be getting taken care of in context files or at the beginning of a unit test.  We don't really want to ever reset this after application startup.

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
```diff
@@ -42,38 +42,43 @@ public void configure() throws Exception {
...
-                    .to("direct-vm:replication")
-                    .to("direct:process.enhancements")
-                .when(simple("${headers[org.fcrepo.jms.identifier]} regex '.*techmd_fits'"))
-                    .to("direct-vm:replication")
-                .otherwise()
-                    .log(LoggingLevel.WARN, "Cannot process binary metadata for ${headers[org.fcrepo.jms.identifier]}")
-            .end();
+                .removeHeaders("CamelHttp*")
+                .to("fcrepo:{{fcrepo.baseUrl}}?preferInclude=ServerManaged&accept=text/turtle")
+                .process(mdProcessor)
```
bbpennel: This is no longer behind a check to make sure the incoming object is a binary, so I am guessing running the BinaryMetadataProcessor on a folder will either have undesirable effects or at least not do anything useful.  I'm also guessing that the mdProcessor doesn't do anything required for the solr processor.  Maybe this could be moved into direct:process.binary.original?

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
```diff
@@ -42,38 +42,43 @@ public void configure() throws Exception {
...
+                .multicast()
+                .to("direct:process.binary.original", "direct:process.solr");
 
+        from("direct:process.binary.original")
```
bbpennel: Since this processes original_file|techmd_fits it might be best to rename the route to just "process.binary"

services-camel/src/main/java/edu/unc/lib/cdr/routing/MetaServicesRouter.java
```diff
@@ -42,38 +42,43 @@ public void configure() throws Exception {
...
-            .routeId("ProcessEnhancements")
+            .routeId("AddBinaryEnhancements")
             .multicast()
-                .to("direct-vm:imageEnhancements","direct-vm:extractFulltext");
+            .to("direct-vm:imageEnhancements","direct-vm:extractFulltext");
+
+        from("direct:process.solr")
+            .routeId("IngestSolrIndexing")
+            .log(LoggingLevel.DEBUG, "Ingest solr indexing for ${headers[org.fcrepo.jms.identifier]}")
+            .filter(simple("${headers[org.fcrepo.jms.resourceType]} not contains '" + Binary.getURI() + "'"))
```
bbpennel: This is too permissive, and would let through DepositRecords with the current stack.  I'd suggest listing the types that are allowed rather than aren't allowed, since we are likely to add more (currently) unindexed types, such as premis events, agents, agreements, licenses, etc.  The types that should be indexed are Collections, Units, Works, FileObjects

services-camel/src/main/java/edu/unc/lib/cdr/solr/SolrRouter.java
```diff
@@ -0,0 +1,46 @@
...
+package edu.unc.lib.cdr.solr;
+
+import org.apache.camel.BeanInject;
+import org.apache.camel.LoggingLevel;
+import org.apache.camel.builder.RouteBuilder;
+
+import edu.unc.lib.cdr.SolrIngestProcessor;
+
+/**
+ *
```
bbpennel: Please add a short description for non-test classes

services-camel/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -55,10 +55,249 @@
...
+		<property name="targetMethod" value="setRepository" />
+		<property name="arguments">
+			<list>
+				<ref bean="repository" />
+			</list>
+		</property>
+	</bean>
+	
+	<!-- Solr settings -->
+	<bean id="collectionsPid" class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
```
bbpennel: I'm surprised this isn't causing errors since there's no tripleStoreQueryService?  We're going to need to find another way to work this.  Maybe guess is this is going to go away completely and be replaced by a method call on the Repository object or some replacement

From the time being, I'd like to remove this bean and you can probably pass a null into SolrQueryLayerService with something like:
<property name="collectionsPid" value="null" />

services-camel/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -55,10 +55,249 @@
...
+				<value>/Collections</value>
+			</list>
+		</property>
+	</bean>
+
+	<bean id="solrFullUpdatePipeline"
+		class="edu.unc.lib.dl.data.ingest.solr.indexing.DocumentIndexingPipeline">
+		<property name="filters">
+			<list>
+				<ref bean="setDisplayOrderFilter" />
```
bbpennel: Switch to a different Filter, this one isn't updated yet and is probably going to be removed.  A good alternative is setRecordDatesFilter, which I thought you'd done on the previous branch?

services-camel/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -101,8 +340,15 @@
 	<bean id="replicationProcessor" class="edu.unc.lib.cdr.ReplicationProcessor">
 		<constructor-arg ref="repository"/>
 		<constructor-arg value="${cdr.replication.locations}" />
+	</bean>
+	
+	<bean id="solrIngestProcessor" class="edu.unc.lib.cdr.SolrIngestProcessor">
+		<constructor-arg ref="repository"/>
 		<constructor-arg value="${error.maxRedeliveries}" />
 		<constructor-arg value="${error.retryDelay}" />
+		<constructor-arg ref="factory" />
```
bbpennel: I don't think "factory" exists?  You probably need to create a DocumentIndexingPackageFactory.  Also, give it a more informative name

services-camel/src/test/java/edu/unc/lib/cdr/BinaryMetadataProcessorTest.java
```diff
@@ -89,9 +89,9 @@ public void validTest() throws Exception {
 
         processor.process(exchange);
 
-        verify(message).setHeader("Checksum", checksum);
-        verify(message).setHeader("MimeType", mimetype);
-        verify(message).setHeader("BinaryPath", BINARY_BASE + "61/67/3d/" + checksum);
+        verify(message).setHeader("CdrChecksum", checksum);
```
bbpennel: Since there are constants for these now and this is what constants are supposed to protect you from, lets switch this over to using the constants :)

PR #613, opened 2017-07-21
Cdr 1882 remove deprecated modules

harringj:


bbpennel:
You can actually go ahead and delete the contents of the three modules that you removed as well.  They don't automatically go away because of being ignored since they are already registered

PR #615, opened 2017-07-24
Updated xml editor to use version 1.3.0

bbpennel:


static/js/xmleditor/xsd2json.js
```diff
@@ -306,36 +304,22 @@ SchemaManager.prototype.processIncludes = function(schema) {
...
-				np : true
-			};
+	this.originatingRoot = {
+			schema : true,
+			ns : this.originatingSchema.targetNSIndex,
+			namespaces : [],
+			elements : [],
+			np : true
+		};
+
```
harringj: Ouch for the O = n^3 algorithm!

static/js/xmleditor/xsd2json.js
```diff
@@ -358,20 +342,51 @@ SchemaManager.prototype.mergeRootLevelElements = function() {
...
+	this.addNamespacePrefixes(namespacePrefixes, prefixUsed, {'xs': self.xsNS});
+	this.addNamespacePrefixes(namespacePrefixes, prefixUsed, self.originatingSchema.localNamespaces);
+	this.addNamespacePrefixes(namespacePrefixes, prefixUsed, {'': self.originatingSchema.targetNS});
+	for (var targetNS in self.imports) {
+		self.imports[targetNS].forEach(function(schema) {
+			self.addNamespacePrefixes(namespacePrefixes, prefixUsed, schema.localNamespaces);
 		});
 	}
+	var namespaceRegistry = [];
+	var anonymousNamespaceIndex = 1;
```
harringj: Why is this set to 1 rather than to 0?

static/js/xmleditor/xsd2json.js
```diff
@@ -418,17 +422,27 @@ SchemaManager.prototype.resolveTypeReferences = function(definition) {
...
 							+ " from definition " + definition.name);
 				
 				// Compute nested types depth first before merging in this type
 				this.resolveTypeReferences(typeDef);
 				
-				this.mergeType(definition, typeDef);
+				this.mergeType(definition, typeDef, typeRef.mergeMode);
 			}
 		}
+		if (definition.elements) {
```
harringj: Maybe name the function and pass it in rather than defining it twice?

static/js/xmleditor/xsd2json.js
bbpennel: Yeah, that's some solid nesting, luckily the data its dealing with is generally pretty small (usually only 0 to 3 imports)

static/js/xmleditor/xsd2json.js
```diff
@@ -466,25 +480,77 @@ SchemaManager.prototype.resolveSchema = function(schema, name) {
...
 };
 
+SchemaManager.prototype.mergeTypeArray = function(baseArray, typeArray, key, mergeMode) {
+	var mergeFunction = this['mergeType_' + key + '_' + mergeMode];
+	if (mergeFunction === undefined) {
+		throw Error('Invalid definition key ' + JSON.stringify(key) + ' or merge mode ' + JSON.stringify(mergeMode));
+	}
+	return mergeFunction.call(this, baseArray, typeArray);
+};
+
```
harringj: I'm sure this is just me not really understanding this application, but why have three different prototypes that return the exact same thing?

static/js/xmleditor/xsd2json.js
```diff
@@ -466,25 +480,77 @@ SchemaManager.prototype.resolveSchema = function(schema, name) {
...
+	var typeResult = [];
+	for (var j = 0; j < typeArray.length; j++) {
+		var typeItem = typeArray[j];
+		if (!baseNames[typeItem.name]) {
+			typeResult.push(typeItem);
+		}
+	}
+	return typeResult.concat(baseArray);
+};
+
```
harringj: Make that six prototypes returning the same thing :-)

static/js/xmleditor/xsd2json.js
bbpennel: the prefixes are 1 based, I guess because ns0 looked weird?

static/js/xmleditor/xsd2json.js
bbpennel: this method is gluing together components to form a function name and then calling it, with a bit of validation so that if something goes awry it won't be as hard to debug.  I think vog's intention was to not have a mountain of if statements to determine which merge function to call, to instead have this thing work it out.  I guess it's a very javascripty approach to the problem

static/js/xmleditor/xsd2json.js
bbpennel: My understanding is that the behavior can vary based component type, but the difference is only implemented for 2 of them right now

PR #617, opened 2017-08-03
Cdr 1852 Refactor caching in repository class

bbpennel:


deposit/src/main/resources/fcrepo-clients-context.xml
```diff
@@ -48,8 +48,15 @@
 		<property name="aclService" ref="aclService" />
 	</bean>
 	
-	<bean id="repository" class="edu.unc.lib.dl.fcrepo4.Repository">
+	<bean id="repositoryObjectCacheLoader" class="edu.unc.lib.dl.fcrepo4.RepositoryObjectCacheLoader">
 		<property name="client" ref="fcrepoClient" />
+		<property name="repository" ref="repository" />
```
harringj: wouldn't  the repository bean need to be defined before the repositoryObjectCacheLoader, since the latter refers to the former?

bbpennel: the order in which beans are initialized gets worked out by spring as long as there aren't any impossible situations (such as two constructors referring to eachother's objects.  which is why injecting the dependencies as setters instead of as constructor parameters is often safer)

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Repository.java
```diff
@@ -80,6 +82,18 @@
 
     private RepositoryObjectDataLoader repositoryObjectDataLoader;
 
+    private LoadingCache<PID, RepositoryObject> contentObjCache;
```
harringj: maybe this field should be named "repositoryObjCache" to make it clearer what's being cached?

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectCacheLoaderTest.java
```diff
@@ -0,0 +1,242 @@
...
+    @Test(expected = ObjectTypeMismatchException.class)
+    public void loadUnsupportedTypeTest() throws Exception {
+        mockResponseBodyWithType(pid, Fcrepo4Repository.Pairtree);
+
+        objectCacheLoader.load(pid);
+    }
+
+    @Test(expected = ObjectTypeMismatchException.class)
+    public void loadInvalidPidTest() throws Exception {
+        mockResponseBodyWithType(pid, Fcrepo4Repository.Pairtree);
```
harringj: might be nice to have a valid object type here to isolate the pid as the problem

fcrepo-clients/src/test/java/edu/unc/lib/dl/fcrepo4/RepositoryTest.java
```diff
@@ -111,108 +107,38 @@ public void createDepositRecordTest() {
...
-    public void getContentObjectInvalidBaseTest() {
-        PID pid = PIDs.get("invalid/43f3016b-9cb0-4d7b-92a4-0e2921362a66");
-
-        repository.getContentObject(pid);
-    }
-
-    @Test(expected = ObjectTypeMismatchException.class)
-    public void getContentObjectInvalidComponentTest() {
-        PID pid = getBinaryPid();
+        FedoraException fedoraE = new NotFoundException("Not found");
```
harringj: is a not found exception really a fedora exception?

bbpennel: yeah, its generated from a 404 from the fedora api

PR #628, opened 2017-08-21
Autoclose deposit streams

bbpennel:


deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/IngestDeposit.java
```diff
@@ -371,9 +371,7 @@ private void ingestObject(String ingestPid, boolean confirmExisting) throws Depo
 		uploadIngestFiles(foxmlDoc, pid);
 
 		// Ingest the object's FOXML
-		try {
-
-			ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+		try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
```
harringj: Should outputStream be declared as an OutputStream instead of the more specific type?

deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/MakeFOXML.java
```diff
@@ -319,8 +319,7 @@ private void addEventsDS(PID p, Document foxml) {
 	}
 
 	private void saveRELSEXTtoFOXMl(Model relsExt, Document foxml) {
-		try {
-			ByteArrayOutputStream os = new ByteArrayOutputStream();
+		try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
```
harringj: Ditto

deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/IngestDeposit.java
bbpennel: In this case I left it as ByteArrayOutputStream because it needs to call outputStream.toByteArray() at line 382, which is specific to ByteArrayOutputStream

deposit/src/main/java/edu/unc/lib/deposit/fcrepo3/MakeFOXML.java
bbpennel: double ditto :)

PR #636, opened 2017-10-12
Update solr access filtering

bbpennel:


admin/src/main/webapp/WEB-INF/solr-search-context.xml
```diff
@@ -89,8 +89,9 @@
 		<property name="searchSettings" ref="searchSettings" />
 		<property name="searchStateFactory" ref="searchStateFactory" />
 		<property name="facetFieldUtil" ref="facetFieldUtil" />
-		<property name="collectionsPid" ref="collectionsPid" />
 		<property name="pathFactory" ref="pathFactory" />
+		<property name="disablePermissionFiltering" value="false" />
+		<property nmae="globalPermissionEvaluator" ref="globalPermissionEvaluator" />
```
harringj: "nmae" should be "name"

PR #645, opened 2017-10-20
Check for child elements

lfarrell:


bbpennel:
It may help to add an exclusion for related items inside of related items since they are the one subelement that we don't render currently

PR #647, opened 2017-10-23
Rename packages in services-camel

lfarrell:


lfarrell:
See https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/648 for pull request that combines package renaming with moving processors into more descriptive packages.

PR #649, opened 2017-10-24
Move processors into more descriptive packages

bbpennel:


bbpennel:
When I start services-camel now I am getting `Error creating bean with name 'CdrServiceSolrUpdate'...  java.lang.ClassNotFoundException: edu.unc.lib.dl.services.camel.SolrIngestProcessor`, so I think there may be a reference to the wrong class name somewhere

I would prefer if the processors were in the same package as the routes they apply to rather than in a subpackage.  Similar to how it works over in this camel toolbox example:
https://github.com/fcrepo4-exts/fcrepo-camel-toolbox/tree/master/fcrepo-audit-triplestore/src/main/java/org/fcrepo/camel/audit/triplestore
I've updated one of the processors to make it works as expected.   Could you do the same with the others?  Otherwise we end up with a lot of packages with one thing in them
https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/649/commits/4938ef7e6ecc0db27c0b6cde4774a8812104b121

PR #650, opened 2017-10-26
Cdr 1922

bbpennel:
Adds updated mark for deletion workflow.

Injects pidMinter into repo objs where it is needed.

Adds username as a principal, introduces bean for holding all authentication details

Adds sparql update service for updates to fedora.


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
```diff
@@ -55,11 +55,13 @@
     protected PremisLogger premisLog;
 
     protected RepositoryObject(PID pid, RepositoryObjectLoader repositoryObjectLoader,
-            RepositoryObjectDataLoader dataLoader, RepositoryObjectFactory repoObjFactory) {
+            RepositoryObjectDataLoader dataLoader, RepositoryObjectFactory repoObjFactory,
```
harringj: what was the reason for adding a pid minter to RepositoryObject and each of its subclasses? i thought at one point we were thinking of restricting its use mainly to the factory

metadata/src/main/java/edu/unc/lib/dl/rdf/Premis.java
```diff
@@ -192,7 +195,7 @@ public static String getURI() {
             "http://www.loc.gov/premis/rdf/v1#TermOfRestriction");
 
     /** Premis event types
```
harringj: nitpick, but maybe move comment down one line from the "/**"

metadata/src/main/java/edu/unc/lib/dl/sparql/SparqlUpdateHelper.java
```diff
@@ -0,0 +1,209 @@
...
+
+        StringBuilder query = new StringBuilder();
+        addInsert(query, subjUri, property, objectString);
+        query.append("\nWHERE {}");
+
+        return query.toString();
+    }
+
+    /**
+     * Constructs a sparql query which replaces all instances of a property on a resource with a new valu
```
harringj: "value", not "valu" :-)

metadata/src/test/java/edu/unc/lib/dl/util/RDFModelUtilTest.java
```diff
@@ -1,98 +0,0 @@
-/**
```
harringj: is RDFModelUtil being retired completely? i noticed you left some methods in the class, but wasn't sure whether these were just to avoid compile errors for the time being

security/src/main/java/edu/unc/lib/dl/acl/util/AccessGroupSet.java
```diff
@@ -62,7 +66,7 @@ public void addAccessGroup(String group) {
      * Determines is any of the objects contained within the specified
      * collection are present in the access group set. If the objects contain
      * pid prefixes, they are stripped off before checking
-     * 
+     *
      * @param c
      *            collection to be checked for matches.
```
harringj: this comment should be moved up one line beside the  @param

sword-server/pom.xml
```diff
@@ -56,6 +56,10 @@
 					<artifactId>servlet-api</artifactId>
 					<groupId>javax.servlet</groupId>
 				</exclusion>
```
harringj: what's the sword server doing in this PR, and why didn't it have jena previously?

metadata/src/main/java/edu/unc/lib/dl/rdf/Premis.java
bbpennel: That was one of the weird things that the rdf schema to java class tool generated, i'll fix it though

metadata/src/test/java/edu/unc/lib/dl/util/RDFModelUtilTest.java
bbpennel: Its not being totally retired, I just readjusted the scope of it so that it is just about serializing, deserializing and streaming models.  Creating sparql query strings seemed pretty distinct from those other actions.

security/src/main/java/edu/unc/lib/dl/acl/util/AccessGroupSet.java
bbpennel: eclipse defaulted to putting descriptions of parameters on the following lines for some weird reason, so that stuff is scattered around.  trying to move away from it though

sword-server/pom.xml
bbpennel: This is actually excluding jena from sword-server, since it was importing an old version of jena.  It was mainly being a pain because eclipse kept offering me the wrong classpath for jena classes in the services module because of this.

metadata/src/main/java/edu/unc/lib/dl/sparql/SparqlUpdateHelper.java
bbpennel: no, val you!

PR #651, opened 2017-10-26
Cdr 1963 clear empty solr fields

harringj:


solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/indexing/SolrUpdateDriver.java
```diff
@@ -92,19 +86,25 @@ public void addDocument(IndexDocumentBean idb) throws IndexingException {
...
+
+                Object value = field.getValue();
+
+                // Id field needs to be set like a non-partial update
+                if (ID.getSolrField().equals(fieldName)) {
+                    sid.addField(fieldName, value);
+                    continue;
                 }
             }
+
```
bbpennel: I think a part is missing here
https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/634/files#diff-b26ca2d0f8ef083ca3f4b788413f0fb9R107

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/IndexDocumentBean.java
```diff
@@ -15,503 +15,481 @@
...
-
-    protected Date dateCreated;
-    protected Date dateAdded;
-    protected Date dateUpdated;
-
-    protected String citation;
-
-    protected String fullText;
+
+    private Map<String, Object> fields;
```
bbpennel: Oh yeah, I guess this is why I started calling the getters everywhere instead of referencing variables

PR #654, opened 2017-10-31
adds new service and controller classes to handle creation of containers

harringj:


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/AddContainerService.java
```diff
@@ -0,0 +1,108 @@
...
+
+            child = repoObjFactory.createFolderObject(model);
+            repoObjFactory.createMemberLink(URI.create(parentResc.getURI()), child.getPid().getRepositoryUri());
+        } else if (Cdr.Work.equals(containerType)) {
+            aclService.assertHasAccess(
+                    "User does not have permissions to create collections",
+                    parentPid, agent.getPrincipals(), Permission.ingest);
+
+            child = repoObjFactory.createWorkObject(model);
+            repoObjFactory.createMemberLink(URI.create(parentResc.getURI()), child.getPid().getRepositoryUri());
```
bbpennel: I'd suggest doing this by getting the parentObject and calling addMember on it, since that includes validation that its a legitimate child type being added.  You could probably move this out of the ifs since its the same.  Should also make sure that the child was actually created (was a valid type) before continuing.  Might be best to to perform the creation and adding (and maybe the premis) in a transaction so you could back the creation of the object out if its an invalid child type (like adding an AdminUnit to a folder)

services/src/test/java/edu/unc/lib/dl/cdr/services/processing/AddContainerIT.java
```diff
@@ -0,0 +1,187 @@
...
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doThrow;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+import java.util.Map;
+
+import org.codehaus.jackson.map.ObjectMapper;
+import org.codehaus.jackson.type.TypeReference;
+import org.jgroups.util.UUID;
```
bbpennel: hmm, this should actually be the `java.util.UUID` implementation of UUID.  Looks like I have the same issue in the MarkForDeletionIT too, must be something about the services webapp that is making eclipse suggest the wrong UUID first.  Could you update these?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/AddContainerController.java
```diff
@@ -15,67 +15,88 @@
...
+    public ResponseEntity<Object> createWork(@PathVariable("id") String id) {
+        return createContainer(id, Cdr.Work);
+    }
+
+    private ResponseEntity<Object> createContainer(String id, Resource containerType) {
+        Map<String, Object> result = new HashMap<>();
+        result.put("action", "create");
+
+        PID parentPid = PIDs.get(id);
+        result.put("pid", parentPid.getUUID());
```
bbpennel: This is fine, since if we move to an asynchronous model later we may not know the pid of the newly created container.  At some point we will probably want to revisit the responses from these if we want to improve responsiveness of the admin ui

services/src/test/java/edu/unc/lib/dl/cdr/services/processing/AddContainerIT.java
```diff
@@ -0,0 +1,187 @@
...
+
+        assertChildContainerAdded(parent);
+
+        // Verify response from api
+        Map<String, Object> respMap = getMapFromResponse(result);
+        assertEquals(parentPid.getUUID(), respMap.get("pid"));
+        assertEquals("create", respMap.get("action"));
+    }
+
+    @SuppressWarnings("unchecked")
```
bbpennel: According to eclipse these suppressions aren't needed on any of the test methods in this class

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/AddContainerService.java
```diff
@@ -0,0 +1,128 @@
...
+     * @param parentPid pid of parent obj to add child container to
+     * @param containerType the type of new container to be created
+     */
+    public void addContainer(AgentPrincipals agent, PID parentPid, Resource containerType) {
+
+        FedoraTransaction tx = txManager.startTransaction();
+        try {
+            // Create the new container
+            Model model = ModelFactory.createDefaultModel();
+            model.createResource(parentPid.getRepositoryPath());
```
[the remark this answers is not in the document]
harringj: ok, didn't realize that createResource() didn't affect the model. removed those lines and am now passing null to all of the create-object methods

services/src/test/java/edu/unc/lib/dl/cdr/services/processing/AddContainerIT.java
harringj: removed. funny, since i thought eclipse suggested these at one point. fickle IDE!

PR #655, opened 2017-10-31
Order display correctly

lfarrell:
Fix sort ordering


access/src/main/java/edu/unc/lib/dl/ui/controller/AdvancedSearchFormController.java
```diff
@@ -47,7 +47,7 @@
 @Controller
 @RequestMapping("/advancedSearch")
 public class AdvancedSearchFormController extends AbstractSolrSearchController {
-    HashMap<String, String> dropdownVals = new HashMap<String, String>();
+    LinkedHashMap<String, String> dropdownVals = new LinkedHashMap<String, String>();
```
bbpennel: Just for the future, you don't need to specify the implementation on the left, can just do:
`Map<String, String> dropdownVals = new LinkedHashMap<>();`

PR #657, opened 2017-11-01
Move sortSubmit data-main to collectionBrowse page

lfarrell:


access/src/main/webapp/WEB-INF/jsp/collectionBrowse.jsp
```diff
@@ -35,4 +35,5 @@
 	<c:param name="showFolderFacet">false</c:param>
 	<c:param name="entryTemplate">searchResults/browseResultEntry.jsp</c:param>
 </c:import>
-</div>
\ No newline at end of file
+</div>
+<script type="text/javascript" src="/static/js/lib/require.js" data-main="/static/js/public/sortSubmit"></script>
```
bbpennel: You could rename this to be something like collectionBrowse.js so that its clear that it is the main method for that page?

PR #661, opened 2017-11-01
Fcrepo4 map txt and set default date added

bbpennel:


solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -87,6 +87,9 @@ public void filter(DocumentIndexingPackage dip) throws IndexingException {
         if (dip.getDocument().getTitle() == null) {
             idb.setTitle(getAlternativeTitle(dip));
         }
+        if (dip.getDocument().getDateCreated() == null) {
+            idb.setDateCreated(idb.getDateAdded());
```
harringj: is getDateAdded() guaranteed to return a non-null value here? if not, how are we handling a null value?

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilterTest.java
```diff
@@ -63,13 +60,14 @@
  */
```
harringj: fix author javadoc format

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilter.java
bbpennel: yes, date added should be always be non-null since it is a generated field from fedora.  If it weren't available, then the date created field would just get cleared.

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilterTest.java
bbpennel: Fixed

bbpennel:
Not directly caused by these changes, but solr-ingest has a number of test failures which I was fixing on this branch.  I've created another ticket to address those now though, so we should be able to merge this one

PR #662, opened 2017-11-03
Remove TripleStoreQueryServiceFusekiImpl class

lfarrell:
Not sure if we need the references in services/src/test/resources/spring-test/cdr-client-container.xml. It doesn't appear the class uses it anymore.


bbpennel:
I haven't tried running the tests yet, but this looks good so far.  Unfortunately there's a bunch of merge conflicts, so you'll need to merge in changes from fcrepo4

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectDriver.java
```diff
@@ -175,6 +178,42 @@ public InputStream getBinaryStream(BinaryObject obj) throws FedoraException {
 
     /**
      *
```
bbpennel: Since this is a public method, could you add a description?

PR #665, opened 2017-11-07
Deploy access

lfarrell:
The main page now loads. It ignores solr at this point.

### DIFF
--- a/access/pom.xml
+++ b/access/pom.xml
@@ -63,11 +63,19 @@
 					<systemProperties>
 						<systemProperty>
 							<name>server.properties.uri</name>
-							<value>file:${project.basedir}/../etc/server.properties</value>
+							<value>classpath:server.properties</value>
+						</systemProperty>
+						<systemProperty>
+							<name>acl.properties.uri</name>
+							<value>file:${project.basedir}/target/${project.build.finalName}/acl-config.properties</value>
 						</systemProperty>
 						<systemProperty>
 							<name>log.dir</name>
 							<value>${project.basedir}/target/</value>
+						</systemProperty>
+						<systemProperty>
+							<name>fcrepo.baseUri</name>
+							<value>http://localhost:8080/fcrepo/rest/</value>
 						</systemProperty>
 					</systemProperties>
 				</configuration>
--- a/access/pom.xml
+++ b/access/pom.xml
@@ -39,10 +39,12 @@
 				<configuration>
 					<scanIntervalSeconds>2</scanIntervalSeconds>
 					<stopKey>STOP</stopKey>
-					<webApp>
-						<contextPath>/</contextPath>
-						<allowDuplicateFragmentNames>true</allowDuplicateFragmentNames>
-					</webApp>
+				<webApp>
+					<webInfIncludeJarPattern>.*access-common-.*</webInfIncludeJarPattern>
+					<contextPath>/access</contextPath>
+					<allowDuplicateFragmentNames>true</allowDuplicateFragmentNames>
+					<extraClasspath>${basedir}/target/test-classes/</extraClasspath>
+				</webApp>
 					<contextHandlers>
 						<contextHandler implementation="org.eclipse.jetty.server.handler.ContextHandler">
 							<contextPath>/static</contextPath>
--- a/access/pom.xml
+++ b/access/pom.xml
@@ -70,6 +70,20 @@
 					</systemProperties>
 				</configuration>
 			</plugin>
+			
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-war-plugin</artifactId>
+				<version>3.1.0</version>
+				<configuration>
+					<webResources>
+						<resource>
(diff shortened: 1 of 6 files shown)


access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
```diff
@@ -39,13 +38,13 @@
...
- * 
+ *
  * @author bbpennel
  */
 @Controller
 public class SearchActionController extends AbstractSolrSearchController {
-    @Autowired(required = true)
-    protected PID collectionsPid;
+  //  @Autowired(required = true)
+    protected String collectionsPid = "cdr:collection";
```
bbpennel: You can get the `contentRootPid` from `RepositoryPaths`, which is the current equivalent to the collections pid.  It is static, so you don't really need to store it in an instance variable

lfarrell: Updated.

access/pom.xml
```diff
@@ -70,6 +70,20 @@
...
+				<version>3.1.0</version>
+				<configuration>
+					<webResources>
+						<resource>
+							<filtering>true</filtering>
+							<directory>../etc</directory>
+						</resource>
+					</webResources>
+				</configuration>
+			</plugin>
```
bbpennel: One other thing to add, if you look in admin/pom.xml in the jetty plugin, you'll see some webapp specific config.  Some of the settings significantly improve startup time my making it smarter about where to get jsp and other files from:
```
<webApp>
    <webInfIncludeJarPattern>.*access-common-.*</webInfIncludeJarPattern><allowDuplicateFragmentNames>true</allowDuplicateFragmentNames>
    <extraClasspath>${basedir}/target/test-classes/</extraClasspath>
</webApp>
```

lfarrell:
Updated the pom file

access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
```diff
@@ -90,7 +87,7 @@ public String list(@PathVariable("pid") String pid, Model model, HttpServletRequ
     @RequestMapping("/list")
     public String list(Model model, HttpServletRequest request) {
         SearchRequest searchRequest = generateSearchRequest(request);
-        searchRequest.setRootPid(collectionsPid.getPid());
+   //     searchRequest.setRootPid(collectionsPid.getPid());
```
harringj: delete this line?

access/src/main/webapp/WEB-INF/access-fedora-context.xml
```diff
@@ -96,7 +95,7 @@
 		<property name="externalContentPath" value="${external.base.url}" />
 	</bean>
 	
-	<bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
+<!--  	<bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
```
harringj: delete this block?

access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
bbpennel: Can actually specify the rootPid using RepositoryPaths here too

access/src/main/webapp/WEB-INF/web.xml
```diff
@@ -113,12 +113,12 @@
 		<url-pattern>/*</url-pattern>
 	</filter-mapping>
 	
-	<filter>
+	<!--  <filter>
```
bbpennel: This could should actually still be usable, might just be missing initializing some beans?

access/src/main/webapp/WEB-INF/access-fedora-context.xml
bbpennel: I think this is still going to be used, just appears to be commented out so things will start.  I'm not sure why this would fail though

access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
lfarrell: Updated

access/src/main/webapp/WEB-INF/access-fedora-context.xml
lfarrell: Seems to work fine. Maybe commented out because the referencing beans didn't see to be in the file

access/src/main/webapp/WEB-INF/web.xml
lfarrell: There currently  is no bean for it to refer to ` No bean named 'storeAccessLevelFilter' is defined`

bbpennel: You commented it out in the context file in this PR :)  but i understand that was because it was missing dependencies.  I've pushed a commit that reenables its and adds access control dependencies (stolen from admin webapp)

PR #666, opened 2017-11-07
Sort by work -> folder instead of work -> file

lfarrell:


access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
```diff
@@ -96,24 +96,24 @@ public String list(Model model, HttpServletRequest request) {
...
 		return search(searchRequest, model, request);
 	}
-	
+
 	@RequestMapping("/listContents/{pid}")
 	public String listContents(@PathVariable("pid") String pid, Model model, HttpServletRequest request) {
 		SearchRequest searchRequest = generateSearchRequest(request);
 		searchRequest.getSearchState().setResourceTypes(
-				Arrays.asList(ResourceType.Aggregate.name(), ResourceType.File.name()));
+				Arrays.asList(ResourceType.Aggregate.name(), ResourceType.Folder.name()));
```
bbpennel: ListContent is actually specifically set up to not show folders (its a flat view of a collection).  Haven't tested it yet, but i'm guessing this would make it so that files that aren't in works wouldn't show up in the ListContent view?

lfarrell: Is it currently sorting by aggregate and then by file, or just pulling in anything that's an aggregate or a loose file?

bbpennel: Its using whatever sort is provided, which is probably the default sort unless overridden.  The default sort includes a sort by the `resourceTypeSort` field.  That field causes files and aggregates to be sorted with the same weight.  Setting the resourceType field of the search state will cause it to filter to only return objects of the listed types

PR #667, opened 2017-11-08
Cdr 1920 set primary object

harringj:


metadata/src/main/java/edu/unc/lib/dl/rdf/Cdr.java
```diff
@@ -95,7 +95,7 @@ public static String getURI() {
             "http://cdr.unc.edu/definitions/model#invalidTerm" );
 
     /** Relation from an AggregateWork to a child which will be treated as the primary
-     *  work for this object.
+     *  object for this work.
```
bbpennel: The ol' switch-a-roo

bbpennel: Might as well change the documentation about it from "AggregateWork" to just "Work" while you're at it.

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
```diff
@@ -422,7 +422,7 @@ public void createMemberLink(URI parentUri, URI memberUri) throws FedoraExceptio
      * @param object
      */
     public void createRelationship(PID subject, Property property, Resource object) {
-        String sparqlUpdate = SparqlUpdateHelper.createSparqlInsert(subject.getRepositoryPath(), property, object);
+        String sparqlUpdate = SparqlUpdateHelper.createSparqlReplace(subject.getRepositoryPath(), property, object);
```
bbpennel: I would actually suggest making another method named "createExclusiveRelationship", since other methods may need to support repeating relationships, and the other createRelationship method does not replace existing relations.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetPrimaryObjectService.java
```diff
@@ -0,0 +1,100 @@
...
+     * @param fileObjPid the file object to be set as primary object
+     */
+    public void setPrimaryObject(AgentPrincipals agent, PID fileObjPid) {
+
+        aclService.assertHasAccess("Insufficient privileges to set " + fileObjPid.getUUID() + " as primary object",
+                fileObjPid, agent.getPrincipals(), editResourceType);
+
+        RepositoryObject repoObj = repoObjLoader.getRepositoryObject(fileObjPid);
+
+        if (!(repoObj instanceof FileObject)) {
```
bbpennel: It would actually make sense for work.setPrimaryObject to enforce this restriction

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/SetPrimaryObjectController.java
```diff
@@ -0,0 +1,81 @@
...
+public class SetPrimaryObjectController {
+
+    private static final Logger log = LoggerFactory.getLogger(SetPrimaryObjectController.class);
+
+    @Autowired
+    private SetPrimaryObjectService setPrimaryObjectService;
+
+    @RequestMapping(value = "/edit/setPrimaryObject/{id}", method = RequestMethod.PUT)
+    @ResponseBody
+    public ResponseEntity<Object> setPrimaryObject(@PathVariable("id") String id) {
```
bbpennel: So this is the id of the file object being made into a primary obj, right?  I wonder if it'd make sense to use a different name for the action in the url.  Generally I would expect the `set` verb to operate like a setter being called on the subject object, but this is reversed.  "assignAsPrimaryObject" maybe?  or "setAs", "make".  I'm not sure of the convention for this kind of reversed operation, although I think the structure makes sense for the operation.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetPrimaryObjectService.java
```diff
@@ -0,0 +1,100 @@
...
+
+        RepositoryObject parent = repoObj.getParent();
+        if(!(parent instanceof WorkObject)) {
+            throw new InvalidOperationForObjectType("Object of type " + parent.getClass().getName()
+            + " cannot have a primary object.");
+        }
+        repoObjFactory.createRelationship(parent.getPid(), Cdr.primaryObject, parent.getModel()
+                .createResource(fileObjPid.getRepositoryPath()));
+
+        repoObj.getPremisLog()
```
bbpennel: I think you already know that we weren't going to add a premis entry for this action, so this is probably out of date

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObjectFactory.java
harringj: good thought. i wasn't clear about all of the use-cases for this method

metadata/src/main/java/edu/unc/lib/dl/rdf/Cdr.java
harringj: 👍

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetPrimaryObjectService.java
harringj: done!

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetPrimaryObjectService.java
harringj: yep, already gone :-)

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/SetAsPrimaryObjectIT.java
```diff
@@ -0,0 +1,186 @@
...
+
+        // Verify response from api
+        Map<String, Object> respMap = getMapFromResponse(result);
+        assertEquals(fileObjPid.getUUID(), respMap.get("pid"));
+        assertEquals("setAsPrimaryObject", respMap.get("action"));
+    }
+
+    @Test
+    public void testAuthorizationFailure() throws Exception {
+        doThrow(new AccessRestrictionException()).when(aclService)
```
bbpennel: This test is good, but I'd suggest adding the fileObj as a member like in the previous test (maybe make a helper method?) so that it is a legitimate operation other than the access control exception

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/SetAsPrimaryObjectIT.java
```diff
@@ -0,0 +1,186 @@
...
+public class SetAsPrimaryObjectIT {
+
+    @Autowired
+    private WebApplicationContext context;
+    @Autowired
+    private RepositoryObjectFactory repositoryObjectFactory;
+    @Autowired
+    private AccessControlService aclService;
+    @Autowired
+    private JenaSparqlQueryServiceImpl queryService;
```
bbpennel: it would probably make more sense to wire in the model rather than the query service here since its just being used to get the model

PR #668, opened 2017-11-08
Fix collection sort order - Remove roll up and ignore children

lfarrell:


access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
```diff
@@ -101,10 +101,10 @@ public String list(Model model, HttpServletRequest request) {
 	public String listContents(@PathVariable("pid") String pid, Model model, HttpServletRequest request) {
 		SearchRequest searchRequest = generateSearchRequest(request);
 		searchRequest.getSearchState().setResourceTypes(
-				Arrays.asList(ResourceType.Aggregate.name(), ResourceType.File.name()));
+				Arrays.asList(ResourceType.Aggregate.name()));
```
bbpennel: Sorry, thought I'd already left a comment on this, but probably losing track of things at the conference.
We still want Files to show up when they aren't in works in fcrepo3 land.  What we want instead is to suppress children of works.  You can do this by calling `searchState.setIncludeParts(false)`.  Having rollup turned off is also good.

access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
```diff
@@ -278,13 +278,13 @@ private void applyContainerSettings(String pid, Document foxml, Model model, boo
 		if (settings.getViews().contains(ContainerView.LIST_CONTENTS.name())) {
 			SearchState searchState = stateFactory.createSearchState();
 			searchState.setResourceTypes(
-					Arrays.asList(ResourceType.Aggregate.name(), ResourceType.File.name()));
+					Arrays.asList(ResourceType.Aggregate.name()));
```
bbpennel: Same as below, keep standalone files

lfarrell:
Setting searchState.setIncludeParts(false) gives a solr error `edu.unc.lib.dl.ui.controller.ErrorController - An unc
aught exception occurred at /record/uuid:794941c6-fd7d-41de-a46a-c0298fbeb5df
org.apache.solr.client.solrj.impl.HttpSolrServer$RemoteSolrException: undefined field null`

PR #671, opened 2017-11-16
Cdr 2043 add cdr event sending

harringj:


services/src/test/java/edu/unc/lib/dl/cdr/services/processing/AddContainerServiceTest.java
```diff
@@ -142,31 +158,65 @@ public void addCollectionToFolderTest() {
...
+        destinations.add(parentPid);
+        Collection<PID> added = new ArrayList<>();
+        added.add(childPid);
 
         service.addContainer(agent, parentPid, Cdr.Folder);
 
         verify(premisLogger).buildEvent(eq(Premis.Creation));
         verify(eventBuilder).write();
+
+        verify(messageSender).sendAddOperation(anyString(), destinationsCaptor.capture(),
```
bbpennel: I'm trying to work through my feelings on what is important to have in unit tests or not, so don't take this as a recommendation.  After a number of times now in the fcrepo4 branch of having to go back and fix unit tests when implementation details change, I've been wanting to think through our testing strategies some.  Verifying that a message is sent does seem important to me, but it also makes the test significantly longer and more complicated.

I understand that in a unit test it is helpful to check all these things to catch programming errors.  On the other hand, are aspects like sending messages and recording events side effects/implementation details or actual expected outcomes of the method according to the public api?  It could be argued that they are since this is a preservation repository, but the explicit purpose of the class is to create containers.  Should we be testing that extra stuff like this happens in a unit test or an integration test?

I was reading a few articles about testing practices:
https://gist.github.com/Integralist/7944948
http://codebetter.com/iancooper/2011/10/06/avoid-testing-implementation-details-test-behaviours/
https://gist.github.com/androidfred/8210e67f37f5ca21da51e2fa5368853b (this one feels a bit contrived to me)

Anyways, this is probably fine (other than maybe using the `Arrays.asList` thing), it was just making me question my junit existence some more.

persistence/src/main/java/edu/unc/lib/dl/persist/services/delete/MarkForDeletionService.java
```diff
@@ -41,11 +46,19 @@
...
+            pid = PIDs.get(id);
+            Runnable job = new MarkForDeletionJob(pid, agent, repositoryObjectLoader,
+                    sparqlUpdateService, aclService);
             job.run();
+
+            removed.add(pid);
+        }
+        if (pid != null) {
+            PID destination = repositoryObjectLoader.getRepositoryObject(pid).getParent().getPid();
+            operationsMessageSender.sendRemoveOperation(agent.getUsername(), destination, removed);
```
bbpennel: Hmm, the Remove event has actually been used for destroy operations in the past, so actual removal of the object from the repo rather than marking for deletion.  Looking at master, there does not appear to be an existing event for this.  I would suggest making a new MarkForDeletion operation.  
Additionally, the event doesn't impact hierarchy (nothing moves or is removed), so I would suggest that the event not take a param for parents, just the updated objects

persistence/src/main/java/edu/unc/lib/dl/persist/services/delete/MarkForDeletionJob.java
```diff
@@ -44,9 +44,8 @@
     private AgentPrincipals agent;
     private PID pid;
 
-    public MarkForDeletionJob(PID pid, AgentPrincipals agent,
-            RepositoryObjectLoader repositoryObjectLoader, SparqlUpdateService sparqlUpdateService,
-            AccessControlService aclService) {
+    public MarkForDeletionJob(PID pid, AgentPrincipals agent, RepositoryObjectLoader repositoryObjectLoader,
```
bbpennel: this is okay, but in general it's safest to avoid formatting changes in unchanged files since those can tend to result in merge conflicts

services/src/test/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectServiceTest.java
```diff
@@ -0,0 +1,144 @@
+/**
```
bbpennel: I'm not sure why this file is showing up as new?  Maybe rebasing your branch on fcrepo4 will resolve this?

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
```diff
@@ -0,0 +1,98 @@
+/**
```
bbpennel: Not sure why this file is new?

persistence/src/main/java/edu/unc/lib/dl/persist/services/delete/MarkForDeletionJob.java
harringj: this file had other changes, before i realized they needed to be in the service class instead. this must just have been left over. i can remove it if you'd like. there was an extra param in the constructor for the message sender, which is why the formatting got changed

persistence/src/main/java/edu/unc/lib/dl/persist/services/delete/MarkForDeletionService.java
harringj: i think i fixed this. let me know if you still see there are changes that need to be made

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
harringj: ??

services/src/test/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectServiceTest.java
harringj: ??

i will try rebasing

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
```diff
@@ -63,6 +68,10 @@ public void setAsPrimaryObject(AgentPrincipals agent, PID fileObjPid) {
         }
         WorkObject work = (WorkObject) parent;
         work.setPrimaryObject(fileObjPid);
+
+        // Send message that the action completed
+        operationsMessageSender.sendIndexingOperation(agent.getUsername(), Arrays.asList(fileObjPid),
```
bbpennel: Its been a little while, but weren't we thinking of making an event specific to this action rather than sending an indexing operation?

persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
```diff
@@ -100,6 +100,30 @@ public String sendAddOperation(String userid, Collection<PID> destinations, Coll
...
 
+    /**
+     * Sends a MarkForDeletion operation message, indicating that an object is being marked for deletion
+     * from the repository
+     *
+     * @param userid id of user who triggered the operation
+     * @param marked object marked for deletion
+     * @return id of operation message
+     */
+    public String sendMarkForDeletionOperation(String userid, PID marked) {
```
bbpennel: Given that all the other events take a list of affected pids and mark for deletion is handled as a (potentially) bulk operation in the CDR, it would make sense to have this receive a list as well.  And for it to be called by the MarkForDeletionService instead of the job.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
harringj: hmm, i don't remember what we decided. i recall you saying at one point that it was currently handled this way and that it would be a fine starting point. i can create another operation, though.

persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
harringj: fixed!

persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
```diff
@@ -100,6 +100,55 @@ public String sendAddOperation(String userid, Collection<PID> destinations, Coll
...
+     * Sends a RestoreFromDeletion operation message, indicating that an object is being un-marked for deletion
+     *
+     * @param userid id of user who triggered the operation
+     * @param unmarked object being un-marked for deletion
+     * @return id of operation message
+     */
+    public String sendRestoreFromDeletionOperation(String userid, Collection<PID> unmarked) {
+        Element contentEl = createAtomEntry(userid, unmarked.iterator().next(), CDRActions.RESTORE_FROM_DELETION);
+
+        Element mark = new Element("markForDeletion", CDR_MESSAGE_NS);
```
bbpennel: Element name should match the name of the action

harringj: whoops, somehow missed that

PR #677, opened 2017-11-29
Cdr 2020 update mods

harringj:


persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
```diff
@@ -314,6 +314,32 @@ public String sendEditTypeOperation(String userid, Collection<PID> pids, Resourc
...
     }
 
+    /**
+     * Sends Update MODS operation message.
+     *
+     * @param userid id of user who triggered the operation
+     * @param pids objects whose MODS changed
+     * @return id of operation message
+     */
+    public String sendUpdateMODSOperation(String userid, Collection<PID> pids) {
```
bbpennel: how about sendUpdateDescriptionOperation?  And updateDescription generally.  Doesn't tie us quite as tightly to the schema, for whenever we stop using mods.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/UpdateMODSService.java
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @author harring
+ *
+ */
+public class UpdateMODSService {
+
+    private AccessControlService aclService;
+    private RepositoryObjectLoader repoObjLoader;
+    private OperationsMessageSender operationsMessageSender;
+    private MODSValidationUIPFilter modsFilter;
```
bbpennel: I would recommend not using this class since all of the UIP things are intended to be part of a pipeline and have a number of fedora 3 expectations in them.

It would actually be good to have a common MODS validator class, that we could use here and in the deposit pipeline.  Basically duplicate the deposit pipeline code.  Provide a premis logger, whatever agent should be used, and a mods document.

bbpennel: The abdera stuff only shows up in the update pipeline because it was being pulled out of sword requests, which you shouldn't need to deal with here.

metadata/src/main/java/edu/unc/lib/dl/validation/MODSValidator.java
```diff
@@ -0,0 +1,96 @@
...
+ * as per the configured schematron validator.
+ *
+ * @author bbpennel
+ * @author harring
+ *
+ */
+public class MODSValidator {
+    private static final Logger log = LoggerFactory.getLogger(MODSValidator.class);
+
+    private SchematronValidator schematronValidator = null;
```
harringj: is there a reason why these are set explicitly to null?

metadata/src/main/java/edu/unc/lib/dl/validation/MODSValidator.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Validates a MODS description. NB: this method only supports InputStream types that can be reset
+     * for multiple reads. Other types will be converted to a ByteArrayInputStream prior to validation.
+     *
+     * @param docStream
+     * @throws MetadataValidationException
+     * @throws IOException
+     */
+    public InputStream validate(InputStream docStream) throws MetadataValidationException, IOException {
+        if (!docStream.markSupported()) {
```
bbpennel: Looking at this now, I feel like returning an inputstream from a validation method is kind of confusing, in part because its supporting one case.  It would seem like the client calling the validator would have a better idea of what kind of data it was receiving and how to handle its datastream?  What would you think of having it instead throw an IllegalArgumentException if the inputstream doesn't support `mark` instead?

As a part of that, the javadoc should list that exception and when it would be thrown.

Alternatively, both the validate methods need to state what they are returning.

metadata/src/main/java/edu/unc/lib/dl/validation/MODSValidator.java
```diff
@@ -0,0 +1,122 @@
...
+ * as per the configured schematron validator.
+ *
+ * @author bbpennel
+ * @author harring
+ *
+ */
+public class MODSValidator {
+    private static final Logger log = LoggerFactory.getLogger(MODSValidator.class);
+
+    private SchematronValidator schematronValidator = null;
```
bbpennel: Nitpick, but these don't need to be initialized to null, they will be null by default.

persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
```diff
@@ -314,6 +314,31 @@ public String sendEditTypeOperation(String userid, Collection<PID> pids, Resourc
...
+    /**
+     * Sends Update MODS operation message.
+     *
+     * @param userid id of user who triggered the operation
+     * @param pids objects whose MODS changed
+     * @return id of operation message
+     */
+    public String sendUpdateDescriptionOperation(String userid, Collection<PID> pids) {
+        Element contentEl = createAtomEntry(userid, pids.iterator().next(),
+                CDRActions.UPDATE_MODS);
```
bbpennel: Mind if we make this action UPDATE_DESCRIPTION to match most of the rest of the naming, since this is effectively api for external consumption?

persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
```diff
@@ -314,6 +314,31 @@ public String sendEditTypeOperation(String userid, Collection<PID> pids, Resourc
...
+
+        Element updateEl = new Element("updateMods", CDR_MESSAGE_NS);
```
bbpennel: I know we're not doing this elsewhere, but it seems like we could use the CDRAction enum for giving the name to this element, since the action name has to match the element name

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/UpdateDescriptionService.java
```diff
@@ -0,0 +1,89 @@
...
+    private OperationsMessageSender operationsMessageSender;
+    private MODSValidator modsValidator;
+
+    public void updateDescription(AgentPrincipals agent, PID pid, InputStream modsStream)
+            throws MetadataValidationException, IOException {
+        aclService.assertHasAccess("User does not have permissions to update description",
+                pid, agent.getPrincipals(), Permission.editDescription);
+
+        String username = agent.getUsername();
+        modsStream = modsValidator.validate(modsStream);
```
bbpennel: If we keep this structure (see previous comment), I would prefer if we used a new variable rather than changing the value of the parameter passed to this method.

services/src/test/java/edu/unc/lib/dl/cdr/services/processing/UpdateDescriptionServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+        verify(messageSender).sendUpdateDescriptionOperation(anyString(), pidsCaptor.capture());
+        Collection<PID> pids = pidsCaptor.getValue();
+        assertEquals(pids.size(), 1);
+        assertTrue(pids.contains(objPid));
+    }
+
+    @Test(expected = AccessRestrictionException.class)
+    public void insufficientAccessTest() throws Exception {
+        doThrow(new AccessRestrictionException()).when(aclService)
+        .assertHasAccess(anyString(), eq(objPid), any(AccessGroupSet.class), any(Permission.class));
```
bbpennel: indent

services/src/test/resources/mods/invalid-mods.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bbpennel: Not so much invalid mods as invalid XML, but it will do :)

harringj: well, it's empty xml, and hence invalid mods

metadata/src/main/java/edu/unc/lib/dl/validation/MODSValidator.java
bbpennel: nope :)  probably copy paste

services/src/test/java/edu/unc/lib/dl/cdr/services/processing/UpdateDescriptionServiceTest.java
harringj: done

metadata/src/main/java/edu/unc/lib/dl/validation/MODSValidator.java
harringj: hey, i already nitpicked you on this! i left a comment asking why these were being set to null :-)

persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
harringj: done!

persistence/src/main/java/edu/unc/lib/dl/services/OperationsMessageSender.java
harringj: sure, why not

PR #678, opened 2017-11-30
Link out to RLA site codes

lfarrell:
Don't know if this is too specific, but it works for the provided use case.


access/src/main/resources/recordTransformations/modsToFullRecord.xsl
```diff
@@ -525,6 +525,11 @@
 								<xsl:when test="local-name() = 'cartographics'">
 									<xsl:call-template name="cartographics"/>
 								</xsl:when>
+								<xsl:when test="starts-with(../@displayLabel, 'RLA') and ../@xlink:href">
```
bbpennel: this doesn't need to be RLA specific, any xlink:href in a subject should be fine to link to

PR #682, opened 2017-12-12
Cdr 2037 reindexing api

harringj:


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/IndexingService.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Performs an in-place recursive reindexing of an object and its descendants
+     * or a clean reindexing, as indicated by the request
+     *
+     * @param agent security principals of the agent making request
+     * @param objPid the PID where reindexing should begin in the tree
+     * @param inplace whether in-place reindexing has been requested
+     * @throws IndexingException
+     */
+    public void reindexObjectAndChildren(AgentPrincipals agent, PID objectPid, Boolean inplace) {
```
bbpennel: You should be able to safely make `inplace` a `boolean` rather than an object in this case, which will save you the null check below.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
```diff
@@ -70,8 +70,7 @@ public void setAsPrimaryObject(AgentPrincipals agent, PID fileObjPid) {
         work.setPrimaryObject(fileObjPid);
 
         // Send message that the action completed
-        operationsMessageSender.sendIndexingOperation(agent.getUsername(), Arrays.asList(fileObjPid),
-                IndexingActionType.SET_AS_PRIMARY_OBJECT);
+        operationsMessageSender.sendSetAsPrimaryObjectOperation(agent.getUsername(), Arrays.asList(fileObjPid));
```
bbpennel: This message should include the parent and the file object (the parent is actually what needs to be reindexed).

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
```diff
@@ -89,7 +88,7 @@ public void setRepositoryObjectLoader(RepositoryObjectLoader repoObjLoader) {
     }
 
     /**
-     * @param operatonsMessageSender the message sender to set
+     * @param indexingMessageSender the message sender to set
```
bbpennel: need to switch this javadoc back to `operatonsMessageSender`

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
```diff
@@ -15,104 +15,116 @@
...
-        PID pid = new PID(id);
+    public ResponseEntity<Object> reindex(@PathVariable("id") String id) {
+        return indexObject(id);
+    }
 
-        if (!hasPermission(id)) {
-            response.setStatus(401);
-            return;
-        }
+        private ResponseEntity<Object> indexObject(String id) {
```
bbpennel: Indent is off

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
```diff
@@ -15,104 +15,116 @@
...
+
+            log.info("Updating object " + id);
+            try {
+                indexingService.reindexObject(AgentPrincipals.createFromThread(), objectPid);
+            } catch (Exception e) {
+                result.put("error", e.getMessage());
+                if (e instanceof AuthorizationException || e instanceof AccessRestrictionException) {
+                    return new ResponseEntity<>(result, HttpStatus.FORBIDDEN);
+                } else {
+                    log.error("Failed to reindex repository due to {}",  e);
```
bbpennel: Could this be switched to something like `log.error("Failed to reindex object {}", id, e);`?  Also, the exception doesn't need to be formatted into the message.

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
```diff
@@ -15,104 +15,116 @@
...
+            }
+            try {
+                indexingService.reindexObjectAndChildren(AgentPrincipals.createFromThread(), objectPid, inplace);
+            } catch (Exception e) {
+                result.put("error", e.getMessage());
+                Throwable t = e.getCause();
+                if (t instanceof AuthorizationException || t instanceof AccessRestrictionException) {
+                    return new ResponseEntity<>(result, HttpStatus.FORBIDDEN);
+                } else {
+                    log.error("Failed to reindex repository due to {}",  e);
```
bbpennel: "Failed to reindex {} object and its children"

services/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -235,6 +236,21 @@
 	    <property name="operationsMessageSender" ref="operationsMessageSender" />
 	</bean>
 	
+	<bean id="solrUpdateJmsTemplate" class="org.springframework.jms.core.JmsTemplate">
+        <property name="connectionFactory" ref="jmsFactory" />
+        <property name="defaultDestinationName" value="activemq:queue:repository.solrupdate?asyncConsumer=true" />
```
bbpennel: Replace the value with a property from the properties file.

services/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -235,6 +236,21 @@
...
+        <property name="pubSubDomain" value="false" />
+    </bean>
+    
+    <bean id="indexingMessageSender" class="edu.unc.lib.dl.services.IndexingMessageSender" >
+        <property name="jmsTemplate" ref="solrUpdateJmsTemplate" />
+    </bean>
+	
+	<bean id="indexingService" class="edu.unc.lib.dl.cdr.services.processing.IndexingService" >
```
bbpennel: I think the indent characters around here are getting sort of mixed up.  Might as well leave it i guess.  We need to do a bulk conversion of all our spring context files and pom.xml files to space indented very soon.

services/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -245,7 +261,7 @@
   <bean id="setAsPrimaryObjectService" class="edu.unc.lib.dl.cdr.services.processing.SetAsPrimaryObjectService" >
       <property name="aclService" ref="aclService" />
       <property name="repositoryObjectLoader" ref="repositoryObjectLoader" />
-      <property name="operationsMessageSender" ref="operationsMessageSender" />
+      <property name="indexingMessageSender" ref="indexingMessageSender" />
```
bbpennel: should be operationsMessageSender

services/src/test/java/IndexingServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+        Collection<PID> collections = pidsCaptor.getValue();
+        assertEquals(collections.size(), 1);
+        assertTrue(collections.contains(objPid));
+    }
+
+    @Test(expected = AccessRestrictionException.class)
+    public void insufficientAccessTest() {
+        doThrow(new AccessRestrictionException()).when(aclService)
+        .assertHasAccess(anyString(), any(PID.class), any(AccessGroupSet.class), eq(reindex));
```
bbpennel: indent

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingIT.java
```diff
@@ -0,0 +1,176 @@
...
+import edu.unc.lib.dl.test.TestHelper;
+
+/**
+ *
+ * @author harring
+ *
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextHierarchy({
+    @ContextConfiguration("/spring-test/test-fedora-container.xml"),
```
bbpennel: Is it correct that this code doesn't actually communicate with Fedora?  It seems like it just interacts with the message sender and acl service.  In which case you can probably leave out starting up fedora.  One of the tests is creating objects, but I'm not sure if they actually need to exist in this case

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingIT.java
```diff
@@ -0,0 +1,176 @@
...
+        Map<String, Object> respMap = getMapFromResponse(result);
+        assertEquals(objPid.getUUID(), respMap.get("pid"));
+        assertEquals("reindex", respMap.get("action"));
+        assertFalse(respMap.containsKey("error"));
+    }
+
+    @Test
+    public void testCleanReindex() throws Exception {
+        PID fileObjPid = makePid();
+        FileObject fileObj =repoObjFactory.createFileObject(fileObjPid, null);
```
bbpennel: space after =

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
harringj: do you want both pids included, or shall i simply pass the parent pid and then have the method retrieve the child pid to add to the message before it's sent?

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
harringj: done

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
harringj: fixed

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
harringj: done

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
harringj: done

services/src/main/webapp/WEB-INF/service-context.xml
harringj: couldn't figure out where to do this, actually, and meant to ask you. i tried creating properties in a few different places, but spring couldn't find them

services/src/main/webapp/WEB-INF/service-context.xml
harringj: fixed

services/src/test/java/IndexingServiceTest.java
harringj: fixed

persistence/src/main/java/edu/unc/lib/dl/services/AbstractMessageSender.java
```diff
@@ -0,0 +1,48 @@
+package edu.unc.lib.dl.services;
```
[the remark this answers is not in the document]
harringj: inserted

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/IndexingService.java
harringj: done

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
bbpennel: I would provide both pids to the method

services/src/main/webapp/WEB-INF/service-context.xml
bbpennel: I believe it is /etc/server.properties
Generally, the config files used for local testing are found in the /etc/ folder.
More specifically, to find this, at the top of this context file it is specifying that it is going to try to get properties from a parameter named `server.properties.uri`.  Next, if you find where the webapp is deployed to jetty (in admin/pom.xml) you can see where `server.properties.uri` is specified (or do a search for the property).  This file gets included over in `services/pom.xml` when the war file is built.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/SetAsPrimaryObjectService.java
harringj: done

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingIT.java
harringj: hmm, i thought i posted a question here already, but looks like i didn't hit the button or something...do you think we want this test to communicate with fedora? i wasn't sure about those objects, either. this was the work-in-progress section i mentioned the other day. i was trying to find a way to differentiate the clean and inplace reindexing

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
```diff
@@ -15,104 +15,115 @@
...
-            response.setStatus(401);
-            return;
+    private ResponseEntity<Object> indexObject(String id) {
+        Map<String, Object> result = new HashMap<>();
+        result.put("action", "reindex");
+        result.put("pid", id);
+
+        PID objectPid = PIDs.get(id);
+
+        log.info("Updating object " + id);
```
bbpennel: Could you change this to `"Indexing object {}", id`?

services/src/main/webapp/WEB-INF/web.xml
```diff
@@ -1,138 +1,138 @@
...
+        <realm-name>ServicesRealm</realm-name>
+    </login-config>
+    
+    <security-role>
+        <role-name>sword-depositor</role-name>
+    </security-role>
+    
+    <security-role>
+        <role-name>group-forwarding</role-name>
+    </security-role>
```
bbpennel: I guess this PR is changing the indent characters in xml files in the impacted modules?  Well, lets go through and do that to all the other context files in another PR so that we can set eclipse to indent consistently after this.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/solrUpdate/CdrEventToSolrUpdateProcessorTest.java
```diff
@@ -89,22 +84,25 @@ public void init() throws Exception {
...
 
+    @SuppressWarnings("unchecked")
     @Test
     public void testNoMessageBody() throws Exception {
         when(msg.getBody()).thenReturn(null);
 
         processor.process(exchange);
 
-        verify(jmsTemplate, never()).send(any(MessageCreator.class));
+        verify(messageSender, never()).sendIndexingOperation(anyString(), any(Collection.class),
```
bbpennel: If you use `anyCollectionOf(PID.class)` then you could skip the warning suppression

services-camel/src/test/java/edu/unc/lib/dl/services/camel/solrUpdate/CdrEventToSolrUpdateProcessorTest.java
```diff
@@ -114,92 +112,56 @@ public void testUnknownAction() throws Exception {
...
+                any(IndexingActionType.class));
     }
 
     @Test
     public void testMoveAction() throws Exception {
         List<PID> subjects = pidList(NUM_TEST_PIDS);
 
-        Document msgDoc = buildMessage(MOVE.toString(), MOVE.getName(), targetPid, subjects);
+        Document msgDoc = buildMessage(CDRActions.MOVE.toString(), CDRActions.MOVE.getName(), targetPid, subjects);
```
bbpennel: Just curious about the reason for this change?  I can just imagine myself switching it back at some point in the future without remembering it'd already changed once before.

harringj: Confusingly, we have CDRActions.MOVE and IndexingActionsType.MOVE. I was finding that I was having trouble keeping these straight while working on the code, and decided greater specificity might help others (and me) avoid confusion in future

services-camel/src/main/java/edu/unc/lib/dl/services/camel/solr/CdrEventToSolrUpdateProcessor.java
```diff
@@ -86,25 +83,15 @@ public void process(Exchange exchange) throws Exception {
         List<String> subjects = populateList("subjects", contentBody);
 
         if (MOVE.equals(solrActionType)) {
-            offer(targetId, IndexingActionType.MOVE, subjects);
+            messageSender.sendIndexingOperation(userid, Collections.singletonList(PIDs.get(targetId)),
```
bbpennel: I think this change is the cause of the ClassCastException.  Previously it was passing `subjects` in as the children, which we should probably keep doing

services-camel/src/test/java/edu/unc/lib/dl/services/camel/cdrEvents/CdrEventRouterTest.java
```diff
@@ -99,15 +99,15 @@ public void testPublishFilter() throws Exception {
     public void testReindexFilter() throws Exception {
```
bbpennel: Reindex events aren't being sent to the cdr events queue anymore after OperationsMessageSender was split, right?  So I think this test can go

PR #683, opened 2017-12-12
CDR 2066 Switch from log4j to logback

bbpennel:


metadata/src/main/java/edu/unc/lib/dl/xml/FOXMLJDOMUtil.java
```diff
@@ -84,7 +85,7 @@ public String toString() {
     private static final String getDatastreamXpath = "/f:digitalObject/f:datastream[@ID='%DSID%']";
     private static final String labelXpath =
             "/f:digitalObject/f:objectProperties/f:property[@NAME='info:fedora/fedora-system:def/model#label']/@VALUE";
-    private static Logger log = Logger.getLogger(FOXMLJDOMUtil.class);
+    private static final Logger log = LoggerFactory.getLogger(FOXMLJDOMUtil.class);
```
lfarrell: Does this still need to refer to FOXML?

fcrepo-clients/pom.xml
```diff
@@ -123,34 +123,27 @@
 	</build>
 
 	<dependencies>
+	    <!-- Logging -->
+        <dependency>
```
lfarrell: Does this still need a log4j dependency? Doesn't seem to be in the other packages

metadata/src/main/java/edu/unc/lib/dl/xml/FOXMLJDOMUtil.java
bbpennel: This class is going to be deleted, but I needed to clear out the log4j references in it in the meantime.

fcrepo-clients/pom.xml
bbpennel: Don't think it has a log4j dependency?   slf4j is still used as an implementation independent api for logging.

access-common/pom.xml
```diff
@@ -13,19 +13,16 @@
 		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
-			<version>1.6.1</version>
```
harringj: why was (only) the version removed here?

bbpennel: slf4j is being retained, but the version is being set at the root pom instead of in each module pom

metadata/src/main/java/edu/unc/lib/dl/xml/FOXMLJDOMUtil.java
```diff
@@ -138,7 +139,7 @@ public static Element getDatastream(final Document foxml, final String id) {
     }
 
     public static List<Element> getFileLocators(final Document foxml) {
-        List<Element> result = new ArrayList<Element>();
```
harringj: even though this file is about to go away, you just couldn't let it go, could you ;-)

PR #687, opened 2017-12-20
Fix 404 errors for admin actions

lfarrell:


static/js/admin/src/URLUtilities.js
```diff
@@ -58,6 +58,18 @@ define('URLUtilities', ['jquery'], function($) {
 			if (newParameterString)
 				return baseURL + "?" + newParameterString;
 			return baseURL;
+		},
+
+		getBaseUrl: function(url, access) {
+			if (access === undefined || !access) {
+				return url.substring(0, url.indexOf("/admin/") + 7);
```
harringj: not a huge deal, but wonder whether using  length() function here instead of 7 would make for code that is better self-documenting

lfarrell: Updated in new commit

static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -97,7 +97,12 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 		var self = this;
 		var resultObject = $trigger.parents(self.options.containerSelector).data('resultObject');
 		var metadata = resultObject.metadata;
-		
+
+		var fullUrl = document.location.href;
+		var baseUrl = URLUtilities.getBaseUrl(document.location.origin, true);
```
harringj: just curious because i don't understand this fully: why are you using document.location.origin here instead of the value of fullUrl?

static/js/admin/src/action/ChangeLocationAction.js
```diff
@@ -5,15 +5,12 @@ define('ChangeLocationAction', [ 'jquery'], function($) {
 	
 	ChangeLocationAction.prototype.execute = function() {
 		var url;
-		if (this.context.application == "access") {
+		if (this.context.application !== undefined) {
```
harringj: where does this.context.application get set? and does this change mean that as long as some value is stored in that var we want to set the url to accessBaseUrl?

PR #689, opened 2018-01-03
Cdr 2006 verify deposit objects

harringj:


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -212,7 +210,22 @@ public void runJob() {
             failJob(e, "Failed to ingest content for deposit {0}", getDepositPID().getQualifiedId());
         }
 
-        // TODO verify objects present in fcrepo
+        // Verify objects from deposit are present in fcrepo
+        if (!areObjectsInFedora(depositBag)) {
+            failJob("Some objects did not make it to Fedora from this deposit: ", getDepositPID().getQualifiedId());
```
bbpennel: It'd be helpful for debugging purposes which objects didn't make it into fedora.  Also, for string formatting like this you need to put `{0}` into the message where you want to have the variable display

PR #690, opened 2018-01-08
Required solr fcrepo4

lfarrell:


solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/indexing/SolrUpdateDriver.java
```diff
@@ -54,6 +54,16 @@
 
     private static String SET_OPERATION = "set";
     private static String UPDATE_TIMESTAMP = "timestamp";
+    private static String[] REQUIRED_INDEXING_FIELDS = new String[] {
```
bbpennel: Since its the structure used for the rest of the solr config, this should probably be coming in through SolrSettings (and therefore solr.properties).  That would sort of save you from having to duplicate the list in the test (although just making this constant public would do that too).  

A good follow up ticket may also be to make a single solr.properties file that is shared by all the applications.  I'm not sure if search.properties will be possible to do that with.

PR #694, opened 2018-01-11
Trendline

lfarrell:


static/js/admin/performance_visualizations/draw.js
```diff
@@ -114,9 +113,14 @@ CdrGraphs.prototype.draw = function() {
     
     this.statsDisplay("#throughput-stats", throughput_all, throughput);
     var throughput_date = this.showAxises("#throughput-date", xAxis, yAxis, width, "Throughput (MB)");
-    
+
+    // Calculate regression and data for trendline
+    var trendline_data = this.trendLineData(throughput_all, xScale.domain(), "throughput_bytes");
```
cazzerson: Should this use the `throughput` variable like on the next line?

lfarrell: Yes, good catch. Fixed now.

PR #698, opened 2018-01-12
Check for fcrepo availability before starting services worker

lfarrell:


bbpennel:
Didn't this PR originally have similar polling to see if fedora was available when executing jobs?  It would be good to have a similar approach there if a 401 unauthorized response is returned.

Have you tried this on QA?  I suspect that managementClient will be null since the context hasn't been initialized yet.

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/worker/WorkerDaemon.java
```diff
@@ -31,26 +33,53 @@ public void init(DaemonContext context) throws DaemonInitException, Exception {
 
 	@Override
 	public void start() throws Exception {
+		long backoffDelay = 10000;
```
bbpennel: Please make class level static constants for delay and max attempts.

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/worker/WorkerDaemon.java
```diff
@@ -31,26 +33,53 @@ public void init(DaemonContext context) throws DaemonInitException, Exception {
...
+				LOG.warn("Unable to connect to fedora. Retrying starting Worker Daemon. "
+					+ "Retry attempt " + backoffAttempts);
+
+				try {
+					Thread.sleep(backoffDelay * backoffAttempts);
+				} catch (InterruptedException e) {
+					return;
+				}
+
+				if (backoffAttempts == 5) {
```
bbpennel: Use the constant for max number of attempts

bbpennel: Also, if you put a return in here then you wouldn't need to repeat the cut off in the while loop, and could do a `while (true)`

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/processing/ApplyEnhancementServicesJob.java
```diff
@@ -22,70 +25,101 @@
...
-				applyService(service);
-				metricsClient.incrFinishedEnhancement(service.getClass().getName());
-			} catch (EnhancementException e) {
-				LOG.error("Error applying service " + service.getClass().getName() + " to object " + message.getTargetID(), e);
-				metricsClient.incrFailedEnhancement(service.getClass().getName());
-			} catch (Throwable t) {
-				metricsClient.incrFailedEnhancement(service.getClass().getName());
-				throw t;
+
+			while (backoffAttempts <= maxBackoffAttempts) {
```
bbpennel: While looking at this, I started wondering if it'd make sense to have this retry forever when Fedora is not available or throws an authentication/WebServiceIOException exception.  The reason being that the failure isn't an issue with the job, but with the environment, so skipping over the job won't help.  What do you think?  Would it make sense to maybe increase the delay and make it retry forever in the two cases I mentioned?

For startup of the work daemon i think its fine for it to have max retries before giving up since nothing from the queue would be lost.

It'd be kind of nice to get notifications of some sort when these things are failing, but I don't really want to over engineer this since it's not going to be used for that long.  So lets hold off on that until the need comes up.

PR #701, opened 2018-01-17
Cdr 2103 bulk update mods

harringj:


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/XMLExportJob.java
```diff
@@ -0,0 +1,194 @@
...
+            File mdExportFile = File.createTempFile("xml_export", ".xml");
+
+            try (FileOutputStream xfop = new FileOutputStream(mdExportFile)) {
+                xfop.write(exportHeaderBytes);
+
+                XMLOutputter xmlOutput = new XMLOutputter(Format.getRawFormat());
+                for (String pidString : request.getPids()) {
+                    PID pid = PIDs.get(pidString);
+                    ContentObject obj = (ContentObject) repoObjLoader.getRepositoryObject(pid);
+                    BinaryObject mods = obj.getMODS();
```
bbpennel: Two suggestions here:
1) the `<object pid="blah"></object>` tag needs to be present even if there is no MODS for the object.
2) I would suggest verifying the user has access before pulling the MODS or adding the `<object>` tag.  It is a much cheaper operation than pulling the MODS binary from disk.

static/js/admin/src/action/ExportMetadataXMLBatchAction.js
```diff
@@ -58,7 +56,6 @@ define('ExportMetadataXMLBatchAction', [ 'jquery', 'AbstractBatchAction', "tpl!.
 		
 		this.$form.submit(function(e){
 			var email = $("#xml_recipient_email", self.$form).val();
-			var includeChildren = $("#export_xml_include_children", self.$form).prop("checked");
```
bbpennel: We still need the boolean from this checkbox, right?  It is just going in as a parameter now rather than changing the endpoint used

static/js/admin/src/action/ExportMetadataXMLBatchAction.js
```diff
@@ -71,7 +68,7 @@ define('ExportMetadataXMLBatchAction', [ 'jquery', 'AbstractBatchAction', "tpl!.
 			}
 			
 			$.ajax({
-				url : includeChildren? "exportContainerXML" : "exportXML",
```
bbpennel: I think this will need to go to the services application now, which probably just means switching to "/services/exportXML"

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/XMLExportJob.java
```diff
@@ -0,0 +1,197 @@
...
+        this.groups = groups;
+        this.request = request;
+    }
+
+    @Override
+    public void run() {
+        long startTime = System.currentTimeMillis();
+
+        try {
+            GroupsThreadStore.storeGroups(groups);
```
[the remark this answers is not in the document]
harringj: you already fixed this, no?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/ExportXMLController.java
```diff
@@ -0,0 +1,130 @@
...
+
+        return exportXML(exportRequest);
+    }
+
+    private ResponseEntity<Object> exportXML(XMLExportRequest exportRequest) {
+        Map<String, Object> result = new HashMap<>();
+        result.put("action", "export xml");
+
+        try {
+            service.exportXml(GroupsThreadStore.getUsername(), GroupsThreadStore.getGroups(), new XMLExportRequest(
```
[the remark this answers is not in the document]
harringj: done

services/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -270,6 +270,19 @@
       <property name="operationsMessageSender" ref="operationsMessageSender" />
       <property name="modsValidator" ref="modsValidator" />
   </bean>
+  
+  <bean id="xMLExportService" class="edu.unc.lib.dl.cdr.services.processing.XMLExportService" >
```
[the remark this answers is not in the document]
harringj: i'm not sure i agree. i like to keep the context stuff playing by the same rules. what if we added getters later or something? it looks weird, but it's consistent with how we've been naming other things

static/js/admin/src/action/ExportMetadataXMLBatchAction.js
harringj: ok, sorry, i thought you had asked me to remove it

static/js/admin/src/action/ExportMetadataXMLBatchAction.js
harringj: fixed

services/src/test/resources/acl-global-config.properties
```diff
@@ -0,0 +1,3 @@
+cdr.acl.globalRoles.administrator=adminGroup
```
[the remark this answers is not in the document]
harringj: i'm not clear how to do this

PR #705, opened 2018-01-31
Cdr-1902 Implement Move Operation

bbpennel:
* Updates move workflow to work against fcrepo4
  * Adds unit and integration tests
  * Adds move log
  * Currently removes reporting on move operations to users
* Adds helper methods to ObjectPathFactory and ObjectPath classes to make them more usable, adds unit test
* Removes getGroups method from AgentPrincipal class in favor of getPrincipals


etc/solr-config/access/conf/elevate.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
harringj: what is this file for?

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsJob.java
```diff
@@ -0,0 +1,302 @@
+package edu.unc.lib.dl.persist.services.move;
```
harringj: needs license header

etc/solr-config/access/conf/elevate.xml
bbpennel: Its a default solr config file that I copied over from one of the other projects, but it looks like the reference to it is commented out, so we don't need it here, I'll remove it

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsJob.java
bbpennel: fixed

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsJob.java
```diff
@@ -0,0 +1,317 @@
...
+    private RepositoryObjectLoader repositoryObjectLoader;
+    private TransactionManager transactionManager;
+    private SparqlQueryService sparqlQueryService;
+    private FcrepoClient fcrepoClient;
+    private OperationsMessageSender operationsMessageSender;
+    private ObjectPathFactory objectPathFactory;
+    private ActivityMetricsClient operationMetrics;
+
+    private AgentPrincipals agent;
+    private PID destination;
```
harringj: would you mind renaming this "destinationPid" for consistency and clarity?

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsJob.java
```diff
@@ -0,0 +1,317 @@
...
+        }
+        destContainer = (ContentContainerObject) destObj;
+    }
+
+    private void moveObject(PID objPid) {
+        aclService.assertHasAccess("Agent " + agent.getUsername() + " does not have permission to move object "
+                + objPid, objPid, agent.getPrincipals(), Permission.move);
+
+        RepositoryObject moveObj = repositoryObjectLoader.getRepositoryObject(objPid);
+        ContentObject moveContent = (ContentObject) moveObj;
```
harringj: can't lines 149-150 be combined?

bbpennel: updated

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsJob.java
bbpennel: renamed

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsService.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Move a list of objects to the destination container as the provided
+     * agent.
+     *
+     * @param agent
+     * @param destination
+     * @param pids
+     */
+    public String moveObjects(AgentPrincipals agent, PID destination, List<PID> pids) {
```
harringj: destinationPid, please :-)

security/src/main/java/edu/unc/lib/dl/acl/util/AgentPrincipals.java
```diff
@@ -35,10 +34,9 @@
      * @param username
      * @param groups
```
harringj: change param name to "principals"

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
```diff
@@ -49,23 +45,10 @@
...
-
     @Autowired
-    private ActivityMetricsClient operationMetrics;
+    private MoveObjectsService moveObjectsService;
 
     public MoveObjectsController() {
-        moveRequests = new HashMap<>();
     }
 
     @RequestMapping(value = "edit/move", method = RequestMethod.POST)
```
harringj: i think it might be a good idea to have this controller's organization match what we've been doing with our other controllers recently. so, e.g., moveObjects() should call a private method within the controller that would itself return a ResponseEntity. also, at line 60 there is an opening '{' that never gets closed. and while we're at it, maybe remove the unnecessary default constructor for MoveRequest at line 94

solr-search/src/test/java/edu/unc/lib/dl/search/solr/test/TestCorpus.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.stream.Collectors;
+
+import org.apache.solr.common.SolrInputDocument;
+
+import edu.unc.lib.dl.fcrepo4.PIDs;
+import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.util.ResourceType;
+
+/**
+ * Utility which constructs a small corpus of solr documents for testing and
```
harringj: nifty! :-)

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
bbpennel: I usually try to split out to a private method if the content of the method is going to be shared by more than one controller endpoint.  Since there's only one controller method currently and it is just doing controllery stuff, I'm inclined to leave it, but I'd be happy to refactor this later if we have multiple controller methods in this class that need the same functionality though.

Went ahead and removed that default constructor.

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsService.java
bbpennel: fixed

security/src/main/java/edu/unc/lib/dl/acl/util/AgentPrincipals.java
bbpennel: fixed

bbpennel:
I have delete/restore and move over in persistence since they are update operations that could reasonably be used outside of the services web application (services-camel or some other api), and that is the package where update operations have been contained historically (DigitalObjectManager, update pipeline).  Some of our modules aren't very well defined, and persistence could use a revisit.  That is something I am still hoping to get to, possibly before this migration is completed.

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
harringj: ok, that's fine. how would you feel about at least changing the return type to be a ResponseEntity, and to use predefined constants for the response codes? then you could also remove HttpResponse as a param to the controller method. i just think that this ```return new ResponseEntity<>(result, HttpStatus.OK)``` looks much cleaner than this '''response.setStatus(200);``` followed by ```return results;```

bbpennel: Ah good call, i mentally skipped over that big in the first comment, sorry.  Should be addressed now

PR #706, opened 2018-02-01
BXC-2107 Queries for populating front page take into account user permissions

bbpennel:


access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -1014,26 +1017,24 @@ public void setCollectionsPid(PID collectionsPid) {
...
 		try {
-			response = this.executeQuery(query);
+			StringBuilder queryBuilder = new StringBuilder("*:*");
+			addAccessRestrictions(queryBuilder, accessGroups);
+
+			SolrQuery query = new SolrQuery();
+			query.setQuery(queryBuilder.toString());
+			query.setRows(0);
+			query.addFacetField("department");
+			query.setFacetLimit(-1);
```
harringj: what does -1 signal here? i don't see a good explanation of this in the solr docs

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -1014,26 +1017,24 @@ public void setCollectionsPid(PID collectionsPid) {
...
+			
+			QueryResponse response = this.executeQuery(query);
 			return response.getFacetField("department").getValueCount();
-		} catch (SolrServerException e) {
-			LOG.error("Error retrieving Solr object request: " + e);
+		} catch (SolrServerException | AccessRestrictionException e) {
+			LOG.error("Error retrieving department list", e);
 		}
 
 		return -1;
```
harringj: can this return statement be moved inside the catch-block, or will the compiler complain? i think it would make for clearer code to move it inside, since that return value only gets returned if the catch-block is executed

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -1110,21 +1108,20 @@ public long getCollectionsCount() {
...
-		QueryResponse response;
-
-		query = new SolrQuery();
-		query.setQuery("*:*");
-		query.setRows(0);
-		query.addFacetField("contentType");
-		query.setFacetLimit(-1);
-
+	public Map<String, Long> getFormatCounts(AccessGroupSet accessGroups) {
 		HashMap<String, Long> counts = new HashMap<String, Long>();
```
harringj: are we not using the newer syntax for declaring collections variables?

HashMap<String, Long> counts = new HashMap<>();

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -1110,21 +1108,20 @@ public long getCollectionsCount() {
...
-
+		
 		try {
-			response = this.executeQuery(query);
+			StringBuilder queryBuilder = new StringBuilder("*:*");
```
harringj: i wouldn't mind seeing a private function that encompasses the logic in lines 1115-1120, and their equivalents above. it could take a string param to pass to the StringBuilder constructor

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
bbpennel: It means unlimited, the default is some rather low number

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
bbpennel: Trying to think of how that would be named, `makeSolrQueryWithAccessRestrictions`?  Hmm, I feel like `addAccessRestrictions` may already be an encapsulation of the common functionality here.  Often the two builders (StringBuilder and SolrQuery) would both need to be available to the method constructing a query, although query construction logic in these three cases is pretty simple.  I'm leaning towards leaving this as it is for now, but I would like to refactor this class significantly in the future since it doesn't have a real purpose, its basically a functional dumping ground of queries at present.

bbpennel:
I would like to have tests for these, but I don't want to try to work those into master at this point.  I need to repeat these changes on fcrepo4 as well, but I think I may want to wait on adding tests to this until we've had a chance to think about the structure of this class a bit more.

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
harringj: ok, sounds reasonable

PR #707, opened 2018-02-02
F4 BXC-2107 Front Page Queries

bbpennel:


access/src/main/java/edu/unc/lib/dl/ui/controller/FrontPageController.java
```diff
@@ -41,11 +43,12 @@
     @RequestMapping(method = RequestMethod.GET)
     public String handleRequest(Model model, HttpServletRequest request) {
         LOG.debug("In front page controller");
+        AccessGroupSet groups = GroupsThreadStore.getGroups();
```
harringj: you may have answered this before, but aren't we switching to using principals instead of groups?

bbpennel: We haven't changed the language in the `GroupsThreadStore` class at this point, although perhaps we should.  Also, the AgentPrincipals class hasn't made it into the solr api so far, but we should keep that in mind

harringj: yeah, changing the language might not be a bad idea. otherwise, people like me are going to continually be confused about whether we're using groups or principals :-)

bbpennel: true, we are still using groups since they are a subset of principals, which also includes whatever other authentication tokens we have flying around.  Currently its groups, authentication status, and usernames, but will potentially expand to others

PR #710, opened 2018-02-13
Enhancement failure recovery adjustments

bbpennel:


services-worker/src/main/java/edu/unc/lib/dl/cdr/services/processing/ApplyEnhancementServicesJob.java
```diff
@@ -95,32 +97,34 @@ private void applyService(ObjectEnhancementService service) throws EnhancementEx
...
 				
 				service.getEnhancement(message).call();
 				message.getCompletedServices().add(service.getClass().getName());
-
 				break;
 			} catch (EnhancementException e) {
-				if (e.getSeverity() == EnhancementException.Severity.RECOVERABLE) {
-					LOG.error("Retrying service for recoverable exception: " + service.getClass().getName(), e);
+				if (backoffAttempts < MAX_RETRIES && e.getSeverity() == Severity.RECOVERABLE) {
+					LOG.warn("Retrying service for recoverable exception {}, attempt {}: {}: ",
```
harringj: i think there's an extra {}: in the logging string

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/techmd/TechnicalMetadataEnhancement.java
```diff
@@ -257,10 +258,10 @@ public Element call() throws EnhancementException {
 					(System.currentTimeMillis() - startJob));
 		} catch (FileSystemException e) {
 			throw new EnhancementException(e, Severity.FATAL);
-		} catch (NotFoundException e) {
-			throw new EnhancementException(e, Severity.UNRECOVERABLE);
-		} catch (FedoraException e) {
+		} catch (ObjectExistsException | OptimisticLockException e) {
 			throw new EnhancementException(e, Severity.RECOVERABLE);
+		} catch (FedoraException e) {
```
harringj: should this be limited to catching a FedoraException? is it impossible that an exception other than FileSystemException, ObjectExistsException, or OptimisticLockException could get thrown by this method?

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/text/FullTextEnhancement.java
```diff
@@ -120,10 +122,9 @@ public Element call() throws EnhancementException {
 			}
 		} catch (FileSystemException e) {
 			throw new EnhancementException(e, Severity.FATAL);
-		} catch (NotFoundException e) {
-			throw new EnhancementException(e, Severity.UNRECOVERABLE);
-		} catch (FedoraException e) {
-			throw new EnhancementException("Full Text Enhancement failed to process " + dsid, e, Severity.RECOVERABLE);
+		} catch (ObjectExistsException | OptimisticLockException e) {
+			throw new EnhancementException("Failed to update object " + pid + " with " + dsid,
```
harringj: maybe use string-replacement here, too, to match the way you wrote the other catch-blocks?

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/techmd/TechnicalMetadataEnhancement.java
bbpennel: Only FedoraException is explicitly thrown by the code, but there do seem to be some uncaught spring exceptions that come out of there and the other enhancements are catching Exception, so I'll bring it in line with the others.

services-worker/src/main/java/edu/unc/lib/dl/cdr/services/text/FullTextEnhancement.java
bbpennel: EnhancementException doesn't have a string formatting method, that's mainly available through the logger

PR #715, opened 2018-03-21
Fix static import path in admin and base url issues

bbpennel:


access/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -67,6 +67,11 @@
             value="${google.trackingId}" />
     </bean>
     
+    <bean id="accessBaseUrl" class="java.lang.String">
+        <constructor-arg
```
harringj: a nitpick for sure, but does the constructor-arg need to be spread over two lines?

bbpennel: All of the `constructor-args` for these string beans should be single lined now

PR #716, opened 2018-03-27
Bxc-2141-4 Services config fixes

bbpennel:
Gets services webapp deployable on FES and deals with issues related to it being deployed in multi-server environment.

* Adds optional Host header to requests to Fedora to deal with cross domain requests (by default, fedora responds with URIs that match the URI it was addressed at.  So if one service addresses fedora at repo.example.com and another as localhost, they will get back different URIs and triples) that was impacting indexing to triple store.
* Adds missing initialization of move and acl retrieval services
* Adjusts property names to match deployed values

### DIFF
--- a/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FcrepoClientFactory.java
+++ b/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FcrepoClientFactory.java
@@ -29,10 +29,17 @@ public class FcrepoClientFactory {
 
     private String baseUri;
 
+    private String hostHeader;
+
     private static FcrepoClientFactory instance;
 
     private FcrepoClientFactory(String base) {
         baseUri = base;
+    }
+
+    private FcrepoClientFactory(String base, String hostHeader) {
+        baseUri = base;
+        this.hostHeader = hostHeader;
     }
 
     /**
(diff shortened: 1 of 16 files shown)


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FcrepoClientFactory.java
```diff
@@ -83,6 +103,7 @@ public FcrepoClient makeAuthenticatedClient(String host, String user, String pas
                 .credentials(user, password)
                 .authScope(host)
                 .throwExceptionOnFailure()
+                .hostHeader(hostHeader)
```
harringj: what happens here when hostHeader is null?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/TransactionalFcrepoClient.java
```diff
@@ -56,19 +58,27 @@
...
     protected TransactionalFcrepoClient(String username, String password, String host,
                 Boolean throwExceptionOnFailure, String baseUri) {
         super(username, password, host, throwExceptionOnFailure);
         txBasePattern = Pattern.compile(baseUri + TX_ID_REGEX);
         txRemovePattern = Pattern.compile(TX_RESPONSE_REGEX);
     }
 
+    protected TransactionalFcrepoClient(String username, String password, String host,
+            Boolean throwExceptionOnFailure, String baseUri, String hostHeader) {
+    this(username, password, host, throwExceptionOnFailure, baseUri);
```
harringj: needs indent

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/TransactionalFcrepoClient.java
```diff
@@ -134,7 +148,11 @@ private URI rewriteUri(URI rescUri) {
         URI txUri = FedoraTransaction.txUriThread.get();
         String rescId = rescUri.toString();
         int txIdIndex = rescId.indexOf(REST);
-        rescId = rescId.substring(txIdIndex + REST.length());
+        int rescPathIndex = txIdIndex + REST.length();
```
harringj: i'm not sure i completely follow what's going on in lines 151-155. maybe a comment would be helpful. if i'm understanding correctly, if there's something left in the uri beyond the txid itself, these lines will join that something to the txUri?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FcrepoClientFactory.java
bbpennel: the client built by the factory won't add the host header

PR #717, opened 2018-03-29
Bxc 2146 - Reindexing of Triples

bbpennel:
Adds API endpoint for reindexing of triples
Adds camel routes for reindexing of triples
Allows for creation of PID objects for some reserved objects, specially the root of the repository and the content container.


access/src/test/resources/server.properties
```diff
@@ -58,10 +58,6 @@ solr.context=solr
 # Access control
 access.group.admin=admin
 
-# Java Message Server (localhost, 61616)
-jms.host=localhost
-jms.port=:61617
```
harringj: odd that this port number didn't match up with the comment, or with the jms.port in deposit.properties

etc/server.properties
```diff
@@ -46,9 +46,8 @@ redis.port=6379
 # Access control
 access.group.admin=unc:app:lib:cdr:admin
 
-# Java Message Server (localhost, 61616)
-jms.host=localhost
-jms.port=:61617
+# Java Message Server
+jms.brokerUrl=tcp://localhost:61617
```
harringj: should this be 61617 or 61616? i'm just not clear why the old comment had a different port listed

access/src/test/resources/server.properties
bbpennel: Yeah, one got changed without the other being updated.  I believe we are using 61617 for the VM, which is slightly off from the default jms port.

etc/services-camel.properties
```diff
@@ -57,6 +60,7 @@ input.stream=direct-vm:index.start
 
 # The camel URI for handling reindexing events.
 triplestore.reindex.stream=activemq:queue:triplestore.reindex
```
harringj: should this uri be deleted in favor of the one you added?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPathConstants.java
```diff
@@ -71,6 +73,14 @@
...
 
     // Regex pattern for decomposing a repository URI for an object or component of an object
+    // Group 1 = transaction id
+    // Group 2 = base container
+    // Group 3 = expanded object id
+    // Group 4 = hashed path to object
+    // Group 5 = object uuid
+    // Group 6 = reserved object id
+    // Group 7 = component path (nested resources in object)
+    // Group 8 = relative component path
```
harringj: 💯

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPaths.java
```diff
@@ -92,12 +126,19 @@ public static String getVocabulariesBase() {
...
+     * @param uri base uri for repository
+     */
     private static void setContentBase(String uri) {
         baseUri = uri;
         if (!baseUri.endsWith("/")) {
             baseUri += "/";
         }
-        contentBase = URIUtil.join(baseUri, RepositoryPathConstants.CONTENT_BASE);
-        contentRootPid = PIDs.get(URIUtil.join(contentBase, RepositoryPathConstants.CONTENT_ROOT_ID));
+        rootPid = new FedoraPID(REPOSITORY_ROOT_ID, REPOSITORY_ROOT_ID, null, URI.create(baseUri));
```
harringj: why is REPOSITORY_ROOT_ID being passed in here twice?

services-camel/src/main/java/edu/unc/lib/dl/services/camel/triplesReindexing/IndexingMessageProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+        final Message in = exchange.getIn();
+
+        Document msgBody = MessageUtil.getDocumentBody(in);
+        Element body = msgBody.getRootElement();
+
+        String pidValue = body.getChild("pid", ATOM_NS).getTextTrim();
+        String action = body.getChild("actionType", ATOM_NS).getTextTrim();
+        IndexingActionType actionType = IndexingActionType.getAction(action);
+
+        PID pid = PIDs.get(pidValue);
```
harringj: a small thing, but would you consider moving this line under line 52. that organization strikes me as being more logical, since those two lines are used to get the pid, while lines 53-54 are used to get the actionType

services-camel/src/main/java/edu/unc/lib/dl/services/camel/triplesReindexing/TriplesReindexingRouter.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @BeanInject(value = "indexingMessageProcessor")
+    private IndexingMessageProcessor indexingMessageProcessor;
+
+    @Override
+    public void configure() throws Exception {
+        onException(Exception.class)
+            .redeliveryDelay("{{error.retryDelay}}")
+            .maximumRedeliveries("{{error.maxRedeliveries}}")
+            .backOffMultiplier(2)
```
harringj: would it make sense to capture the backoff multiplier value in a property as well?

services-camel/src/test/resources/config.properties
```diff
@@ -43,8 +43,12 @@ cdr.stream.camel=direct-vm:index.start
 cdr.solrupdate.stream=direct-vm:index.start
 cdr.solrupdate.stream.camel=direct-vm:index.start
 
+cdr.triplesupdate.stream=direct-vm:index.start
```
harringj: what's the difference between these two properties? do we really need both?

services-camel/src/test/resources/triples-reindexing-it-config.properties
```diff
@@ -0,0 +1,28 @@
+fcrepo.baseUrl=http://localhost:8080/fcrepo/rest
+fcrepo.baseUri=localhost:8080/fcrepo/rest
+fcrepo.serverUri=http://localhost:8080/
```
harringj: i wish we could get more consistent about how we name URLs and URIs throughout our codebase. seems like this one should be serverUrl to match baseUrl

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
```diff
@@ -48,35 +48,52 @@
...
+     * @param id the identifier of the object to be reindexed
+     * @return
+     */
+    @RequestMapping(value = "edit/triples/reindex/{id}", method = RequestMethod.POST)
+    public ResponseEntity<Object> reindexTriples(@PathVariable("id") String id) {
+        return indexObjectAndChildren(id, false, triplesIndexingService, "reindexTriples");
     }
 
-    private ResponseEntity<Object> indexObject(String id) {
+    private ResponseEntity<Object> indexObject(String id, IndexingService indexingService, String action) {
```
harringj: why is "action" being passed in as a param here? wouldn't the action be "reindex triplestore"?

etc/services-camel.properties
bbpennel: Its a bit confusing, but they are separate routes. `cdr.triplesupdate.stream.camel` is where requests for reindexing actions comes from (from our clients). `triplestore.reindex.stream` is the internal route in the reindexing router that crawls the hierarchy and triggers indexing of objects.

harringj: oh, ok, thanks for explaining! would it be worth putting a comment to this effect in the properties file?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPaths.java
bbpennel: FedoraPIDs are usually constructed with an id and a qualifier. The id is the actual uuid or other identifier for the object. The qualifier is usually a root container in fedora (like "content" or "deposit") that groups together objects of the same purpose. In this branch I introduced PIDs for those root objects (see contentBasePid). None of those are in a subcontainer, so I introduced the REPOSITORY_ROOT_ID to use indicate that they are at the root rather than guessing if a blank string was intentional or not. It also adds rootPid, and I ended up deciding to give it REPOSITORY_ROOT_ID as its qualifier too, but it may be okay to just give it a blank for the qualifier, not sure.

services-camel/src/test/resources/config.properties
bbpennel: the `.stream` versus `.stream.camel` properties? In this environment they are the same, but in most environments they are different.

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/IndexingController.java
bbpennel: Yes, the controller methods pass in different values depending on what is being indexed (`reindexSolr`, `reindexTriples`, `updateSolr`), but share the same methods for kicking off the indexing request. Its a bit unnecessary in this case because there isn't an endpoint for indexing a single object in the triple store, I just added it for consistency so that the controller method was where the action name was always coming from. I can remove it if you'd prefer though.

services-camel/src/test/resources/triples-reindexing-it-config.properties
bbpennel: Yeah, there isn't really any clear distinction, I think we will have to clean them up everywhere at once in another ticket

services-camel/src/main/java/edu/unc/lib/dl/services/camel/triplesReindexing/IndexingMessageProcessor.java
bbpennel: Done

services-camel/src/main/java/edu/unc/lib/dl/services/camel/triplesReindexing/TriplesReindexingRouter.java
bbpennel: Added

PR #718, opened 2018-04-03
Bxc 2149 - Permissions in Access UI

bbpennel:
Gets basic permissions working in the access interface.  Download links still do not work, pending future tickets.

* Refactors views to use ACL service via an instantiated helper class.  Removes previous static methods for acl checks.
* Uses ACL service to check permissions for full record and mappings refreshes.
* Updates to filter which extracts ACLs and storage to thread.
* Disables FedoraContentService behavior for future work
* Removes behaviors related to List Only view, which is no longer being used.


access-common/src/main/java/edu/unc/lib/dl/ui/controller/LorisContentController.java
```diff
@@ -39,10 +46,9 @@
 
 //    @Autowired
```
harringj: is there a reason this field is commented out?

access-common/src/main/java/edu/unc/lib/dl/ui/controller/LorisContentController.java
```diff
@@ -108,8 +114,10 @@ public void getRegion(@PathVariable("id") String id,
     @RequestMapping("/jp2Proxy/{id}/{datastream}")
     public void getMetadata(@PathVariable("id") String id,
             @PathVariable("datastream") String datastream, HttpServletResponse response) {
+
+        PID pid = PIDs.get(id);
         // Check if the user is allowed to view this object
-        if (this.hasAccess(id, datastream)) {
+        if (this.hasAccess(pid, datastream)) {
 //            try {
```
harringj: can this stuff be removed?

access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
```diff
@@ -142,51 +145,49 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
...
+            if (briefObject.getResourceType().equals(searchSettings.resourceTypeCollection)) {
+                facetsToRetrieve = new ArrayList<>(searchSettings.collectionBrowseFacetNames);
+            } else if (briefObject.getResourceType().equals(searchSettings.resourceTypeAggregate)) {
+                facetsToRetrieve = new ArrayList<>();
+                facetsToRetrieve.add(SearchFieldKeys.CONTENT_TYPE.name());
+            }
 
-                briefObject.getCountMap().put("child", resultResponse.getResultCount());
+            LOG.debug("Retrieving supplemental information for container at path "
+            + briefObject.getPath().toString());
```
harringj: indent?

access/src/main/java/edu/unc/lib/dl/ui/controller/RefreshMappingsController.java
```diff
@@ -47,10 +50,13 @@
...
 
+    @Autowired
+    private GlobalPermissionEvaluator globalPermissionEvaluator;
+
     @RequestMapping(method = RequestMethod.GET)
     public void handleRequest(HttpServletRequest request, HttpServletResponse response) throws Exception {
-        AccessGroupSet accessGroups = getUserAccessGroups(request);
-        if (!accessGroups.contains(AccessGroupConstants.ADMIN_GROUP)) {
+        AccessGroupSet principals = getAgentPrincipals().getPrincipals();
+        if (globalPermissionEvaluator.hasGlobalPermission(principals, Permission.reindex)) {
```
harringj: i'm confused. why are we throwing an exception when there IS a global permission?

access-common/src/main/java/edu/unc/lib/dl/ui/controller/LorisContentController.java
bbpennel: It will probably end up being used once we have loris functioning again, so I left it behind as a placeholder

access-common/src/main/java/edu/unc/lib/dl/ui/controller/LorisContentController.java
bbpennel: I was generally leaving this class commented out until we work on loris integration again since I'm guessing a lot of parts of it will be the same but haven't been updated yet

security/src/main/java/edu/unc/lib/dl/acl/util/AgentPrincipals.java
```diff
@@ -36,8 +37,12 @@
      */
     public AgentPrincipals(String username, AccessGroupSet principals) {
         this.username = username;
-        this.principals = new AccessGroupSet(principals);
-        if (username != null) {
+        if (principals == null) {
+            this.principals = new AccessGroupSet();
+        } else {
+            this.principals = new AccessGroupSet(principals);
```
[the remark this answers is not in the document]
bbpennel: It modifies the the set of principals at line 46 by adding in a principal for the user, so I didn't want to modify the input.

I'm trying to remember why I didn't simply add the user principal to the set of groups when GroupsThreadStore is populated in the first place. If I recall correctly, I was thinking that we would want to be able to differentiate between the actual groups assigned and the other types of principals (username, maybe IP address in the future), but to also be able to retrieve them all together. Feels a bit messy currently, but not immediately sure of what to do about it.

PR #721, opened 2018-04-09
Bxc-2148 - Structure fix and refactor

bbpennel:
* Fixes issues regarding display and retrieval of structure browse
* Refactors ChildrenCount and Structure queries into their own services with expanded integration tests.
* Deletes a lot of unused configuration and code related to structure queries.

### DIFF
--- a/access/src/main/java/edu/unc/lib/dl/ui/controller/StructureBrowseController.java
+++ b/access/src/main/java/edu/unc/lib/dl/ui/controller/StructureBrowseController.java
@@ -35,10 +35,9 @@ import edu.unc.lib.dl.search.solr.util.SearchStateUtil;
 import edu.unc.lib.dl.ui.util.SerializationUtil;
 
 /**
- * Handles requests for the heirarchical structure browse view. The request may either be for an entire stand alone
- * view, or if the ajax option is true then a portion of the tree starting from the room node. The request can specify
- * the max depth in terms of nodes in the tree it will return, where the max depth is limited by the application wide
- * structured depth property.
+ * Handles requests for the hierarchical structure-browse view. The request may
+ * either be for an entire stand-alone view or, if the ajax option is true, for
+ * a portion of the tree starting from the root node.
  *
  * @author bbpennel
  */
--- a/solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/HierarchicalBrowseResultResponse.java
+++ b/solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/HierarchicalBrowseResultResponse.java
@@ -23,7 +23,6 @@ import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
 
-import org.apache.solr.client.solrj.response.FacetField;
 import org.apache.solr.common.SolrDocument;
 import org.apache.solr.common.SolrDocumentList;
 import org.slf4j.Logger;
@@ -39,14 +38,12 @@ import edu.unc.lib.dl.util.ResourceType;
 public class HierarchicalBrowseResultResponse extends SearchResultResponse {
     protected static final Logger log = LoggerFactory.getLogger(HierarchicalBrowseResultResponse.class);
 
-    private Map<String, Long> subcontainerCounts;
     private Set<String> matchingContainerPids = null;
     private Long rootCount;
     private ResultNode rootNode;
 
     public HierarchicalBrowseResultResponse() {
         super();
-        subcontainerCounts = new HashMap<>();
         matchingContainerPids = new HashSet<>();
         setResultList(new ArrayList<>());
     }
@@ -57,30 +54,6 @@ public class HierarchicalBrowseResultResponse extends SearchResultResponse {
         this.setGeneratedQuery(response.getGeneratedQuery());
         this.setResultList(response.getResultList());
         this.setSearchState(response.getSearchState());
-    }
-
-    public Map<String, Long> getSubcontainerCounts() {
-        return subcontainerCounts;
-    }
-
-    public void setSubcontainerCounts(Map<String, Long> subcontainerCounts) {
-        this.subcontainerCounts = subcontainerCounts;
-    }
-
-    public void populateSubcontainerCounts(List<FacetField> facetFields) {
-        subcontainerCounts = new HashMap<>();
-        for (FacetField facetField : facetFields) {
-            if (facetField.getValues() != null) {
-                for (FacetField.Count facetValue : facetField.getValues()) {
-                    log.debug("Popsub|" + facetValue.getName() + ":" + facetValue.getCount());
-                    int index = facetValue.getName().indexOf(",");
-                    index = facetValue.getName().indexOf(",", index + 1);
-                    if (index != -1) {
-                        subcontainerCounts.put(facetValue.getName().substring(0, index), facetValue.getCount());
-                    }
-                }
-            }
-        }
     }
 
     public void removeContainersWithoutContents() {
@@ -99,18 +72,6 @@ public class HierarchicalBrowseResultResponse extends SearchResultResponse {
                     log.debug("Removing container " + briefObject.getId()
                             + "from hierarchical result because it has no children");
                     resultIt.remove();
-                    // If an item is being filtered out, then decrement the counts for it and all its ancestors in
-                    // subcontainer counts
-                    if (briefObject.getAncestorPathFacet() != null
-                            && briefObject.getAncestorPathFacet().getFacetNodes() != null) {
-                        for (HierarchicalFacetNode facetTier : briefObject.getAncestorPathFacet().getFacetNodes()) {
-                            String tierIdentifier = facetTier.getSearchValue();
-                            Long count = this.subcontainerCounts.get(tierIdentifier);
-                            if (count != null) {
-                                this.subcontainerCounts.put(tierIdentifier, count - 1);
-                            }
-                        }
-                    }
                 }
             }
         }
--- a/solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
+++ b/solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
@@ -60,15 +60,6 @@ public class ChildrenCountService extends AbstractQueryService {
     public long getChildrenCount(BriefObjectMetadata container, AccessGroupSet principals) {
         SolrQuery solrQuery = createBaseQuery(principals);
 
-        // Add access restrictions to query
-        StringBuilder query = new StringBuilder("*:*");
-        restrictionUtil.add(query, principals);
-        solrQuery.setQuery(query.toString());
-
-        solrQuery.setStart(0);
-        solrQuery.setRows(0);
-        solrQuery.setFacet(true);
-
         StringBuilder filterQuery = new StringBuilder();
         addFilter(filterQuery, ANCESTOR_PATH, container.getPath().getSearchValue());
         solrQuery.addFilterQuery(filterQuery.toString());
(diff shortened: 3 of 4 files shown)


access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
```diff
@@ -216,7 +219,7 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
 
 //        if (briefObject.getResourceType().equals(searchSettings.resourceTypeCollection)
 //                || briefObject.getResourceType().equals(searchSettings.resourceTypeFolder)) {
-//            applyContainerSettings(pidString, foxmlView, model, fullObjectView != null);
+//            applyContainerSettings(pidString, principals, foxmlView, model, fullObjectView != null);
```
harringj: i'm guessing this will be turned back on at some point in the future?

access/src/main/java/edu/unc/lib/dl/ui/controller/StructureBrowseController.java
```diff
@@ -28,22 +28,18 @@
...
 import edu.unc.lib.dl.search.solr.model.SearchState;
-import edu.unc.lib.dl.search.solr.model.SimpleIdRequest;
 import edu.unc.lib.dl.search.solr.util.SearchFieldKeys;
 import edu.unc.lib.dl.search.solr.util.SearchStateUtil;
-import edu.unc.lib.dl.search.solr.model.HierarchicalBrowseResultResponse;
-import edu.unc.lib.dl.ui.exception.ResourceNotFoundException;
 import edu.unc.lib.dl.ui.util.SerializationUtil;
 
 /**
  * Handles requests for the heirarchical structure browse view. The request may either be for an entire stand alone
```
harringj: i think this comment could profit from some extra punctuation, and the fixing of a typo or two. what do you think of this: "Handles requests for the hierarchical structure-browse view. The request may either be for an entire stand-alone view or, if the ajax option is true, for a portion of the tree starting from the root node. The request can specify the max depth in terms of nodes in the tree it will return, where the max depth is limited by the application-wide structured-depth property."

services-camel/src/test/resources/solr-update-it-context.xml
```diff
@@ -229,12 +229,17 @@
...
+    <bean id="solrAccessRestrictionUtil" class="edu.unc.lib.dl.search.solr.util.AccessRestrictionUtil">
+        <property name="searchSettings" ref="searchSettings" />
+        <property name="disablePermissionFiltering" value="true" />
+    </bean>
+    
     <bean id="solrSearchService" class="edu.unc.lib.dl.search.solr.service.SolrSearchService">
         <property name="solrSettings" ref="solrSettings" />
         <property name="searchSettings" ref="searchSettings" />
         <property name="facetFieldUtil" ref="facetFieldUtil" />
         <property name="solrClient" ref="embeddedSolrServer" />
```
harringj: i'm noticing that this bean has one more property `<property name="solrClient" ref="embeddedSolrServer" />` than the SolrSearchService bean in admin/src/main/webapp/WEB-INF/solr-search-context.xml. why is this?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/HierarchicalBrowseResultResponse.java
```diff
@@ -67,7 +68,7 @@ public void setSubcontainerCounts(Map<String, Long> subcontainerCounts) {
     }
 
     public void populateSubcontainerCounts(List<FacetField> facetFields) {
-        subcontainerCounts = new HashMap<String, Long>();
+        subcontainerCounts = new HashMap<>();
```
harringj: since subcontainerCounts is already initialized in the constructor, do you also want it set to a fresh hashMap here?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/HierarchicalBrowseResultResponse.java
```diff
@@ -86,11 +87,12 @@ public void removeContainersWithoutContents() {
...
         while (resultIt.hasPrevious()) {
             BriefObjectMetadata briefObject = resultIt.previous();
-            if (briefObject == null || briefObject.getContentModel() == null) {
+            if (briefObject == null || briefObject.getResourceType() == null) {
                 continue;
             }
+            String resourceType = briefObject.getResourceType();
             if ((!briefObject.getCountMap().containsKey("child") || briefObject.getCountMap().get("child") == 0)
-                    && briefObject.getContentModel().contains(ContentModelHelper.Model.CONTAINER.toString())) {
+                    && !ResourceType.File.equals(resourceType)) {
```
harringj: why was this part of the condition changed to check for the type 'File', when before it checked for 'CONTAINER'?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
```diff
@@ -0,0 +1,228 @@
...
+        } catch (SolrServerException e) {
+            throw new SolrRuntimeException(e);
+        }
+    }
+
+    /**
+     * Adds a count of the number of children contained by each object in result
+     * list.
+     *
+     * This count is stored in the containers count map under the "child" key.
```
harringj: "container's"

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
```diff
@@ -0,0 +1,228 @@
...
+    public void addChildrenCounts(List<BriefObjectMetadata> containers, AccessGroupSet principals) {
+        addChildrenCounts(containers, principals, CHILD_COUNT, null);
+    }
+
+    /**
+     * Adds a count of the number of children contained by each object in result
+     * list.
+     *
+     * The count will be calculated based on any restrictions from baseQuery if
+     * it is provided. The count will be stored in each containers count map
```
harringj: "container's"

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
```diff
@@ -0,0 +1,228 @@
...
+        if (baseQuery == null) {
+            // Create a base query since we didn't receive one
+            solrQuery = createBaseQuery(principals);
+        } else {
+            // Starting from a base query
+            solrQuery = baseQuery.getCopy();
+            // Make sure we aren't returning any normal results
+            solrQuery.setRows(0);
+            // Remove all facet fields so we are only getting ancestor path
+            if (solrQuery.getFacetFields() != null) {
```
harringj: can we store the result of `solrQuery.getFacetFields()` in a variable to avoid repeated calls?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
```diff
@@ -0,0 +1,228 @@
...
+        try {
+            QueryResponse queryResponse = this.executeQuery(solrQuery);
+            assignChildrenCounts(queryResponse.getFacetField(ancestorPathField), containers, countKey);
+        } catch (SolrServerException e) {
+            throw new SolrRuntimeException(e);
+        }
+    }
+
+    private SolrQuery createBaseQuery(AccessGroupSet principals) {
+        SolrQuery solrQuery = new SolrQuery();
```
harringj: since lines 160-169 are the same as lines 61-70, can you create a private method to set up the query?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
```diff
@@ -0,0 +1,228 @@
...
+    private void assignChildrenCounts(FacetField ancestorsField, List<BriefObjectMetadata> containerObjects,
+            String countName) {
+        if (ancestorsField.getValues() == null) {
+            return;
+        }
+
+        // Get the list of counts per pid
+        List<Count> counts = ancestorsField.getValues();
+        // Determine the most efficient algorithm for searching the counts
+        boolean binarySearch = counts.size() > 64;
```
harringj: why 64?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/StructureQueryService.java
```diff
@@ -0,0 +1,351 @@
...
+
+        AccessGroupSet principals = browseRequest.getAccessGroups();
+
+        // Start hierarchy tree at root of the repository
+        BriefObjectMetadata contentRootMd = getContentRootMetadata(principals);
+        browseResponse.getResultList().add(contentRootMd);
+
+        // Get the path of the root object being requested for iteration
+        CutoffFacet path = getObjectPath(browseRequest.getRootPid(), principals);
+        if (path.getFacetNodes() == null) {
```
harringj: can we store `path.getFacetNodes()' in a var?

access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
bbpennel: It needs to be updated... unless hongs work determines that we don't need this feature anymore

access/src/main/java/edu/unc/lib/dl/ui/controller/StructureBrowseController.java
bbpennel: 👍 I'm probably going to delete the bit about max depth too since that feature is not used anywhere anymore and my refactor in this PR removes some of the complexity related to it (although we could re-add it without too much difficulty)

services-camel/src/test/resources/solr-update-it-context.xml
bbpennel: I wouldn't mind simplifying this at some point, but solrClient can either be injected as a property or get initialized in the initializeSolrServer method (now found in `AbstractQueryService`). The solrClient is being provided by another client (`solrSettings`) in an unsuccessful attempt to centralize creation of the solrClient, but I don't think its a good pattern. It wasn't related to this PR though so I was trying to control my refactor urges.

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/HierarchicalBrowseResultResponse.java
bbpennel: ContentModel is a fedora3 concept, so was removing that. It was changed from checking that a thing was a container to checking that a thing was NOT a file, which are roughly the same for this purpose.

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
bbpennel: Oops, `createBaseQuery` is actually already the private method that was supposed to replace 61-70, and is being used at line 61, but i forgot to delete the redundant lines

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/ChildrenCountService.java
bbpennel: I can't actually remember where 64 came from. My recollection from when this was written was that when I was looking up the binary search method there was some recommendation not to use it for data sets below a certain point (maybe they said 64). But to be honest it may not be worth the readability hit having two searches.

When I was working on this refactoring I played around with a way shorter implementation that basically called `getChildrenCount` for each container individually instead of doing the one mega query and sorting out the counts. It was way less code, but I couldn't tell from my testing if there was a performance hit since I couldn't load enough items into solr to find out (embedded server would die about 120000 records). I opted not to change it if there wasn't a clear benefit since I assume I implemented more complex approach because of scaling issues. Once we integrate metrics tracking, I'm hoping to have a measurement for this method so we can track if its a problem.

harringj: 👍

PR #723, opened 2018-04-11
Bxc-2164 slow neighbors

bbpennel:
Adds test for neighbor query
Refactors neighbor query to be faster for large numbers of siblings.

### DIFF
--- a/access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
+++ b/access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
@@ -304,7 +304,7 @@ public class SolrQueryLayerService extends SolrSearchService {
 	 *           Access groups of the user making this request.
 	 * @return
 	 */
-	public List<BriefObjectMetadataBean> getNeighboringItems(BriefObjectMetadataBean metadata, int windowSize,
+	public List<BriefObjectMetadataBean> getNeighboringItems(BriefObjectMetadata metadata, int windowSize,
 			AccessGroupSet accessGroups) {
 
 		// Get the common access restriction clause (starts with "AND ...")
@@ -382,7 +382,7 @@ public class SolrQueryLayerService extends SolrSearchService {
 		}
 
 		if (right >= total) {
-			left -= (right - total) + 1;
+			left -= (right - total);
 			if (left < 0) {
 				left = 0;
 			}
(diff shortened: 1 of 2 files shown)


access-common/src/test/java/edu/unc/lib/dl/ui/service/NeighborQueryTest.java
```diff
@@ -0,0 +1,317 @@
...
+		docs.add(newDoc);
+		
+		newDoc = makeContainerDocument(folderPid, "Collection", ResourceType.Folder, rootPid, collectionPid);
+		addAclProperties(newDoc, "public", "admin");
+		docs.add(newDoc);
+
+		return docs;
+	}
+	
+	public SolrInputDocument makeContainerDocument(PID pid, String title, ResourceType type, PID... ancestors) {
```
harringj: why is this method (and the four following) public?

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -304,7 +304,7 @@ public SearchResultResponse getFacetList(SearchRequest searchRequest) {
 	 *           Access groups of the user making this request.
 	 * @return
 	 */
-	public List<BriefObjectMetadataBean> getNeighboringItems(BriefObjectMetadataBean metadata, int windowSize,
+	public List<BriefObjectMetadataBean> getNeighboringItems(BriefObjectMetadata metadata, int windowSize,
```
harringj: just curious why you made this change. do you envision scenarios where a different type of BriefObjectMetadata might be used?

access-common/src/test/java/edu/unc/lib/dl/ui/service/NeighborQueryTest.java
bbpennel: Oh, they were copy and pasted from a test utility in the fcrepo4 branch, i should change them to private here

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
bbpennel: Generally I'd prefer to pass in the interface when possible. There is actually another implementation of BriefObjectMetadata. I ended up reverting this though because I needed to add the record to the result list, which needed to be the implementation.

PR #725, opened 2018-04-12
Update derivative processors to save files to file system

lfarrell:


services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
```diff
@@ -80,69 +71,27 @@
...
-    public void extractFulltextRetryFailTest() throws Exception {
-
-        when(binary.getParent()).thenThrow(new RuntimeException());
-
-        try {
-            processor.process(exchange);
-        } finally {
-            verify(binary, times(maxRetries + 1)).getParent();
-        }
+        assertTrue(file.exists());
```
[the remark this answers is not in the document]
lfarrell: I don't think these get written to the tmp dir. I think they go directly to their final home.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
```diff
@@ -80,69 +74,28 @@
...
     public void init() throws Exception {
         initMocks(this);
-        processor = new FulltextProcessor(repoObjLoader, slug, fileName, maxRetries, retryDelay);
-        file = File.createTempFile(fileName, "txt");
-        file.deleteOnExit();
-        when(exchange.getIn()).thenReturn(message);
 
-        when(repoObjLoader.getBinaryObject(any(PID.class))).thenReturn(binary);
+        processor = new FulltextProcessor(tmpDir.newFolder().getAbsolutePath());
+        file = tmpDir.newFile(fileName + ".txt");
```
bbpennel: this makes for file `full_text.txt.txt`, so you can leave off the concatenation here.

services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
```diff
@@ -86,30 +71,16 @@ public void process(Exchange exchange) throws Exception {
             return;
         }
 
-        int retryAttempt = 0;
-
-        InputStream binaryStream = new ByteArrayInputStream(text.getBytes(StandardCharsets.UTF_8));
+        Path filepath = Paths.get(derivativeBasePath, binarySubPath + ".txt");
```
bbpennel: Generally I'm not too picky about variable names, but could you rename this to `derivativePath` or something similar? It will be a little bit easier to track the difference between it and binaryPath, plus `filepath` refers to the original binary path down in `extractText`

services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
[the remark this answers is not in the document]
bbpennel: This is still testing that the original binary exists (`full_text.txt.txt`), it should verify that the derivative exists (`derivative.txt`). It should exist in the path the processor was initialized with. So I think you just need to switch which file is being tested in these two lines. And you are right that it goes directly to its final home, but in this test its home is still in tmpDir

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessorTest.java
```diff
@@ -17,61 +17,57 @@
...
-    private long retryDelay = 10;
+    private final String derivativeSubPath = "derivative";
     private File file;
+    private File mvFile;
 
     private AddDerivativeProcessor processor;
 
     private String extensionlessPath;
 
     private String extensionlessName;
```
bbpennel: Looks like this variable is no longer being used, just initialized

services-camel/src/main/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessor.java
```diff
@@ -48,61 +45,41 @@
...
+        moveFile(binaryUri, binarySubPath, derivativePath);
     }
 
-    private void ingestFile(String binaryUri, String binaryMimeType, String derivativePath)
-            throws FileNotFoundException {
-        String filename = derivativePath.substring(derivativePath.lastIndexOf('/') + 1);
-        InputStream binaryStream = new FileInputStream(derivativePath + "." + fileExtension);
+    private void moveFile(String binaryUri, String binarySubPath, String derivativeTmpPath)
+            throws IOException {
+        Path derivative_path = Paths.get(derivativeBasePath,  binarySubPath + "." + fileExtension);
```
bbpennel: ruby invasion! Please use camel case

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessorTest.java
```diff
@@ -81,12 +77,13 @@
     @Before
     public void init() throws Exception {
         initMocks(this);
-        processor = new AddDerivativeProcessor(repoObjLoader, slug, fileExtension, mimetype, maxRetries, retryDelay);
-        file = File.createTempFile(fileName, ".PNG");
+
+        processor = new AddDerivativeProcessor(fileExtension, moveDir.getRoot().getAbsolutePath());
+
+        file = tmpDir.newFile(fileName + ".PNG");
         file.deleteOnExit();
```
bbpennel: Don't need to call deleteOnExit

services-camel/src/main/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessor.java
```diff
@@ -48,61 +45,41 @@
...
     @Override
     public void process(Exchange exchange) throws Exception {
         Message in = exchange.getIn();
         String binaryUri = (String) in.getHeader(FCREPO_URI);
-        int retryAttempt = 0;
+        String binarySubPath = (String) in.getHeader(CdrBinarySubPath);
 
         final ExecResult result = (ExecResult) in.getBody();
 
         String derivativePath = new BufferedReader(new InputStreamReader(result.getStdout()))
```
bbpennel: Could you make this `derivativeTmpPath`? I seem to be having a hard time following which paths are landing in which variables since everything is derivatives.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessorTest.java
```diff
@@ -109,41 +109,11 @@ public void init() throws Exception {
...
-
-        when(binary.getParent())
-                .thenThrow(new RuntimeException());
-
-        try {
-            processor.process(exchange);
-        } finally {
-            verify(binary, times(maxRetries + 1)).getParent();
-        }
+        assertTrue(mvFile.exists());
```
bbpennel: After stepping through the test, it looks like this is testing that `<mvDir>/small_thumb.PNG` exists.
The file that should be created by the processor is:
`<mvDir>/derivative.PNG`
So it should be verifying that path exists instead

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessorTest.java
```diff
@@ -109,41 +109,11 @@ public void init() throws Exception {
 
     @Test
     public void createEnhancementTest() throws Exception {
-
-        when(repoObjLoader.getBinaryObject(any(PID.class))).thenReturn(binary);
-        when(binary.getParent()).thenReturn(parent);
-        when(message.getBody()).thenReturn(result);
+        mvFile = moveDir.newFile(fileName + ".PNG");
```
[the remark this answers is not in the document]
bbpennel: this creates the file, which is why it exists in the assertion at the end of the test :) I'd suggest just making a `File` object in the root of the moveDir rather than calling newFile.

services-camel/src/main/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessor.java
```diff
@@ -48,61 +45,41 @@
...
+        File derivative = derivative_path.toFile();
+        File parentDir = derivative.getParentFile();
 
-        BinaryObject binary = repoObjLoader.getBinaryObject(PIDs.get(binaryUri));
-        FileObject parent = (FileObject) binary.getParent();
-        parent.addDerivative(slug, binaryStream, filename, binaryMimeType, PcdmUse.ThumbnailImage);
+        if (parentDir != null) {
+            parentDir.mkdirs();
+        }
+        derivative.createNewFile();
```
bbpennel: Looks like this line isn't needed, the `move` command creates the destination file if it doesn't exist

services-camel/src/test/java/edu/unc/lib/dl/services/camel/BinaryMetadataProcessorTest.java
```diff
@@ -141,6 +147,7 @@ public void noLocalBinaryTest() throws Exception {
         verify(message).setHeader(CdrBinaryChecksum, checksum);
         verify(message).setHeader(CdrBinaryMimeType, mimetype);
         verify(message, never()).setHeader(eq(CdrBinaryPath), anyString());
+        verify(message, never()).setHeader(eq(CdrBinaryId), anyString());
```
bbpennel: If this class can never set CdrBinaryId, then you don't really need to test for it

services-camel/src/main/java/edu/unc/lib/dl/services/camel/util/CdrFcrepoHeaders.java
```diff
@@ -27,6 +27,8 @@
 
     public static final String CdrBinaryPath = "CdrBinaryPath";
 
+    public static final String CdrBinaryId = "CdrBinaryId";
```
bbpennel: Don't think this is being used anywhere anymore, other than being initialized in tests from when it was previously used. So I think you can go ahead and remove it

services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
```diff
@@ -77,7 +63,12 @@ public void process(Exchange exchange) throws Exception {
 
         String binaryUri = (String) in.getHeader(FCREPO_URI);
         String binaryPath = (String) in.getHeader(CdrBinaryPath);
+        String fcrepoBinaryUri = (String) in.getHeader("CamelFcrepoUri");
```
bbpennel: `FCREPO_URI` is actually a constant for "CamelFcrepoUri", so you could remove this and use `binaryUri` from line 64 instead

services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
```diff
@@ -16,133 +16,79 @@
...
-        when(exchange.getIn()).thenReturn(message);
 
-        when(repoObjLoader.getBinaryObject(any(PID.class))).thenReturn(binary);
+        derivPath = tmpDir.newFolder().getAbsolutePath();
+        processor = new FulltextProcessor(derivPath);
+        originalFile = tmpDir.newFile(originalFileName);
 
+        when(exchange.getIn()).thenReturn(message);
         when(message.getHeader(eq(FCREPO_URI))).thenReturn(BINARY_URI);
+        when(message.getHeader(eq("CamelFcrepoUri"))).thenReturn(RESC_ID);
```
bbpennel: This is overriding the header from the previous line, so you only need one of these. I'd suggest updating line 73 to use FCREPO_URI and removing line 72. I don't think `BINARY_URI` will be needed anymore after you remove the duplicate line.

services-camel/src/main/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessor.java
```diff
@@ -48,61 +49,42 @@
...
-        this.mimetype = mimetype;
+        this.derivativeBasePath = derivativeBasePath;
     }
 
     @Override
     public void process(Exchange exchange) throws Exception {
         Message in = exchange.getIn();
         String binaryUri = (String) in.getHeader(FCREPO_URI);
-        int retryAttempt = 0;
+        String binaryId = PIDs.get((String) in.getHeader(CdrBinaryUri)).getId();
```
bbpennel: Hmm, CdrBinaryUri is the header that stores the local file path to the original file stored by fedora isn't it? So the file name is actually the checksum of the file rather than the identifier. I think you want to be using binaryUri from line 63 for getting the binaryId instead of CdrBinaryUri

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessorTest.java
```diff
@@ -81,69 +73,43 @@
...
+
+        processor = new AddDerivativeProcessor(fileExtension, finalDir.getAbsolutePath());
+
+        pathId = PIDs.get(RESC_ID).getId();
+
+        file = tmpDir.newFile(pathId);
+
+        when(exchange.getIn()).thenReturn(message);
 
         when(message.getHeader(eq(FCREPO_URI)))
```
bbpennel: This should be RESC_ID instead of what's happening on line 92

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPaths.java
```diff
@@ -126,6 +126,26 @@ public static String getVocabulariesBase() {
         return vocabulariesBase;
     }
 
+    /**
+     * Prepend id with defined levels of hashed containers based on the values.
+     * For example, 9bd8b60e-93a2-4b66-8f0a-b62338483b39 would become
+     *    9b/d8/b6/9bd8b60e-93a2-4b66-8f0a-b62338483b39
+     *
+     * @param id
```
harringj: would someone mind filling out the javadoc here to make it more useful?

services-camel/src/main/java/edu/unc/lib/dl/services/camel/BinaryMetadataProcessor.java
```diff
@@ -45,10 +47,7 @@
  *
  */
 public class BinaryMetadataProcessor implements Processor {
-
     private final int BINARY_PATH_DEPTH = 3;
-    private final int BINARY_PATH_LENGTH = 2;
-
     private String baseBinaryPath;
 
     protected BinaryMetadataProcessor(String baseBinaryPath) {
```
harringj: should this constructor really have protected access? other processors look like they have public constructors

services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
```diff
@@ -86,30 +76,16 @@ public void process(Exchange exchange) throws Exception {
...
-                log.info("Unable to add derivative for {} from {}. Retrying, attempt {}",
-                        binaryUri, binaryPath, retryAttempt);
-                TimeUnit.MILLISECONDS.sleep(retryDelay);
+        if (parentDir != null) {
+            parentDir.mkdirs();
+        }
 
-            }
+        try (PrintWriter fulltext = new PrintWriter(derivativePath.toString())) {
+            fulltext.println(text);
```
harringj: i'm concerned that if the text extraction above fails for any reason other than a tika exception we have a NPE on our hands here at line 88

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessorTest.java
harringj: looks like this was fixed, no?

services-camel/src/test/resources/config.properties
```diff
@@ -13,6 +13,7 @@ cdr.enhancement.thumbnail.mimetype=image/png
 cdr.enhancement.fulltext.fileName=full_text.txt
 cdr.enhancement.jp2.fileExtension=JP2
 cdr.enhancement.jp2.mimetype=image/jp2
+cdr.enhancement.path=/tmp/derivative
```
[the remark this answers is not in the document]
harringj: does this still need to be removed, @bbpennel ?

services-camel/src/main/java/edu/unc/lib/dl/services/camel/BinaryMetadataProcessor.java
bbpennel: fixed

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryPaths.java
bbpennel: fixed

services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
bbpennel: I think most likely issues in here will throw IOExceptions. I have updated FullTextProcessor and AddDerivativesProcessor to now throw IOExceptions if they are unable to create parent directories for the final derivative instead of throwing a file not found exception a few lines down.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/AddDerivativeProcessorTest.java
bbpennel: Yes, this one should be good now

PR #726, opened 2018-04-19
Fix same items always appearing for preceding neighbors

bbpennel:


access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -332,26 +334,28 @@ public SearchResultResponse getFacetList(SearchRequest searchRequest) {
...
-		// Sort neighbors using the title sort
-		addSort(solrQuery, "title", true);
-
 		// Query for a window of results to either side of the target
 		solrQuery.setRows(windowSize - 1);
 		solrQuery.setFields(new String[0]);
 		
 		// Clone base query for reuse in getting succeeding neighbors
-		SolrQuery precedingQuery = solrQuery;
 		SolrQuery succeedingQuery = solrQuery.getCopy();
```
harringj: why is getCopy() being used here? wouldn't it be possible to set succeedingQuery to solrQuery, just like for precedingQuery?

bbpennel: Needed two query objects since from this point on the parameters diverge between the preceding and succeeding queries, which is why succeeding uses `solrQuery.getCopy()`. I opted to not make a third query by copying solrQuery a second time and instead renamed the variable to make it clear that the query was specializing.

PR #729, opened 2018-04-20
VM properties and cleanup

bbpennel:
Adds missing properties, cleans up some unused bean creation, resolves missing dependency and expands error message.

### DIFF
--- a/etc/deposit.properties
+++ b/etc/deposit.properties
@@ -2,6 +2,7 @@
 project.version=4.0
 
 fcrepo.baseUrl=http://localhost:8080/fcrepo/rest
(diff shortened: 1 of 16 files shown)


etc/deposit.properties
```diff
@@ -1,6 +1,8 @@
 # Fedora Commons Repository (https, localhost, 443, fedora, fedoraAdmin, ??)
 project.version=4.0
 
+fcrepo.baseUrl=http://localhost:8080/fcrepo/rest
```
harringj: do we really need both of these properties? i'm concerned that !) they are nearly identical and therefore easily confused and 2) that they are both really URLs

PR #732, opened 2018-04-25
Bxc-2166/2161 - Navigation issues

bbpennel:
* Fixes permission issue that was preventing filtering to specific containers from working
* Solr queries depending on search requests to provide access groups consistently
* Using PID class in solr search requests for better control over formatting of ids.
* Removed duplicate fcrepo.baseUri to fix issue retrieving info from fedora in access interface

### DIFF
--- a/solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/SimpleIdRequest.java
+++ b/solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/SimpleIdRequest.java
@@ -18,54 +18,53 @@ package edu.unc.lib.dl.search.solr.model;
 import java.util.List;
 
 import edu.unc.lib.dl.acl.util.AccessGroupSet;
+import edu.unc.lib.dl.fcrepo4.PIDs;
+import edu.unc.lib.dl.fedora.PID;
 
 /**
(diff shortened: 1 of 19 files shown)


solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/SimpleIdRequest.java
```diff
@@ -18,56 +18,55 @@
...
-        this.accessGroups = accessGroups;
-        this.resultFields = resultFields;
+    public SimpleIdRequest(PID pid, AccessGroupSet accessGroups) {
+        this(pid, null, accessGroups);
     }
 
-    public SimpleIdRequest(String id, List<String> resultFields) {
-        this.id = id;
-        this.resultFields = resultFields;
+    protected SimpleIdRequest(List<String> resultFields, AccessGroupSet accessGroups) {
```
harringj: why is this protected?

bbpennel: It was only being used by a subclass. That's silly though, so I went ahead and deleted this constructor and used a different super constructor in the child class instead. Thanks for making me rethink that

PR #733, opened 2018-05-01
F4 BXC-2164 - Neighbor queries

bbpennel:
Ports changes to neighbor query from master (updated for fcrepo4 branch) and splits neighbor querying into its own service.


solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/NeighborQueryService.java
```diff
@@ -0,0 +1,198 @@
...
+ * @author bbpennel
+ *
+ */
+public class NeighborQueryService extends AbstractQueryService {
+
+    protected FacetFieldUtil facetFieldUtil;
+
+    /**
+     * Retrieves a list of the closest windowSize neighbors within the parent container of the specified object,
+     * using title sort order. The first windowSize / 2 - 1 neighbors are retrieved to each side
```
harringj: Am I correct in thinking you meant to say "...(windowSize -1) / 2 neighbors are retrieved..." Whenever windowSize is even, windowSize - 2 neighbors will be retrieved; otherwise, windowSize - 1 neighbors will be retrieved.

bbpennel: Hmm, yeah I think that's right, I'll fix it

PR #736, opened 2018-05-04
updates all cdr modules to use jdom 2.0.6

harringj:


pom.xml
```diff
@@ -538,8 +538,8 @@
             </dependency>
             <dependency>
                 <groupId>org.jdom</groupId>
-                <artifactId>jdom</artifactId>
-                <version>2.0.2</version>
+                <artifactId>jdom2</artifactId>
+                <version>2.0.6</version>
```
bbpennel: Could you switch to using a variable for the version? I'd generally like to move in that direction for our dependencies.
I didn't realize we were going to need to use a different artifact id, they really switched from jdom to jdom2 between 2.0.2 and 2.0.6?

harringj: sure, i can add that variable. and yes, they did really change the artifactId, which i only realized after trying to build:

https://mvnrepository.com/artifact/org.jdom/jdom2/2.0.6

PR #738, opened 2018-05-08
Bxc 2147 - Download of original file

bbpennel:
* Fixes URLs for download links
* Updates controllers and service involved in stream binaries from fedora
* Adds FileObject.getBinary(name) for retrieval of a single binary by name
* Adds integration tests for binary retrieval controllers
* Updates AnalyticsUtil to be more self contained to make for easier mocking.
* Switched to getOriginalFileUrl helper method for views versus passing in datastream name.

### DIFF
--- a/access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
+++ b/access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
@@ -15,235 +15,98 @@
  */
 package edu.unc.lib.dl.ui.service;
 
+import static edu.unc.lib.dl.acl.fcrepo4.DatastreamPermissionUtil.getPermissionForDatastream;
+import static edu.unc.lib.dl.fcrepo4.RepositoryPathConstants.ORIGINAL_FILE;
+import static org.apache.http.HttpHeaders.CONTENT_LENGTH;
+import static org.apache.http.HttpHeaders.CONTENT_TYPE;
+
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.List;
+import java.io.OutputStream;
 
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
 
+import edu.unc.lib.dl.acl.service.AccessControlService;
 import edu.unc.lib.dl.acl.util.AccessGroupSet;
-import edu.unc.lib.dl.acl.util.GroupsThreadStore;
-import edu.unc.lib.dl.fedora.AccessClient;
-import edu.unc.lib.dl.fedora.AuthorizationException;
-import edu.unc.lib.dl.fedora.FedoraException;
-import edu.unc.lib.dl.search.solr.model.BriefObjectMetadataBean;
-import edu.unc.lib.dl.search.solr.model.Datastream;
-import edu.unc.lib.dl.search.solr.model.SimpleIdRequest;
-import edu.unc.lib.dl.search.solr.util.SearchFieldKeys;
-import edu.unc.lib.dl.search.solr.util.SearchSettings;
-import edu.unc.lib.dl.ui.exception.InvalidRecordRequestException;
-import edu.unc.lib.dl.ui.exception.ResourceNotFoundException;
-import edu.unc.lib.dl.ui.util.AnalyticsTrackerUtil;
-import edu.unc.lib.dl.ui.util.AnalyticsTrackerUtil.AnalyticsUserData;
-import edu.unc.lib.dl.util.ContentModelHelper;
-import edu.unc.lib.dl.util.ContentModelHelper.DatastreamCategory;
+import edu.unc.lib.dl.fcrepo4.BinaryObject;
+import edu.unc.lib.dl.fcrepo4.FileObject;
+import edu.unc.lib.dl.fcrepo4.RepositoryObjectLoader;
+import edu.unc.lib.dl.fedora.PID;
 
 /**
- * Connects to and streams datastreams from Fedora.
+ * Streams binary content from repository objects.
  *
  * @author bbpennel
  */
 public class FedoraContentService {
     private static final Logger LOG = LoggerFactory.getLogger(FedoraContentService.class);
 
-    private AccessClient accessClient;
+    public static final String CONTENT_DISPOSITION = "Content-Disposition";
 
-    private String fedoraHost;
+    private AccessControlService accessControlService;
 
-    @Autowired
-    private SearchSettings searchSettings;
-    @Autowired
-    protected SolrQueryLayerService queryLayer;
+    private RepositoryObjectLoader repositoryObjectLoader;
 
-    @Autowired(required = false)
-    protected AnalyticsTrackerUtil analyticsTracker;
-
-    private final int numberOfRetries = 1;
-
-    private static List<String> resultFields = Arrays.asList(SearchFieldKeys.ID.name(),
-            SearchFieldKeys.DATASTREAM.name(), SearchFieldKeys.LABEL.name(), SearchFieldKeys.RESOURCE_TYPE.name(),
-            SearchFieldKeys.ROLE_GROUP.name(), SearchFieldKeys.PARENT_COLLECTION.name(),
-            SearchFieldKeys.ANCESTOR_PATH.name(), SearchFieldKeys.TITLE.name());
-
-    public void streamData(String pid, String datastream, boolean download, AnalyticsUserData userData,
-            HttpServletResponse response) {
-        AccessGroupSet accessGroups = GroupsThreadStore.getGroups();
-
+    /**
+     * Set content headers and stream the binary content of the specified
+     * datastream from the object identified by pid.
+     *
+     * @param pid pid of object containing datastream
+     * @param dsName name of datastream being requested. If null, then original
+     *            file assume.
+     * @param principals principals of requesting client
+     * @param asAttachment if true, then content-disposition header will specify
+     *            as "attachment" instead of "inline"
+     * @param response response content and headers will be added to.
+     * @throws IOException
+     */
+    public void streamData(PID pid, String dsName, AccessGroupSet principals, boolean asAttachment,
+            HttpServletResponse response) throws IOException {
         // Default datastream is DATA_FILE
-        if (datastream == null) {
-            datastream = ContentModelHelper.Datastream.DATA_FILE.toString();
+        String datastream = dsName == null ? ORIGINAL_FILE : dsName;
+
+        accessControlService.assertHasAccess("Insufficient permissions to access " + datastream + " for object " + pid,
+                pid, principals, getPermissionForDatastream(datastream));
+
+        LOG.debug("Streaming datastream {} from object {}", datastream, pid);
+
+        FileObject fileObj = repositoryObjectLoader.getFileObject(pid);
+        BinaryObject binObj;
+        if (ORIGINAL_FILE.equals(datastream)) {
+            binObj = fileObj.getOriginalFile();
+        } else {
+            binObj = fileObj.getBinaryObject(datastream);
         }
 
-        // Use solr to check if the user is allowed to view this item.
-        SimpleIdRequest idRequest = new SimpleIdRequest(pid, resultFields, accessGroups);
-
-        BriefObjectMetadataBean briefObject = queryLayer.getObjectById(idRequest);
-        // If the record isn't accessible then invalid record exception.
-        if (briefObject == null) {
-            throw new InvalidRecordRequestException();
+        // Set binary detail response headers
+        response.setHeader(CONTENT_LENGTH, Long.toString(binObj.getFilesize()));
+        response.setHeader(CONTENT_TYPE, binObj.getMimetype());
+        String filename = binObj.getFilename() == null ? pid.getId() : binObj.getFilename();
+        if (asAttachment) {
+            response.setHeader(CONTENT_DISPOSITION, "attachment; filename=\"" + filename + "\"");
+        } else {
+            response.setHeader(CONTENT_DISPOSITION, "inline; filename=\"" + filename + "\"");
         }
-        // Block access to thumbnails for non-containers,
-        // TODO Perform ACL check before allowing streaming
-//        if (AccessUtil.hasListAccessOnly(accessGroups, briefObject)
-//                && (searchSettings.resourceTypeFile.equals(briefObject.getResourceType())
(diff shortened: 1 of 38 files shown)


access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
```diff
@@ -15,235 +15,98 @@
...
+    private AccessControlService accessControlService;
+
+    private RepositoryObjectLoader repositoryObjectLoader;
+
+    /**
+     * Set content headers and stream the binary content of the specified
+     * datastream from the object identified by pid.
+     *
+     * @param pid pid of object containing datastream
+     * @param dsName name of datastream being requested. If null, then original
```
harringj: Wording is a bit awkward here: "If null, then original file assume."

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
```diff
@@ -15,235 +15,98 @@
...
+     *            file assume.
+     * @param principals principals of requesting client
+     * @param asAttachment if true, then content-disposition header will specify
+     *            as "attachment" instead of "inline"
+     * @param response response content and headers will be added to.
+     * @throws IOException
```
harringj: perhaps consider adding a description about how/when this exception gets thrown?

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
```diff
@@ -15,235 +15,98 @@
...
+     */
+    public void streamData(PID pid, String dsName, AccessGroupSet principals, boolean asAttachment,
+            HttpServletResponse response) throws IOException {
+        // Default datastream is DATA_FILE
+        String datastream = dsName == null ? ORIGINAL_FILE : dsName;
```
harringj: i know this pattern is popular in the Fedora codebase, and i know it's not that hard to understand, but i tend to dislike the ternary operator + assignment in a single statement. it's less clear than, say

`String datastream;
if (dsName == null) {
    datastream = ORIGINAL_FILE;
} else {
    datastream = dsName;`
}
or even
` String datastream = dsName;
if (datastream == null) {
    datastream = ORIGINAL_FILE;
}`

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
```diff
@@ -15,235 +15,98 @@
...
-
-    public void setQueryLayer(SolrQueryLayerService queryLayer) {
-        this.queryLayer = queryLayer;
-    }
 
-    public void setAnalyticsTracker(AnalyticsTrackerUtil analyticsTracker) {
-        this.analyticsTracker = analyticsTracker;
+        // Stream binary content to http response
+        OutputStream outStream = response.getOutputStream();
+        IOUtils.copy(binObj.getBinaryStream(), outStream, 4096);
```
harringj: why 4096 here? can this be stored in a class var?

access-common/src/main/java/edu/unc/lib/dl/ui/util/AnalyticsTrackerUtil.java
```diff
@@ -82,6 +90,28 @@ public void setGaTrackingID(String trackingID) {
         this.gaTrackingID = trackingID;
     }
 
+    /**
+     * Track an event with the specified action for object pid for the active user on the request.
+     *
+     * @param request request
+     * @param action action for the event
+     * @param pid pid identifying the object involved in the event
+     * @param principals
```
harringj: missing javadoc comment for principals

access-common/src/main/java/edu/unc/lib/dl/ui/util/AnalyticsTrackerUtil.java
```diff
@@ -82,6 +90,28 @@ public void setGaTrackingID(String trackingID) {
...
+     */
+    public void trackEvent(HttpServletRequest request, String action, PID pid, AccessGroupSet principals) {
+        try {
+            AnalyticsUserData userData = new AnalyticsUserData(request);
+
+            BriefObjectMetadata briefObject = solrSearchService.getObjectById(new SimpleIdRequest(pid, principals));
+            trackEvent(userData, briefObject.getParentCollection() == null ? "(no collection)"
```
harringj: i find the method call in lines 106-108 to be a little too compressed for easy readability. would you be willing to break this up into a few simpler steps?

access-common/src/main/java/edu/unc/lib/dl/ui/util/AnalyticsTrackerUtil.java
```diff
@@ -94,6 +124,13 @@ public void trackEvent(AnalyticsUserData userData, String category, String actio
...
+    /**
+     * @param solrSearchService the solrSearchService to set
+     */
+    public void setSolrSearchService(SolrSearchService solrSearchService) {
+        this.solrSearchService = solrSearchService;
+    }
+
     public static class AnalyticsUserData {
         public String uip;
         public String cid;
```
harringj: what does "cid" stand for? it might be helpful to use more self-documenting var names here, or to put a comment above each of these String vars to assist in comprehension of what this code is doing

access-common/src/main/java/edu/unc/lib/dl/ui/util/DatastreamUtil.java
```diff
@@ -67,26 +55,35 @@ public String getDatastreamUrl(BriefObjectMetadata metadata, String datastreamNa
...
         url.append("content/");
-        if (preferredDS.getOwner() == null) {
+        if (isBlank(preferredDS.getOwner())) {
             url.append(metadata.getId());
         } else {
             url.append(preferredDS.getOwner());
         }
-        if (!ContentModelHelper.Datastream.DATA_FILE.getName().equals(datastreamName)) {
+        if (!ORIGINAL_FILE.equals(datastreamName)) {
             url.append("/").append(preferredDS.getName());
```
harringj: why is preferredDS.getName() being appended to the url rather than datastreamName?

access/pom.xml
```diff
@@ -252,5 +285,87 @@
             <artifactId>solr-search</artifactId>
             <version>${cdr.version}</version>
         </dependency>
+        <!-- integration testing -->
+        <dependency>
+            <groupId>org.fcrepo</groupId>
+            <artifactId>fcrepo-http-commons</artifactId>
+            <version>${fcrepo4.version}</version>
```
harringj: wonder if we should think about changing the property namespace to be simply "fcrepo", since fcrepo5 is coming before too long? not that this change needs to happen on this PR

access/src/main/java/edu/unc/lib/dl/ui/controller/FedoraContentController.java
```diff
@@ -37,64 +53,94 @@
...
     public void getDefaultDatastream(@PathVariable("pid") String pid,
             @RequestParam(value = "dl", defaultValue = "false") boolean download,
             HttpServletRequest request, HttpServletResponse response) {
-//        fedoraContentService.streamData(pid, ContentModelHelper.Datastream.DATA_FILE.getName(), download,
-//                new AnalyticsUserData(request), response);
+        streamData(pid, ORIGINAL_FILE, download, request, response);
     }
 
     @RequestMapping("/content/{pid}/{datastream}")
     public void getDatastream(@PathVariable("pid") String pid, @PathVariable("datastream") String datastream,
```
harringj: why are both getIndexableContent() and getDatastream() needed? they appear to behave identically other than their request mappings. ditto, for getIndexableContentByParameters and getDefaultDatastream(). maybe some javadoc would help clarify the purpose of the various methods :-)

access/src/test/java/edu/unc/lib/dl/ui/controller/FedoraContentControllerIT.java
```diff
@@ -0,0 +1,261 @@
...
+    public void testGetMultipleDatastreams() throws Exception {
+        testGetMultipleDatastreams("/content/");
+    }
+
+    @Test
+    public void testGetMultipleIndexableDatastreams() throws Exception {
+        testGetMultipleDatastreams("/indexablecontent/");
+    }
+
+    public void testGetMultipleDatastreams(String requestPath) throws Exception {
```
harringj: method should be private

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/FileObject.java
```diff
@@ -181,4 +182,30 @@ public BinaryObject addBinary(String slug, InputStream contentStream, String fil
 
         return binaries;
     }
+
+    /**
+     * Retrieve binary object by name from the set of binaries contained by this
+     * FileObject.
+     *
+     * @param name
```
harringj: would you mind completing this javadoc?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
```diff
@@ -15,57 +15,88 @@
...
+import edu.unc.lib.dl.fcrepo4.PIDs;
+import edu.unc.lib.dl.fedora.PID;
 import edu.unc.lib.dl.ui.service.FedoraContentService;
-import edu.unc.lib.dl.ui.util.AnalyticsTrackerUtil.AnalyticsUserData;
-import edu.unc.lib.dl.util.ContentModelHelper;
+import edu.unc.lib.dl.ui.util.AnalyticsTrackerUtil;
 
 /**
- * 
+ *
```
harringj: javadoc class comment?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
```diff
@@ -15,57 +15,88 @@
...
+        analyticsTracker.trackEvent(request, "download", pid, principals);
     }
 
     @RequestMapping("/thumb/{pid}")
     public void getThumbnailSmall(@PathVariable("pid") String pid,
             @RequestParam(value = "size", defaultValue = "small") String size, HttpServletRequest request,
             HttpServletResponse response) {
-        fedoraContentService.streamData(pid, ContentModelHelper.Datastream.THUMB_SMALL.getName(),
-                false, null, response);
+        // TODO implement retrieval of derivatives
```
harringj: which JIRA ticket tracks this TODO?

solr-search/src/test/java/edu/unc/lib/dl/search/solr/model/DatastreamTest.java
```diff
@@ -27,49 +27,70 @@
  *
  */
 public class DatastreamTest {
+    private static final String CHECKSUM = "urn:sha1:2c9e902133d6809b445942dca5a93a9b3e766c15";
 
     @Test
-    public void datastreamParsing() {
-        Datastream ds = new Datastream("DATA_FILE|image/jpeg|jpg|30459|dc93eff50ca7dbd688971716e55e0084|");
+    public void testToString() {
```
harringj: method should be private

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
harringj: is it https://jira.lib.unc.edu/browse/BXC-2154?

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
bbpennel: I feel like this is the situation where ternary operator are appropriate, since it is a simple if/else assignment. That is the impression I am getting from stackoverflow responses as well https://softwareengineering.stackexchange.com/questions/28314/ternary-operator-considered-harmful/28394

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
bbpennel: I will add this but we have often not included descriptions for @throws, are you interested in changing that going forward? We do most likely need a more intentional approach to what we throw and don't throw. My tendency is to add descriptions if it seems like the throw is part of the api that the code calling it will need to know to expect and recover from. Perhaps that means that in other cases we should be recasting to a runtime exception instead.

access-common/src/main/java/edu/unc/lib/dl/ui/util/DatastreamUtil.java
bbpennel: They are the same thing in this case, but I will change the conditional to check `preferredDS.getName()` for consistency

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
harringj: hmm, the stackoverflow threads discussing this topic seem to have been closed due to being not constructive :-) i'm ok with it, but would really prefer parentheses at the very least for improved readability, if you wouldn't be totally opposed to that.

access-common/src/main/java/edu/unc/lib/dl/ui/service/FedoraContentService.java
harringj: i do think descriptions help to understand what to expect from a method, but also agree with you it's most important in the case of exceptions when the throw is part of the api the calling code will need to know to expect and recover from. i would say if the calling code doesn't need to know about and/or be able to recover from a certain exception, maybe it shouldn't be in the javadoc. and 👍  to the idea of a more intentional approach!

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
bbpennel: I can see jira at the moment, but the ticket is in the current sprint

access/src/main/java/edu/unc/lib/dl/ui/controller/FedoraContentController.java
bbpennel: /indexablecontent/ is just for SEO so that we have a different path to identify files that we want search engines to download and index. For instance, we don't really want google to download our 26gb zip files, but we do want them to download and expose pdfs.

I'm actually not sure why I have them as separate controller methods since you can assign multiple paths to the same method, so I have refactored the class to only have a single set of methods.

access-common/src/main/java/edu/unc/lib/dl/ui/util/AnalyticsTrackerUtil.java
bbpennel: I've added comments, including a link to the google analytics documentation where these parameters are coming from

PR #739, opened 2018-05-16
Bxc 1924

harringj:


persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsService.java
```diff
@@ -79,7 +81,7 @@ public String moveObjects(AgentPrincipals agent, PID destinationPid, List<PID> p
         job.setOperationMetrics(operationMetrics);
 
         if (asynchronous) {
-            log.info("User {} is queuing move operation {} of {} objects to destination {}",
+            log.info("User {} is queueing move operation {} of {} objects to destination {}",
```
bbpennel: I'm not claiming to be a spectacular speller, but just saying: https://english.stackexchange.com/questions/60852/queueing-or-queuing :)

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,154 @@
...
+            FileObject file = (FileObject) rootOfTree;
+            List<BinaryObject> binaries = file.getBinaryObjects();
+            for (BinaryObject binary : binaries) {
+                destroyTree(binary);
+            }
+        }
+        // destroy root of tree
+        Model stoneModel = rootOfTree.getModel();
+        stoneModel = convertModelToTombstone(rootOfTree);
+        repoObjFactory.createObject(rootOfTree.getUri(), stoneModel);
```
bbpennel: Its kind of weirding me out that the method is named createObject when its being used to PUT to an existing resource, but it looks like it should work

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,154 @@
...
+
+        //add premis event to tombstone
+        rootOfTree.getPremisLog().buildEvent(Premis.Deletion)
+        .addImplementorAgent(agent.getUsernameUri())
+        .addEventDetail("Item deleted from repository and replaced by tombstone")
+        .write();
+
+        // persist model
+        repoObjFactory.createObject(rootOfTree.getUri(), stoneModel);
```
bbpennel: Is a second call to persist the model needed here?

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private Model convertModelToTombstone(RepositoryObject destroyedObj) {
+        Model model = destroyedObj.getModel();
+        Resource resc = destroyedObj.getResource();
+
+        if (destroyedObj instanceof BinaryObject) {
+            // whitelisted properties for tombstone model of binaries
+            Selector selectEbucoreFilename = new SimpleSelector(resc, Ebucore.filename, (Object) null);
```
bbpennel: I think this approach will work. Looking at the jena documents, a slight variation would be to extend SimpleSelector with a custom .selects method which tests against all of the requested predicates. Its probably slightly more performent, but may improve readability. This is based off the second to last example in the "Querying a Model" section here https://jena.apache.org/tutorials/rdf_api.html#ch-Querying%20a%20Model
```
// Or MultiPropertySelector
class BinaryPropertySelector extends SimpleSelector {
          List<Property> permittedPredicates = // List of predicates

          public BinaryPropertySelector(Resource subject) {
              // call super constructor
           }

          @Override
          public boolean selects(Statement s) {
              return (subject == null   || s.getSubject().equals(subject))
                  && (predicate == null || /* Compare against a list of predicates */)
                  && (object == null    || s.getObject().equals(object)) ;
          }
     }
}
```
And to use:
`StmtIterator iter = model.listStatements(new BinaryPropertySelector(resc));`

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,154 @@
...
+            while (iter.hasNext()) {
+                Statement s = iter.nextStatement();
+                if (!(selectEbucoreFilename.test(s) || selectDcFilename.test(s) || selectPremisOriginalName.test(s)
+                        || selectMimetype.test(s) || selectChecksum.test(s) || selectFilesize.test(s))) {
+                    model.remove(s);
+                }
+            }
+        } else {
+            // retain only the object-type of a non-binary object
+            Selector selectRdfType = new SimpleSelector(resc, RDF.type, (Object) null);
```
bbpennel: We also want to keep the dc:title and premis:hasOriginalFilename here too (imported folders can have an original filename)

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,154 @@
...
+                if (!selectRdfType.test(s)) {
+                    model.remove(s);
+                }
+            }
+        }
+        // determine path and store in tombstone model
+        String path = pathFactory.getPath(destroyedObj.getPid()).toNamePath();
+        model.add(resc, Cdr.historicalPath, path);
+        model.add(resc, RDF.type, Cdr.Tombstone);
+        return model;
```
bbpennel: My impression is that you could populate a new model from the filtered properties rather than editing the existing model and persist the changes back to fedora with no issues with the `repoObjFactory.createObject` approach you're using. I don't feel too strongly about it, seems a little cleaner to not edit the retrieved model though I guess?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Tombstone.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ */
+public class Tombstone extends RepositoryObject {
+
+    private Map<String,String> record;
+
+    public Tombstone(PID pid, RepositoryObjectDriver driver, RepositoryObjectFactory repoObjFactory,
+            Map<String, String> record) {
+        super(pid, driver, repoObjFactory);
+        this.record = record;
```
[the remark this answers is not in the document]
harringj: yeah, it is. you may recall i had at one time been thinking of working with a map, rather than with the Jena model.

i wonder, though, whether we even need this class now?

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
harringj: sure, i could do that. i wrestled with which way made more sense, and for some reason it made more logical sense to me to transform the existing model, since that's the approach we're taking with the object. but i also don't have strong feelings here

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
bbpennel: This might make it a bit easier to have a predefined list of properties as a constant

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,154 @@
...
+             tx.close();
+        }
+    }
+
+    private void destroyTree(RepositoryObject rootOfTree) throws FedoraException {
+        if (rootOfTree instanceof ContentContainerObject) {
+            ContentContainerObject container = (ContentContainerObject) rootOfTree;
+            List<ContentObject> members = container.getMembers();
+            for (ContentObject member : members) {
+                destroyTree(member);
```
bbpennel: I think after you destroy each member you will need to destroy the proxy from the parent to it as well

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Tombstone.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public RepositoryObject validateType() throws FedoraException {
+        if (!isType(Fcrepo4Repository.Tombstone.toString())) {
+            throw new ObjectTypeMismatchException("Object " + pid + " is not a tombstone");
+        }
+        return this;
+    }
+
+    @Override
+    public RepositoryObject getParent() {
```
bbpennel: Since a Tombstone is no longer in the hierarchy (result of removing the proxy), this method should probably either return a null or throw an exception

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsService.java
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param agent security principals of the agent making request
+     * @param ids list of objects to destroy
+     */
+    public void destroyObjects(AgentPrincipals agent, List<String> ids) {
+        List<PID> objsToDestroy = new ArrayList<>();
+
+        for (String id : ids) {
+            PID pid = PIDs.get(id);
+            aclService.assertHasAccess("User does not have permission to destroy this object", pid,
```
bbpennel: Annoyingly, destroy and destroyUnit are two separate permissions right now since they are assigned to different roles. So you will need to figure out if the object is a unit to know which permission to check for.

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
harringj: but i actually do that up in the run() method before i even call destroyTree(). do you think that needs to change?

fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/Tombstone.java
harringj: ok, i'll return null for now

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsService.java
harringj: ok, good to know. how would you like us to handle any AccessRestrictionExceptions that may be thrown? is it enough to eat them, or would you like to log them? obviously we don't want to blow up the for-loop, so i went ahead and wrapped those checks in a try-catch

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
bbpennel: that will only remove the first proxy though, from the start of the tree to its parent, right? There will be proxies between all the children and their children if you are destroying a folder that contains folders.

But it might not actually be necessary, its possible fedora will delete the proxies since they are ldp:contained by the thing being deleted. If you can verify that the proxies don't outlive their parents (and leave tombstones all over the place) then what you have here should be good.

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsService.java
bbpennel: actually i'm not realizing this loop is just over the top level objects being deleted, not for every child object that will be deleted. 

The access restriction exceptions need to be reported in some way, preferably to the client that made the request. I don't think we need to log them. My first inclination is to block the whole destroy operation if the user doesn't have permission to destroy any of the objects being requested (since it was submitted as a single operation), but I'm open to continuing with the other objects if you think that makes more sense. I'm not sure how we are handling this situation in other operations right now.

harringj: I think I'll just leave the calls to aclService.assertHasAccess() as they are, and remove the try-catch stuff. That way the action will get canceled right away and whatever client calls destroyObjects() on the service can handle the AccessRestrictionException. Sound reasonable?

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
```diff
@@ -0,0 +1,268 @@
...
+        assertTrue(folderObj.getModel().contains(folderObj.getResource(), RDF.type, Cdr.Tombstone));
+        assertTrue(folderObj2.getModel().contains(folderObj2.getResource(), RDF.type, Cdr.Tombstone));
+
+        PremisEventObject event = repoObjLoader.getPremisEventObject(folderObj.getPremisLog().listEvents().get(0));
+        assertTrue(event.getResource().hasProperty(Premis.hasEventType, Premis.Deletion));
+        assertTrue(event.getResource().hasProperty(Premis.hasEventDetail,
+                "Item deleted from repository and replaced by tombstone"));
+
+        PremisEventObject event2 = repoObjLoader.getPremisEventObject(folderObj2.getPremisLog().listEvents().get(0));
+        assertTrue(event.getResource().hasProperty(Premis.hasEventType, Premis.Deletion));
```
bbpennel: Lines 187 and 188 should be testing `event2` instead of `event`

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
```diff
@@ -0,0 +1,268 @@
...
+        assertTrue(event.getResource().hasProperty(Premis.hasEventDetail,
+                "Item deleted from repository and replaced by tombstone"));
+    }
+
+    @Test
+    public void destroyFolderTest() {
+        PID folderObjPid = objsToDestroy.get(0);
+        initializeJob(Arrays.asList(folderObjPid));
+        FolderObject folderObj = repoObjLoader.getFolderObject(folderObjPid);
+        WorkObject workObj = repoObjLoader.getWorkObject(folderObj.getMembers().get(0).getPid());
```
bbpennel: Why not cast the member instead of re-retrieving it?
`WorkObject workObj = (WorkObject) folderObj.getMembers().get(0);`

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
```diff
@@ -0,0 +1,268 @@
...
+    @Test
+    public void destroySingleObjectTest() {
+        PID fileObjPid = objsToDestroy.get(2);
+        initializeJob(Arrays.asList(fileObjPid));
+
+        job.run();
+
+        verify(spyProxyService).destroyProxy(fileObjPid);
+        FileObject fileObj = repoObjLoader.getFileObject(fileObjPid);
+        assertTrue(fileObj.getModel().contains(fileObj.getResource(), RDF.type, Cdr.Tombstone));
```
bbpennel: It would be good to verify that the tombstone retains the original rdf type as well in at least one of these tests.

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,211 @@
...
+                addBinaryMetadataToParent(rootOfTree, origFile);
+            }
+        }
+        Model rootModel = rootOfTree.getModel();
+        boolean hasLdpContains = rootModel.contains(rootOfTree.getResource(), Ldp.contains);
+        if (hasLdpContains) {
+            deleteNonContentObjects(rootModel);
+        }
+        // destroy root of sub-tree
+        Model stoneModel = rootOfTree.getModel();
```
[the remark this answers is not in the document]
harringj: yeah, good call. i think this was left over from an earlier impl.

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+        if (rootOfTree instanceof FileObject) {
+            FileObject file = (FileObject) rootOfTree;
+            BinaryObject origFile = file.getOriginalFile();
+            if (origFile != null) {
+                addBinaryMetadataToParent(rootOfTree, origFile);
+            }
+        }
+        Model rootModel = rootOfTree.getModel();
```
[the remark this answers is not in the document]
harringj: no, this is good stuff to be thinking about, and i'm glad you put your thoughts in writing. i hadn't tied the model and resource fetching to the etag issue, so wasn't operating under the assumption that we were paying such a steep penalty and making a lot of requests to fedora. i'll see what i can do here to minimize those calls. we can formulate a more comprehensive strategy once we've measured performance.

persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -0,0 +1,211 @@
...
+            throws IOException, FcrepoOperationFailedException {
+        Model oldModel = destroyedObj.getModel();
+        Resource resc = destroyedObj.getResource();
+
+        Model stoneModel = ModelFactory.createDefaultModel();
+        stoneModel.add(oldModel.listStatements(new TombstonePropertySelector(resc)));
+
+        // determine path and store in tombstone model
+        String path = pathFactory.getPath(destroyedObj.getPid()).toNamePath();
+        stoneModel.add(resc, Cdr.historicalPath, path);
```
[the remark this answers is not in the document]
harringj: good catch!

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
harringj: oops, fixed!

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
harringj: good idea. done!

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
harringj: okie dokie

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
```diff
@@ -143,14 +145,17 @@ public void destroyObjectsInSameTreeTest() {
...
-        FolderObject folderObj = repoObjLoader.getFolderObject(folderObjPid);
-        assertTrue(fileObj.getModel().contains(fileObj.getResource(), RDF.type, Cdr.Tombstone));
-        assertTrue(workObj.getModel().contains(workObj.getResource(), RDF.type, Cdr.Tombstone));
-        assertTrue(folderObj.getModel().contains(folderObj.getResource(), RDF.type, Cdr.Tombstone));
-
-        PremisEventObject event = repoObjLoader.getPremisEventObject(folderObj.getPremisLog().listEvents().get(0));
+        Tombstone stoneFile = repoObjLoader.getTombstone(fileObjPid);
+        Tombstone stoneWork = repoObjLoader.getTombstone(workObjPid);
+        Tombstone stoneFolder = repoObjLoader.getTombstone(folderObjPid);
+        assertTrue(stoneFile.getModel().contains(stoneFile.getResource(), RDF.type, Cdr.Tombstone));
```
bbpennel: I'm okay if you want to keep this as is, but it may be slightly more readable to adjust these assertions to 
`assertTrue(stoneFile.getResource().hasProperty(RDF.type, Cdr.Tombstone));`

persistence/src/main/java/edu/unc/lib/dl/util/TombstonePropertySelector.java
```diff
@@ -37,9 +37,9 @@
 public class TombstonePropertySelector extends SimpleSelector {
 
     List<Property> permittedPredicates = Arrays.asList(
```
bbpennel: Don't think this will do quite what we want, since the Binary will have the server managed triples. There will probably need to be a step that remaps the original properties to the cdr namespace versions

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
harringj: sure, fixed!

PR #740, opened 2018-05-22
Fcrepo 2154 - Thumbnails and Derivatives

bbpennel:
* Enables display of thumbnails in the access ui
* Adds indexing of derivatives held outside of fedora
* Adds enum describing known datastreams and properties to allow for easier sharing of info about file type, storage policy, and permissions of datastreams.
* Thumbnails now coming from services application instead of access app in an effort to consolidate the API
* Moves some enhancement properties over into enum
* switch from CAPS file extensions for derivatives to lowercase.

### DIFF
--- a/access-common/src/main/java/edu/unc/lib/dl/ui/service/DerivativeContentService.java
+++ b/access-common/src/main/java/edu/unc/lib/dl/ui/service/DerivativeContentService.java
@@ -18,6 +18,7 @@ package edu.unc.lib.dl.ui.service;
 import static edu.unc.lib.dl.acl.fcrepo4.DatastreamPermissionUtil.getPermissionForDatastream;
 import static edu.unc.lib.dl.model.DatastreamType.getByIdentifier;
 import static edu.unc.lib.dl.ui.service.FedoraContentService.CONTENT_DISPOSITION;
+import static edu.unc.lib.dl.util.DerivativeService.listDerivativeTypes;
 import static org.apache.http.HttpHeaders.CONTENT_LENGTH;
 import static org.apache.http.HttpHeaders.CONTENT_TYPE;
 
@@ -56,22 +57,25 @@ public class DerivativeContentService {
     }
 
     /**
-     * Set content headers and stream the content of the specified derivative from the object identified by pid.
+     * Set content headers and stream the content of the specified derivative
+     * from the object identified by pid.
      *
      * @param pid pid of object containing the derivative
-     * @param dsName name of derivative being requested.
+     * @param dsName name of derivative being requested. Must be a derivative
+     *            type, otherwise an IllegalArgumentException will be thrown.
      * @param principals principals of requesting client
      * @param asAttachment if true, then content-disposition header will specify
      *            as "attachment" instead of "inline"
      * @param response response content and headers will be added to.
      * @throws IOException if unable to stream content to the response.
-     * @throws ResourceNotFoundException if an invalid derivative type is requested.
+     * @throws ResourceNotFoundException if an invalid derivative type is
+     *             requested.
      */
     public void streamData(PID pid, String dsName, AccessGroupSet principals, boolean asAttachment,
             HttpServletResponse response) throws IOException, ResourceNotFoundException {
 
         DatastreamType derivType = getByIdentifier(dsName);
-        if (derivType == null || !derivativeService.listDerivativeTypes().contains(derivType)) {
+        if (derivType == null || !listDerivativeTypes().contains(derivType)) {
             throw new IllegalArgumentException(dsName + " is not a valid derivative type.");
         }
 
--- a/access-common/src/main/java/edu/unc/lib/dl/ui/service/DerivativeContentService.java
+++ b/access-common/src/main/java/edu/unc/lib/dl/ui/service/DerivativeContentService.java
@@ -1,0 +1,99 @@
+package edu.unc.lib.dl.ui.service;
+
+import static edu.unc.lib.dl.acl.fcrepo4.DatastreamPermissionUtil.getPermissionForDatastream;
+import static edu.unc.lib.dl.model.DatastreamType.getByIdentifier;
+import static edu.unc.lib.dl.ui.service.FedoraContentService.CONTENT_DISPOSITION;
+import static org.apache.http.HttpHeaders.CONTENT_LENGTH;
+import static org.apache.http.HttpHeaders.CONTENT_TYPE;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.io.IOUtils;
+
+import edu.unc.lib.dl.acl.service.AccessControlService;
+import edu.unc.lib.dl.acl.util.AccessGroupSet;
+import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.model.DatastreamType;
+import edu.unc.lib.dl.ui.exception.ResourceNotFoundException;
+import edu.unc.lib.dl.util.DerivativeService;
+import edu.unc.lib.dl.util.DerivativeService.Derivative;
+
+/**
+ * Streams content for derivative files of repository objects.
+ *
+ * @author bbpennel
+ *
+ */
+public class DerivativeContentService {
+
+    private static final int BUFFER_SIZE = 4096;
+
+    private DerivativeService derivativeService;
+
+    private AccessControlService accessControlService;
+
+    public DerivativeContentService() {
+    }
+
+    /**
+     * Set content headers and stream the content of the specified derivative from the object identified by pid.
+     *
+     * @param pid pid of object containing the derivative
+     * @param dsName name of derivative being requested.
+     * @param principals principals of requesting client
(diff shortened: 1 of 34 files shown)


access-common/src/main/java/edu/unc/lib/dl/ui/service/DerivativeContentService.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param dsName name of derivative being requested.
+     * @param principals principals of requesting client
+     * @param asAttachment if true, then content-disposition header will specify
+     *            as "attachment" instead of "inline"
+     * @param response response content and headers will be added to.
+     * @throws IOException if unable to stream content to the response.
+     * @throws ResourceNotFoundException if an invalid derivative type is requested.
+     */
+    public void streamData(PID pid, String dsName, AccessGroupSet principals, boolean asAttachment,
+            HttpServletResponse response) throws IOException, ResourceNotFoundException {
```
harringj: should there be a throws declaration for IllegalArgumentException, too?

access-common/src/main/java/edu/unc/lib/dl/ui/service/PermissionsHelper.java
```diff
@@ -79,26 +79,27 @@ public boolean hasThumbnailAccess(AccessGroupSet principals, BriefObjectMetadata
...
+        return hasDatastreamAccess(principals, JP2_ACCESS_COPY, metadata);
     }
 
     /**
      * Returns true if the provided principals have rights to access the
      * requested datastream, and the datastream is present.
      *
      * @param principals agent principals
-     * @param datastream name of datastream being requested
+     * @param datastream datastream being requested
```
harringj: maybe "type of datastream being requested" instead?

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/DatastreamPermissionUtil.java
```diff
@@ -41,27 +37,32 @@ private DatastreamPermissionUtil() {
...
+    }
+
+    /**
+     * Determine the Permission which applies to accessing the specified datastream.
+     *
+     * @param datastream datastream type to check. Required.
+     * @return permission
+     */
+    public static Permission getPermissionForDatastream(DatastreamType datastream) {
+        Assert.notNull(datastream);
```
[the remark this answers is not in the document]
bbpennel: I think so since its a public method and is called directly, not just by the other `getPermissionForDatastream` implementation

access-common/src/main/java/edu/unc/lib/dl/ui/service/DerivativeContentService.java
bbpennel: I added a note in the javadoc for dsName describing the requirement. I struggled some on whether to include it in the method signature here and other places, but generally we don't have throws statements for IllegalArgumentException and other RuntimeExceptions, and there are a lot of places where IllegalArgumentExceptions are thrown but not reported (anywhere where assert statements are made pretty much).

access-common/src/main/java/edu/unc/lib/dl/ui/service/PermissionsHelper.java
bbpennel: fixed

fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/DatastreamPermissionUtil.java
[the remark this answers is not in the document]
harringj: hmm, sorry, my eye must have skipped right over lines 55-63. it had looked like this was part of the method above :-)

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/DerivativeService.java
```diff
@@ -0,0 +1,156 @@
...
+ * repository objects.
+ *
+ * @author bbpennel
+ *
+ */
+public class DerivativeService {
+
+    private String derivativeDir;
+
+    private volatile static List<DatastreamType> derivativeTypes;
```
harringj: not challenging your choice here, but just curious about why the derivativeTypes need to be stored in main memory

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/DerivativeService.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * List all DatastreamTypes which are considered derivatives.
+     *
+     * @return
+     */
+    public static List<DatastreamType> listDerivativeTypes() {
+        if (derivativeTypes == null) {
+            derivativeTypes = Arrays.stream(DatastreamType.values())
+                    .filter(dt -> EXTERNAL.equals(dt.getStoragePolicy()))
```
harringj: so all externally-stored datastream types are derivative types?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
```diff
@@ -91,14 +107,35 @@ private void recordDownloadEvent(PID pid, String datastream, AccessGroupSet prin
...
+    }
+
+    @ResponseStatus(value = HttpStatus.FORBIDDEN)
+    @ExceptionHandler(AccessRestrictionException.class)
+    public void handleInvalidRecordRequest() {
+    }
+
+    @ResponseStatus(value = HttpStatus.BAD_REQUEST)
+    @ExceptionHandler({ObjectTypeMismatchException.class, IllegalArgumentException.class})
+    public void handleBadRequestException() {
```
harringj: i wonder whether just handleBadRequest() would be a clearer name for this method, since the exception(s) handled are not BadRequestExceptions? this would also match your usage for the previous two methods. also, are you planing at some point to have some logic inside these methods, or are they simply meant as placecholders?

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDatastreamFilter.java
```diff
@@ -57,7 +64,11 @@ public void filter(DocumentIndexingPackage dip) throws IndexingException {
         }
 
         boolean ownedByOtherObject = contentObj instanceof WorkObject;
+        // Retrieve list of datastreams preserved with this object
         List<Datastream> datastreams = getDatastreams(fileObj, ownedByOtherObject);
+        // Retrieve list of derivatives associated with the object
```
harringj: i'm finding the difference between these two retrievals of derivatives to be very subtle. what does "preserved with this object" mean vs. "associated with this object"?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
```diff
@@ -17,7 +17,9 @@
 
 import static edu.unc.lib.dl.acl.util.GroupsThreadStore.getAgentPrincipals;
 import static edu.unc.lib.dl.fcrepo4.RepositoryPathConstants.ORIGINAL_FILE;
+import static edu.unc.lib.dl.util.DerivativeService.isDerivative;
```
harringj: i'm not going to draw a line in the sand or anything, but would you consider naming this class "DatastreamController"? we've moved away from including the word "rest" in our other newer controllers :-)

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/DerivativeService.java
bbpennel: For the time being, the criteria might change later, which is why I wanted it abstracted into a helper method like this.

fcrepo-clients/src/main/java/edu/unc/lib/dl/util/DerivativeService.java
bbpennel: It is going to be a highly accessed value, since every request for a thumbnail from the UI and every datastream at indexing time will get compared against it, so figured it was worth storing.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDatastreamFilter.java
bbpennel: the key word is the preceding one, "datastreams" vs "derivatives". I'll switch to the same verb to draw attention there.

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
bbpennel: Sure, but this is a pre-migration class and there are 3 other classes in the same package with Rest in their name

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DatastreamRestController.java
bbpennel: Fixed. The methods are empty for the time being, at some point we may want to fill them in with an error response, but right now they just serve the purpose of setting a response code and returning a blank body when their exception is thrown.

PR #743, opened 2018-06-07
Bxc 2193: Add label to newly created containers

harringj:


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/AddContainerService.java
```diff
@@ -63,35 +68,39 @@
      * @param parentPid pid of parent obj to add child container to
      * @param containerType the type of new container to be created
```
bbpennel: Update javadoc with new param

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/AddContainerController.java
```diff
@@ -51,39 +51,39 @@
     @Autowired
     private AddContainerService addContainerService;
 
-    @RequestMapping(value = "edit/create/adminUnit/{id}", method = RequestMethod.POST)
+    @RequestMapping(value = "edit/create/adminUnit/{id}/{label}", method = RequestMethod.POST)
```
bbpennel: A label can have a lot of weird stuff in it, like /'s. I feel like I'd expect the label to be passed along as either part of the body of the request or a header, and we aren't generally using headers like this in the CDR. So perhaps `@RequestParam("label")`? I see some of our other POST methods like the mark for deletion stuff are using that.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/AddContainerService.java
harringj: good catch! done!

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/AddContainerController.java
harringj: makes sense. done!

PR #744, opened 2018-06-08
Deploy access control editor to build

lfarrell:


.gitmodules
```diff
@@ -5,3 +5,6 @@
 [submodule "clamavj"]
 	url = https://github.com/UNC-Libraries/clamavj.git
 	path = clamavj
+[submodule "admin-permissions-editor"]
+	path = admin-permissions-editor
```
harringj: this is totally beyond picky of me, but would you mind flipping the order of path and url to match the other entries? :-)

PR #747, opened 2018-06-11
Bxc 2198: deposit metrics

harringj:


metadata/src/main/java/edu/unc/lib/dl/metrics/GaugeFactory.java
```diff
@@ -0,0 +1,67 @@
...
+
+    }
+
+    /**
+     * Creates and registers a gauge under the given name
+     *
+     * @param metricName the name under which to register the gauge
+     * @param duration the length of time elapsed in ms
+     */
+    public static void createDurationGauge(MetricName name, Long duration) {
```
bbpennel: I think these signatures are good for now. I do wonder if they should be named something like "incrementDurationGauge" instead though since they don't actually return an instance of the gauge.

deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
```diff
@@ -184,7 +185,8 @@ private void depositDuration(String depositUUID, Map<String, String> status) {
         long depositEndTime = System.currentTimeMillis();
         long depositTotalTime = depositEndTime - depositStartTime;
 
-        metricsClient.setDepositDuration(depositUUID, depositTotalTime);
+        GaugeFactory.createDurationGauge(MetricRegistry.name("deposit " + depositUUID, "deposit-duration"),
```
bbpennel: might want to make it "deposit:" so that the uuid doesn't run into it.

metadata/src/main/java/edu/unc/lib/dl/metrics/GaugeFactory.java
bbpennel: A bit more generally, I'm wondering if we should be using histograms for this purpose? The description in their documentation is rather sparse, but my impression is it is for taking measurements of occurrences:
https://metrics.dropwizard.io/3.1.0/getting-started/#histograms
I found section 3.4 in here a little clearer:
http://www.baeldung.com/dropwizard-metrics

deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
harringj: 👍

metadata/src/main/java/edu/unc/lib/dl/metrics/GaugeFactory.java
harringj: Re: your first comment, while it's true that no Gauge is returned by the method, nevertheless one is created and registered by the registry. For this reason I do like "create" in the signature. There really isn't any incrementing per se that happens with a gauge, but rather just the recording of a value for a particular deposit.

As for your more general concern, I guess it depends on how we want to use the data. Are we more interested in recording individual durations for deposits, or do we want them all to appear in aggregate in a histogram to make it easier to see trends?

metadata/src/main/java/edu/unc/lib/dl/metrics/HistogramFactory.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final MetricRegistry registry = RegistryService.getInstance().getRegistry();
+
+    private HistogramFactory() {
+
+    }
+
+    /**
+     * Creates and registers a histogram under the given name
+     *
+     * @param metricName the name under which to register the gauge
```
bbpennel: Change "gauge" to "histogram" here and in the other create method.

metadata/src/main/java/edu/unc/lib/dl/metrics/CounterFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+    }
+
+    /**
+     * Creates and registers a counter for the given class
+     *
+     * @param className the class name under which to register the counter
+     * @param metricNames any further specific name(s) for the counter (Optional)
+     */
+    public static Counter createCounter(String className, String... metricNames) {
```
bbpennel: I'd suggest changing the className parameter from a string to `Class<?> metricNameClass` so that it matches `TimerFactory`

deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
```diff
@@ -57,13 +57,14 @@
 import edu.unc.lib.deposit.validate.ExtractTechnicalMetadataJob;
 import edu.unc.lib.deposit.validate.PackageIntegrityCheckJob;
 import edu.unc.lib.deposit.validate.ValidateContentModelJob;
-import edu.unc.lib.deposit.validate.ValidateFileAvailabilityJob;
 import edu.unc.lib.deposit.validate.ValidateDescriptionJob;
+import edu.unc.lib.deposit.validate.ValidateFileAvailabilityJob;
```
bbpennel: Sorry to be predictable, but could you undo the changes to this file other than the actual new functionality? Eclipse went pretty egregious on that formatting so i'm having a hard time picking out what actually changed.
Might be repeating what you already know, so feel free to ignore, but in eclipse, if you right click and go to "Compare With" > "Previous Revision" it should give you the diff, and then you can use the buttons at the top of the diff to copy over the previously versions of chunks of the file (github client probably offers similar)

metadata/src/main/java/edu/unc/lib/dl/metrics/HistogramFactory.java
harringj: whoops, good ctach

metadata/src/main/java/edu/unc/lib/dl/metrics/CounterFactory.java
harringj: ok

PR #750, opened 2018-06-18
Bxc 2190 - Prioritize triple indexing

bbpennel:
Resolves issues with solr indexing of new items before they are indexed in the triple store by breaking enhancements off into a separate queue and using cloned asynchronous calls to trigger them. The effect is that triple indexing is prioritized and generally happens very quickly.

Lays groundwork for allowing subsets of enhancements to be run and for enhancements to be triggered by user requests.

Replication is disabled for the time being while we revisit our approach

### DIFF
--- a/services-camel/src/main/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouter.java
+++ b/services-camel/src/main/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouter.java
@@ -22,6 +22,7 @@ import static edu.unc.lib.dl.rdf.Cdr.FileObject;
 import static edu.unc.lib.dl.rdf.Cdr.Folder;
 import static edu.unc.lib.dl.rdf.Cdr.Work;
 import static edu.unc.lib.dl.rdf.Fcrepo4Repository.Binary;
+import static edu.unc.lib.dl.services.camel.util.CdrFcrepoHeaders.CdrEnhancementSet;
 import static org.apache.camel.LoggingLevel.INFO;
 
 import org.apache.camel.BeanInject;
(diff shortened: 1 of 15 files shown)


services-camel/src/main/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouter.java
```diff
@@ -0,0 +1,106 @@
...
+            .log(INFO, "Processing queued enhancements ${headers[CdrEnhancementSet]} for ${headers[CamelFcrepoUri]}")
+            .to("fcrepo:{{fcrepo.baseUrl}}?preferInclude=ServerManaged&accept=text/turtle")
+            .multicast()
+            .to("direct:process.binary", "direct:process.solr");
+
+        from("direct:process.binary")
+            .routeId("ProcessOriginalBinary")
+            .filter(simple("${headers[CamelFcrepoUri]} contains '/original_file'"
+                    + " && ${headers[org.fcrepo.jms.resourceType]} contains '" + Binary.getURI() + "'"))
+            .threads(enhancementThreads, enhancementThreads, "CdrEnhancementThread")
```
harringj: why does enhancementThreads get passed as a param here twice?

services-camel/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -144,7 +144,7 @@
     </bean>
     
     <bean id="binaryMetadataProcessor" class="edu.unc.lib.dl.services.camel.BinaryMetadataProcessor">
-        <constructor-arg value="${fcrepo.binaryBase}"/>
+        <property name="baseBinaryPath" value="${fcrepo.binaryBase}" />
```
harringj: just wondering what your motivation was for changing the way the med processor gets initialized. i noticed above that you removed the constructor in favor of a setter

services-camel/src/main/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouter.java
bbpennel: Its the pool size and the max pool size

services-camel/src/main/webapp/WEB-INF/service-context.xml
bbpennel: I switched it mainly because I needed to be able to programmatically provide the baseBinaryPath with a temp directory in a test, but it also seems more in line with typical dependency injection patterns.

PR #752, opened 2018-06-25
Bxc 2230 - sword deposits

bbpennel:
Update SWORD to work with current ACL model.
Remove dead code and components of sword implementation that we are not using currently that have not been updated (such as updates to objects)

### DIFF
--- a/deposit/src/main/webapp/WEB-INF/service-context.xml
+++ b/deposit/src/main/webapp/WEB-INF/service-context.xml
@@ -54,14 +54,14 @@
     </bean>
 
     <bean id="jesqueConfig" class="net.greghaines.jesque.Config">
-        <constructor-arg value="${redis.host:localhost}" />
-        <constructor-arg value="${redis.port:6379}" />
-        <constructor-arg value="2000" />
-        <constructor-arg>
+        <constructor-arg value="${redis.host:localhost}" type="java.lang.String" />
+        <constructor-arg value="${redis.port:6379}" type="int" />
+        <constructor-arg value="2000" type="int" />
+        <constructor-arg type="java.lang.String">
             <null />
         </constructor-arg>
-        <constructor-arg value="resque" />
-        <constructor-arg value="0" />
+        <constructor-arg value="resque" type="java.lang.String" />
+        <constructor-arg value="0" type="int" />
     </bean>
 
     <bean id="jesqueClient" class="net.greghaines.jesque.client.ClientPoolImpl"
--- a/sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/managers/AbstractFedoraManager.java
+++ b/sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/managers/AbstractFedoraManager.java
@@ -16,6 +16,7 @@
 package edu.unc.lib.dl.cdr.sword.server.managers;
 
 import static edu.unc.lib.dl.acl.util.GroupsThreadStore.getAgentPrincipals;
+import static org.apache.http.HttpStatus.SC_FORBIDDEN;
 
 import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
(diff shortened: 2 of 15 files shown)


deposit/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -54,14 +54,14 @@
     </bean>
 
     <bean id="jesqueConfig" class="net.greghaines.jesque.Config">
-        <constructor-arg value="${redis.host:localhost}" />
-        <constructor-arg value="${redis.port:6379}" />
-        <constructor-arg value="2000" />
-        <constructor-arg>
+        <constructor-arg value="${redis.host:localhost}" type="java.lang.String" />
```
harringj: seems like a good idea, but just curious what advantage the typing of the constructor args gives us?

sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/managers/AbstractFedoraManager.java
```diff
@@ -56,20 +62,15 @@ protected PID extractPID(String uri, String basePath) {
...
         }
         return targetPID;
     }
 
-    protected boolean hasAccess(AuthCredentials auth, PID pid, Permission permission, SwordConfigurationImpl config) {
-        if (config.getAdminDepositor() != null && config.getAdminDepositor().equals(auth.getUsername())) {
-            return true;
+    protected void assertHasAccess(String message, PID pid, Permission permission) throws SwordError {
+        if (!aclService.hasAccess(pid, getAgentPrincipals().getPrincipals(), permission)) {
+            throw new SwordError(ErrorURIRegistry.INSUFFICIENT_PRIVILEGES, 403, message);
```
harringj: would you mind using Response.Status.Forbidden here?

sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/managers/ContainerManagerImpl.java
```diff
@@ -48,58 +36,10 @@
...
-
-        DepositReceipt receipt = new DepositReceipt();
-        receipt.setLocation(new IRI(editIRI));
-        receipt.setEmpty(true);
-
-        // Update the objects in progress status
-        this.setInProgress(targetPID, deposit, receipt);
-
-        return receipt;
+        throw new NotImplementedException("Operation not supported");
```
harringj: this invites the Liar's Paradox. "I am not implemented," says the implementation ;-)

deposit/src/main/webapp/WEB-INF/service-context.xml
bbpennel: After updating the version of jesque there was a second constructor that was getting selected over the correct one since they had the same number of parameters I guess, even though the first one was a Set instead of a String, so I needed to give it a bit more direction.

sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/managers/AbstractFedoraManager.java
bbpennel: Good catch, i've updated it. That particular constant wasn't available in the sword-server module, so I ended up using HttpStatus.SC_FORBIDDEN instead

sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/managers/ContainerManagerImpl.java
bbpennel: Are you calling Boxy a liar?

There's actually a OperationNotSupportedException in the java core libraries, but its got some weird connotations going on with it https://docs.oracle.com/javase/8/docs/api/index.html?javax/naming/OperationNotSupportedException.html

PR #755, opened 2018-07-03
Bxc 2203

harringj:


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/MODSRetrievalService.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public MODSRetrievalService() {
+    }
+
+    /**
+     * Retrieves the MODS description of the object whose pid is given
+     *
+     * @param agent security principals of the agent making the request
+     * @param pid the pid of the object whose MODS is requested
+     * @return the MODS as an input stream, or null if the object has no MODS
```
bbpennel: Return type needs the be updated

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/RetrieveMODSController.java
```diff
@@ -0,0 +1,75 @@
...
+            modsString = modsService.retrieveMODS(agent, pid);
+            if (modsString == null) {
+                return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+            }
+        } catch (Exception e) {
+            if (e instanceof AuthorizationException || e instanceof AccessRestrictionException) {
+                log.warn("User {} does not have permission to view metadata for {}", agent.getUsername(), pid);
+                return new ResponseEntity<>(HttpStatus.FORBIDDEN);
+            } else {
+                log.error("Failed to retrieve MODS for object with pid " + pid, e);
```
bbpennel: Use `{}` for variable interpolation

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/RetrieveMODSIT.java
```diff
@@ -0,0 +1,110 @@
...
+                eq(viewMetadata));
+        PID workPid = makePid();
+        File modsFile = setupWorkWithMODS(workPid);
+
+        MvcResult result = mvc.perform(get("/description/" + workPid.getUUID()))
+                .andExpect(status().is2xxSuccessful())
+                .andReturn();
+
+        String responseBody = result.getResponse().getContentAsString();
+        String expectedResponseBody = IOUtils.toString(new FileInputStream(modsFile));
```
bbpennel: You can use `FileUtils.readFileToString` directly to read the file

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/RetrieveMODSIT.java
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @author harring
+ *
+ */
+@ContextHierarchy({
+    @ContextConfiguration("/spring-test/test-fedora-container.xml"),
+    @ContextConfiguration("/spring-test/cdr-client-container.xml"),
+    @ContextConfiguration("/retrieve-mods-it-servlet.xml")
+})
+public class RetrieveMODSIT extends AbstractAPIIT {
```
bbpennel: Probably something we should test for in other cases, but would be good to check what happens when the user requests an object that doesn't exist. I'm guessing it will NPE currently. Should probably return a 404

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/MODSRetrievalService.java
harringj: good catch!

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/RetrieveMODSController.java
harringj: fixed!

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/RetrieveMODSIT.java
harringj: nifty :-)

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/RetrieveMODSIT.java
harringj: i agree, good idea. i've added an else-if to the controller to handle this case

PR #756, opened 2018-07-10
Bxc 2204: update edit description workflow in admin UI

harringj:


harringj:
To test, go to this URL: https://cdr-qa-fe4-fes.lib.unc.edu/admin/describe/ddecbdf8-f807-49b1-aed9-53e19a5ee7a7. Edit the MODS and submit.

bbpennel:
Looks good! I'd be okay with removing the commented out stuff, we can always restore it if we need to and I suspect it's going to work a little differently regardless.
One thing, I noticed that when I submit changes they do go through, but the editor never figures it out. I think the javascript that parses the response doesn't understand the result. Maybe the fields in the response are slightly different? Or the status code. I believe this behavior is in the xml editor itself, so we'll probably have to look at it to figure out what the success criteria is.

PR #757, opened 2018-07-12
Bxc 2234 - Display Works

bbpennel:
Works were not showing up in the admin ui because of changing their name from aggregate to work
Allows primary objects to be set in therapy dog deposits


deposit/src/main/java/edu/unc/lib/deposit/normalize/CDRMETSGraphExtractor.java
```diff
@@ -113,15 +113,17 @@ private void addStructLinkProperties(Model m) {
             String from = link.getAttributeValue("from", XLINK_NS);
             String arcrole = link.getAttributeValue("arcrole", XLINK_NS);
             String to = link.getAttributeValue("to", XLINK_NS);
-            if ("http://cdr.unc.edu/definitions/1.0/base-model.xml#hasAlphabeticalOrder"
+            Resource fromR = m.createResource(helper.getPIDURIForDIVID(from));
+            Resource toR = m.createResource(helper.getPIDURIForDIVID(to));
+            Property role;
+            if ("http://cdr.unc.edu/definitions/1.0/base-model.xml#defaultWebObject"
```
harringj: should this be "primaryObject" instead of DWO?

PR #759, opened 2018-07-13
Bxc 2206: set-as-primary-object workflow

harringj:


static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
```diff
@@ -0,0 +1,66 @@
+define('SetAsPrimaryObjectResultAction', [ 'jquery', 'AjaxCallbackAction', "tpl!../templates/admin/editTypeForm"], function($, AjaxCallbackAction, editTypeTemplate) {
```
bbpennel: can remove the editTypeForm/editTypeTemplate params

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
```diff
@@ -0,0 +1,66 @@
...
+		AjaxCallbackAction.prototype.execute.call(this);
+	};
+	
+	SetAsPrimaryObjectResultAction.prototype.doWork = function() {
+		var self = this;
+		var target = this.context.target;
+		var pid = this.target.getPid();
+		
+		$.ajax({
+			url : "/services/api/edit/setAsPrimaryObject",
```
bbpennel: Looks like the controller is expecting `/edit/setAsPrimaryObject/{id}` with no request body

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
```diff
@@ -0,0 +1,66 @@
...
+			type : "PUT",
+			contentType: "application/json; charset=utf-8",
+			dataType: "json",
+			data : JSON.stringify({
+				pid : pid,
+				clear : this.context.clear == true
+			})
+		}).done(function(reponse) {
+				// Trigger refreshing of results
```
bbpennel: This block looks like its a little over indented

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
```diff
@@ -0,0 +1,66 @@
...
+				self.context.actionHandler.addEvent({
+					action : 'RefreshResult',
+					target : target,
+					waitForUpdate : true
+			});
+			
+			self.context.view.$alertHandler.alertHandler("message", "Assignment of object with pid " + pid + " as primary object has started.");
```
bbpennel: Since this is a synchronous operation when it reaches this line the operation should have completed, how about having this as a success message? I think that there is an alert type for that too, probably "success".

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
harringj: done!

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
harringj: so, besides changing the url to "/services/api/edit/setAsPrimaryObject/" + pid , are you recommending removal of lines 45-49?

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
harringj: fixed

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
bbpennel: yes, I believe they are not needed with the current controller

static/js/admin/src/action/SetAsPrimaryObjectResultAction.js
harringj: sounds reasonable :-)

PR #761, opened 2018-07-16
Create container update

lfarrell:


static/templates/admin/createContainerForm.html
```diff
@@ -1,21 +1,9 @@
-<iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
-<form id="create_container_form" method="post" enctype="multipart/form-data" class="edit_form" target="upload_file_frame">
+<form id="create_container_form" method="post" class="edit_form">
```
bbpennel: We should inform the user in some way what kind of container they are creating

static/js/admin/src/CreateContainerForm.js
```diff
@@ -9,31 +9,48 @@ define('CreateContainerForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStat
...
+
+	CreateContainerForm.prototype.containerType = function (resultObject) {
+		var inputType = $("input[name='container_type']", this.$form);
+		var parentType = resultObject.type;
+
+		if (parentType === "RootObject") {
+			inputType.val("adminUnit");
+		} else if (parentType === "Unit") {
+			inputType.val("collection")
+		} else if (parentType === "Collection") {
```
bbpennel: indent characters changed here

static/js/admin/src/CreateContainerForm.js
```diff
@@ -9,31 +9,48 @@ define('CreateContainerForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStat
...
+            inputType.val("folder");
```
bbpennel: for the time being seems like the Collection and else cases can be merged

static/js/admin/src/AddMenu.js
```diff
@@ -8,17 +8,23 @@ define('AddMenu', [ 'jquery', 'jquery-ui', 'underscore', 'CreateContainerForm',
...
+		if (canIngest) {
 			var self = this;
 
-			items["addContainer"] = {
+			if ((canIngest && (this.container.type === "Folder" || this.container.type === "Collection")) ||
+				($.inArray('createCollection', this.container.permissions) !== -1 && this.container.type === "AdminUnit") ||
+				($.inArray('createAdminUnit', this.container.permissions) !== -1 && this.container.type === "ContentRoot")
+			) {
+				items["addContainer"] = {
 				name : "Add Container",
```
bbpennel: My inclination would be to say "Add Collection", "Add Folder", "Add Admin Unit" here

static/js/admin/src/AbstractForm.js
```diff
@@ -0,0 +1,111 @@
+/**
+ * Implements functionality and UI for the generic add container form
+ */
+define('AbstractForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStateChangeMonitor', 'ModalCreate',
```
bbpennel: Since this is largely the same as AbstractFileUploadForm, it would seem to make sense for that to inherit from this.

static/js/admin/src/AbstractForm.js
```diff
@@ -0,0 +1,111 @@
...
+			var dialogBox = new ModalCreate(this.options);
+			var formContents = dialogBox.formContents(resultObject);
+			var self = this;
+			this.closed = false;
+
+			this.dialog = $("<div class='containingDialog'>" + formContents + "</div>");
+			this.$form = this.dialog.first();
+			this.dialog.dialog = dialogBox.modalDialog(this.dialog, self);
+
+			if (this.containerType) {
```
bbpennel: This seems very specific to the create container form, so I am wondering if there would be a way to abstract this

PR #762, opened 2018-07-24
bxc-2011: verify destination can receive object to be moved

harringj:


persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsService.java
```diff
@@ -86,6 +102,35 @@ public String moveObjects(AgentPrincipals agent, PID destinationPid, List<PID> p
...
+                        "Object with pid: " + pid + " is not a folder or a work and cannot be added to a folder");
+            }
+        } else if (destObj instanceof WorkObject) {
+            if (!(moveObj instanceof FileObject)) {
+                throw new IllegalArgumentException(
+                        "Object with pid: " + pid + " is not a file and cannot be added to a work");
+            }
+        } else {
+            throw new IllegalArgumentException(
+                    "Object with pid: " + pid + " is not a container");
```
bbpennel: how about "Object with pid: \<pid\> cannot be added to destination"

persistence/src/test/java/edu/unc/lib/dl/persist/services/move/MoveObjectsServiceTest.java
```diff
@@ -250,6 +253,54 @@ public void testMoveObject() throws Exception {
...
+        List<PID> movePids = asList(movePid);
+        service.moveObjects(mockAgent, destPid, movePids);
+    }
+
+    @Test
+    public void testMoveAdminUnitIntoFolder() {
+        expectedException.expect(IllegalArgumentException.class);
+        expectedException.expectMessage("is not a folder or a work and cannot be added to a folder");
+
+        FolderObject mockWorkObj = mock(FolderObject.class);
```
bbpennel: very minor, but the variable name is indicating the wrong object type.

harringj: ha, good catch!

persistence/src/main/java/edu/unc/lib/dl/persist/services/move/MoveObjectsService.java
harringj: yes, much better. the pid is the moving object, not the container :-)

PR #763, opened 2018-07-26
Bxc 2212: logging config for move operation

harringj:


services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/MoveObjectsController.java
```diff
@@ -59,6 +59,7 @@ public MoveObjectsController() {
         if (moveRequest == null || moveRequest.moved == null || moveRequest.moved.size() == 0
                 || moveRequest.getDestination() == null || moveRequest.getDestination().length() == 0) {
             results.put("error", "Request must provide a destination destination and a list of ids");
+            log.error("Request must provide a destination destination and a list of ids");
```
bbpennel: While we're here, we might as well deduplicate the "destination" text here and on line 61

harringj: aw, double destination is twice as nice :-)

bbpennel:
Its about the journey, not the double destination.

Or at least that's what i've been telling myself for these last couple of
years of never ending migrations :)

PR #764, opened 2018-07-30
Bxc 2208 - Provide reason for deletion

bbpennel:
* Requires that a message describing the reason for one or more objects being marked for deletion, which will be added to the PREMIS event.
* Adds a simple form and actions to accommodate providing this message.


static/js/admin/src/DeleteForm.js
```diff
@@ -0,0 +1,76 @@
...
+	}
+	
+	DeleteForm.prototype.constructor = DeleteForm;
+	DeleteForm.prototype = Object.create( AbstractForm.prototype );
+	
+	/**
+	* resultObjects is expected to be an array of resultEntry objects
+	*/
+	DeleteForm.prototype.open = function(resultObjects) {
+    if (resultObjects.length == 1) {
```
lfarrell: It looks like the indentation is a bit wonky here

PR #775, opened 2019-01-14
Adjust css for case study without accompanying image

lfarrell:


lfarrell:
Fixed by adding an image

PR #778, opened 2019-03-05
Search page redesign

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2339


access/src/main/webapp/WEB-INF/jsp/searchResults/resultsList.jsp
```diff
@@ -18,17 +18,21 @@
...
-	</c:import>
-	${navigationBar}
-	<c:import url="searchResults/sortForm.jsp">
-		<c:param name="queryMethod">${queryMethod}</c:param>
-		<c:param name="currentSort">${searchState.sortType}</c:param>
-		<c:param name="currentSortOrder">${searchState.sortNormalOrder}</c:param>
-	</c:import>
-</div>
+
+<c:if test="${empty param.collectionBrowse}">
```
bbpennel: How about renaming this `param.excludeNavigationBar` or something similar to make it clear what the behavior is? This seems broader than the collectionBrowse.

static/css/cdr-homepage.css
```diff
@@ -246,10 +250,275 @@ footer .footer-menu {
...
 footer a:hover, footer a:active {
   color: #4B9CD3;
 }
 
+/**
+*
+* Search pages
+*
+**/
+.search-query-text {
```
bbpennel: I'm a little unclear on what's happening with the various CSS files. It seems like historically this file (`cdr-homepage.css`) just had stuff related to the front page and a bit of navigation in it, but it now has search page styles (which appear to be duplicated from dcr.scss). I think the compiled css is supposed to be going into `cdr_access.css`. 

I also wonder if it'd make sense to divide search results css into its own sass file versus all going into dcr.scss, but i'm not as familiar with sass practices.

access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
```diff
@@ -73,13 +70,41 @@ public String search(Model model, HttpServletRequest request) {
 
     private String search(SearchRequest searchRequest, Model model, HttpServletRequest request) {
         SearchResultResponse resultResponse = doSearch(searchRequest, model, request);
+        String queryText = formatQueryText(request);
 
         model.addAttribute("resultType", "searchResults");
-        model.addAttribute("pageSubtitle", "Search Results");
+        model.addAttribute("pageSubtitle", "Search results for <span class=\"query-request\">\""
```
bbpennel: I'd like to keep formatting/mark up out of our controllers. Maybe we could put the `queryText` into its model attribute and use it in `searchResults.jsp`

PR #785, opened 2019-05-10
Gallery view for admin sets

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2357


static/js/public/vueComponents/browseDisplay.vue
```diff
@@ -0,0 +1,260 @@
...
+
+                    if (this.numberOfRecords < page_display) {
+                        return this.numberOfRecords;
+                    }
+
+                    return page_display;
+                },
+
+                numberOfRecords: function() {
+                    return this.record_list.length || 0;
```
bbpennel: I don't think the `|| 0` will ever do anything

static/js/public/vueComponents/browseDisplay.vue
```diff
@@ -0,0 +1,260 @@
...
+                        return 'fa-archive';
+                    } else if (type === 'Folder') {
+                        return 'fa-folder';
+                    } else {
+                        return 'fa-file';
+                    }
+                },
+
+                numberOfColumns: function() {
+                    let screen_size = window.innerWidth;
```
bbpennel: These break points will only take affect when the page first reloads, versus a resize, I assume?

static/js/public/vueComponents/browseDisplay.vue
```diff
@@ -0,0 +1,260 @@
...
+
+            beforeDestroy: function() {
+                window.removeEventListener('resize', this.numberOfColumns);
+            }
+        });
+    });
+</script>
+
+<style>
+    .browse-records-display .columns {
```
bbpennel: is it typical to put the css for a vue component into the vue file rather than in with the rest of the sites css? Seems like if we are using sass it would be easier to make things consistent if the css were centrally managed.

static/js/public/vueComponents/browseSearch.vue
```diff
@@ -0,0 +1,53 @@
...
+
+            methods: {
+                getResults: function () {
+                    let self = this;
+
+                    fetch(`/listJson/${this.recordId}?anywhere=${encodeURI(this.search_query)}`)
+                        .then(function (response) {
+                            return response.json();
+                        }).then(function (data) {
+                        self.$emit('browse-query-results', data);
```
bbpennel: think you lost some indent here

static/js/public/vueComponents/pagination.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+    <div class="columns pagination">
+        <div class="column is-12">
+            <ul v-if="numberOfRecords > 0">
+                <li v-if="currentPage !== 1"><a class="back-next" :href="pageUrl(currentPage - 1)"><<</a></li>
```
bbpennel: Maybe the `<<` and `>>` should be html encoded?

static/js/public/vueComponents/pagination.vue
```diff
@@ -0,0 +1,119 @@
...
+                return {
+                    currentPage: 1,
+                    totalPages: 1
+                }
+            },
+
+            methods: {
+                setPage: function() {
+                    let params = window.location.search;
+                    let page_number = params.split('=');
```
bbpennel: Seems like this will run into problems when there are more than one GET parameter? Such as if the sort is included. Can you use `this.$route.query.page_number`?

lfarrell: We're not currently using Vue router, so I don't think that will work. That said to get the testing the whole view setup will probably need to be overhauled.

static/js/public/vueComponents/browseDisplay.vue
lfarrell: Yeah, it usually is. In theory it can be scoped to just that component and you can use SASS in them. The whole CSS/HMTL in JS movement.

static/js/public/vueComponents/browseDisplay.vue
```diff
@@ -0,0 +1,260 @@
...
+                typeText: function() {
+                    if (this.type === 'AdminUnit') {
+                        return 'administrative unit';
+                    }
+
+                    return this.type.toLowerCase();
+                },
+
+                displayList: function() {
+                    return this.record_list.slice(
```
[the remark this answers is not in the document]
lfarrell: No, it recalculates the records that should be with a given page

static/js/public/vueComponents/browseDisplay.vue
lfarrell: No, it should work on resize, too.

access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
```diff
@@ -31,7 +31,11 @@
 <div class="content-wrap full_record ${isDeleted}${' '}${isProtected}">
 <c:import url="fullRecord/navigationBar.jsp" />
 <c:choose>
+	<c:when test="${briefObject.resourceType.equals('AdminUnit')}">
+		<c:import url="fullRecord/browseView.jsp" />
+	</c:when>
 	<c:when test="${briefObject.resourceType == searchSettings.resourceTypeCollection || briefObject.resourceType == searchSettings.resourceTypeFolder}">
+		<c:import url="fullRecord/aggregateRecord.jsp" />
```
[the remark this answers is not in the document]
lfarrell: Not sure how that got in there.

static/js/public/vueComponents/browseDisplay.vue
bbpennel: Guess it's hard to tell right now since there are only 2 collections within an admin unit

lfarrell: You can see the class reset if you have the console open.

PR #787, opened 2019-05-20
Gallery View Browse with Vue

lfarrell:
Injects Vue as a single page app, includes sample test
https://jira.lib.unc.edu/browse/BXC-2357
https://jira.lib.unc.edu/browse/BXC-2366


static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -0,0 +1,255 @@
...
+            </div>
+        </div>
+        <div class="columns">
+            <div class="column is-10 spacing">
+                <p :class="{ no_results: numberOfRecords === 0}">
+                    There are <strong>{{ numberOfRecords }}</strong> {{ childTypeText }} in this level.
+                </p>
+                <p v-if="numberOfRecords > 0">
+                    Displaying <strong>{{ pagination_settings.start + 1}}</strong> to
+                    <strong>{{ displayNumber }}</strong> of <strong>{{ numberOfRecords }}</strong>
```
bbpennel: I don't think the mockups have this component in it, they just state "There are X items in here", so i think you can drop this line and just leave things up to the pagination to tell people where they are in the result.

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -0,0 +1,255 @@
...
+                this.pagination_settings = pagination_settings;
+            },
+
+            retrieveData() {
+                let self = this;
+                fetch(this.browsePath)
+                    .then(function(response) {
+                        return response.json();
+                    }).then(function(data) {
+                    self.container_metadata = data.container;
```
bbpennel: Think you lost your indent again?

static/js/vue-cdr-access/src/components/browseSearch.vue
```diff
@@ -0,0 +1,53 @@
...
+
+        methods: {
+            getResults() {
+                let self = this;
+
+                fetch(`/listJson/${this.recordId}?anywhere=${encodeURI(this.search_query)}`)
+                    .then(function(response) {
+                        return response.json();
+                    }).then(function(data) {
+                    self.$emit('browse-query-results', data);
```
bbpennel: indent changes again

static/js/vue-cdr-access/src/components/browseSort.vue
```diff
@@ -0,0 +1,76 @@
...
+
+        data() {
+            return {
+                sort_order: ''
+            }
+        },
+
+        methods: {
+            sortRecords() {
+                let sort_values = this.sort_order.split('-');
```
bbpennel: If i'm understanding correctly, the sort in this component is doing an in-memory sort rather than asking the server for sorted paginated results? While that is probably fine for admin sets since they don't have many children, other types like folders and collections will have too many children to pass to the client. Could this component be changed to request data from the server instead?

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -0,0 +1,255 @@
...
+            childTypeText() {
+                if (this.container_metadata.type === 'AdminUnit') {
+                    return 'collections';
+                } else {
+                    return 'items';
+                }
+            },
+
+            displayList() {
+                return this.record_list.slice(
```
bbpennel: I know i asked about this on the previous PR, but is this component asking the server for each page, or is it trying to slice up a single list that is retrieved once?

static/js/lib/require-vuejs.min.js
```diff
@@ -0,0 +1,26 @@
+/*
```
bbpennel: Do we still need this file?

static/js/vue-cdr-access/src/App.vue
```diff
@@ -0,0 +1,35 @@
+<template>
```
bbpennel: Is this sort of the top level definition of the browse component? Would it make sense to name it accordingly, or is this file going to be used for all vue components in the access app?

Makefile
```diff
@@ -35,6 +35,10 @@ ifneq ($(VERSION), "")
 endif
 
 build-access:
+	# Build vue application(s) files
+	npm --prefix static/js/vue-cdr-access install
+	npm --prefix static/js/vue-cdr-access run build
```
bbpennel: Will the tests always run? Not sure if that's a good or a bad thing, probably depends on how long they end up taking.

static/js/vue-cdr-access/src/App.vue
```diff
@@ -0,0 +1,35 @@
...
+    return {
+      record_id: `/listJson/${location.pathname.split('/')[2]}`
+    }
+  }
+}
+</script>
+
+<style scoped lang="scss">
+#app {
+  font-family: 'Avenir', Helvetica, Arial, sans-serif;
```
bbpennel: Did we decide to use a different font family just for the browse?

static/js/vue-cdr-access/src/components/browseDisplay.vue
lfarrell: Copy and past bites me again :)

static/js/vue-cdr-access/src/components/browseSort.vue
lfarrell: Yes, it could

static/js/lib/require-vuejs.min.js
lfarrell: Nope

static/js/vue-cdr-access/src/App.vue
lfarrell: In theory all components. We could rename it. It's the generated file name

Makefile
lfarrell: Not in this setup. They should run on Travis or via the command line

static/js/vue-cdr-access/src/App.vue
lfarrell: No, I'll get rid of it; a generated style

static/js/vue-cdr-access/src/components/browseDisplay.vue
lfarrell: The list is only retrieved once. Then creates a new array from the slice with just the needed records. We could drop this if we pulled new data from the server on each page change

bbpennel: I think we should do that, we are going to need to pull in one page of results at a time, both with sorting and pagination. I'm realizing that even with admin sets, we have over the top cases like UARMS with its 92 collections, which is more than the page size allowed by the server at the moment.

static/js/vue-cdr-access/src/utils/helper_methods.js
```diff
@@ -0,0 +1,55 @@
+const utils = {
+    /**
+     * Put URL paramaters into an object
+     * @returns {{start: number, page: number, sort: string, rows: number}}
+     */
+    urlParams() {
```
bbpennel: Since it appears that we have access to the $route, I wonder if instead of processing the url ourselves we could use `$route.params` or `$route.query`? They both appear to return a hash. I'm just skimming here though so might not know what i'm talking about

https://router.vuejs.org/api/#route-object-properties

PR #800, opened 2019-06-13
Use $route for url params

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2373


lfarrell:
Opening new request without all the extra commits

PR #801, opened 2019-06-17
Use $route for url params

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2373


static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -0,0 +1,82 @@
...
+
+export default {
+    methods: {
+        /**
+         * Put URL parameters into an object
+         * Set default params, if none are present that can be updated
+         * @param params_to_update
+         * @returns {({} & Dictionary<string | (string | null)[]>) | {start: number, page: number, sort: string, rows: number}}
+         */
+        urlParams(params_to_update = {}) {
```
bbpennel: I think this whole method might be able to be refactored to the following algorithm:
```
if $route.query is empty
  then route_params = default values
  else route_params = $route.query

page_params = route_params
if params_to_update is NOT empty
  then page_params = assign(page_params, params_to_update)

return page_params
```

static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -0,0 +1,82 @@
...
+
+            return page_params;
+        },
+
+        /**
+         * Expects an object of parameters. Generally used with urlParams() method above
+         * @param params
+         * @returns {string}
+         */
+        formatParamsString(params) {
```
bbpennel: this approach seems okay, but wanted to leave it up to you if you'd rather use any of the suggestions in this thread:
https://stackoverflow.com/questions/5505085/flatten-a-javascript-object-to-pass-as-querystring
They all seem to encode the param name as well, which might be a good idea. There are lodash and es6 approaches in here.

Seems like the vue router can do this as well, but that would only be useful if you were going through the router. Not sure if that is practical here:
`this.$router.push({ query: { kw: kw, p: 1 }})`

static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -0,0 +1,82 @@
...
+                } else {
+                    updated_params += `&${param_name}=${encodeURIComponent(params[param_name])}`
+                }
+            });
+
+            return updated_params;
+        },
+
+        /**
+         * Check to see if a parameter is in the url query
```
bbpennel: The comment should probably say something like "Check to see if value exists in array" since the $route.query is passed in from outside. Although I sort of wonder if the two uses of this method should be replaced with `if 'something' in array`

static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -0,0 +1,82 @@
...
+         * @param param
+         * @param params
+         * @returns {boolean}
+         */
+        paramExists(param, params) {
+            return `${param}` in params;
+        },
+
+        /**
+         * Add format to component params, other than BrowseImage
```
bbpennel: Don't think this comment needs to mention that it isn't used for `BrowseImage`

static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -0,0 +1,82 @@
...
+                page_params = {
+                    page: 1,
+                    rows: 20,
+                    start: 0,
+                    sort: 'title,normal'
+                };
+            } else if (!isEmpty(params_to_update)) {
+                page_params = Object.assign(route_params, params_to_update);
+
+                if (!this.paramExists('format', params_to_update)) {
```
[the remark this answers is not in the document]
lfarrell: Don't remember why I added this here, too. Seemed necessary at some point.

static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -0,0 +1,82 @@
...
+         * Put URL parameters into an object
+         * Set default params, if none are present that can be updated
+         * @param params_to_update
+         * @returns {({} & Dictionary<string | (string | null)[]>) | {start: number, page: number, sort: string, rows: number}}
+         */
+        urlParams(params_to_update = {}) {
+            let route_params = Object.assign({}, this.$route.query);
+            let page_params;
+
+            if (isEmpty(this.$route.query)) {
```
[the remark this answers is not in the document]
lfarrell: We could, but there wouldn't be any params added.

PR #804, opened 2019-07-09
Fixes to work against dev vm/fedora 5

bbpennel:
* Enables basic auth against fedora, which is now the default
* Renames the ROOT.war to access.war for consistency
* Updates camel dependencies and behaviors for fedora5
* Fix to allow creation of AdminUnits
* Updates to properties being used to eliminate some unused or duplicate properties.

This will not work on qa at the moment since it requires fedora 5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,7 +121,7 @@
         <jena.fuseki.version>3.8.0</jena.fuseki.version>
         
         <fcrepo-camel.version>5.0.0</fcrepo-camel.version>
-        <camel.version>2.20.4</camel.version>
+        <camel.version>2.22.5</camel.version>
         <tika.version>1.19.1</tika.version>
         
         <maven.failsafe.version>2.22.1</maven.failsafe.version>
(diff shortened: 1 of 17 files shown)


pom.xml
```diff
@@ -120,8 +120,8 @@
         <jena.version>3.9.0</jena.version>
         <jena.fuseki.version>3.8.0</jena.fuseki.version>
         
-        <fcrepo-camel.version>4.5.0</fcrepo-camel.version>
-        <camel.version>2.22.1</camel.version>
+        <fcrepo-camel.version>5.0.0</fcrepo-camel.version>
+        <camel.version>2.20.4</camel.version>
```
lfarrell: Why was camel downgraded?

bbpennel: Good point. When I was working on this I tried updating camel to 2.24.1 since it was the most reason version, but it didn't work because camel 2.23 stopped supporting spring 4. Apparently i then decided to set it to the exact version of camel being using by fcrepo-camel, but there's no need to do that. I've updated it to 2.22.5

PR #805, opened 2019-07-10
Folder browse page

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2389

### DIFF
--- a/access/src/main/webapp/WEB-INF/jsp/fullRecord/folderRecord.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/fullRecord/folderRecord.jsp
@@ -23,8 +23,8 @@
 <%@ taglib prefix="cdr" uri="http://cdr.lib.unc.edu/cdrUI" %>
 
 <c:choose>
-    <c:when test="${not empty briefObject.countMap}">
-        <c:set var="childCount" value="${briefObject.countMap.child}"/>
+    <c:when test="${not empty parentBriefObject.countMap}">
+        <c:set var="childCount" value="${parentBriefObject.countMap.child}"/>
     </c:when>
     <c:otherwise>
         <c:set var="childCount" value="0"/>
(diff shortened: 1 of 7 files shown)


access/src/main/webapp/WEB-INF/jsp/fullRecord/folderRecord.jsp
```diff
@@ -0,0 +1,69 @@
...
+    </c:when>
+    <c:otherwise>
+        <c:set var="pluralCount" value="items"/>
+    </c:otherwise>
+</c:choose>
+<c:url var="fullRecordUrl" scope="page" value="record/${briefObject.id}"/>
+<div class="contentarea">
+    <div id="is-folder" class="columns browse-header">
+        <div class="column is-12">
+            <h2><c:out value="${parentBriefObject.title}"/> (<c:out value="${childCount}"/>)</h2>
```
bbpennel: You can remove the child count display here. The mockups don't show it next to the title on the full record anymore, the number "(20529)" is actually the collection number and is part of the title.

You can actually do the same on the collection page too, only the admin unit appears to display the child count at the top.

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -146,14 +147,15 @@
             },
 
             updateUrl() {
-                let params = this.urlParams({ types: 'Work' });
+                let type = this.is_collection ? 'Work' : 'Folder,File';
```
bbpennel: Hmm, i'm not quite sure what this is doing. I see the same results whether i'm in a collection or folder, which is just works (which is correct), probably due to 
https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/805/files#diff-4d444f55adb40ebe5159fd69f6a9113bR159
But the url on a collection page has `type=Folder,File`. Could we remove this line and the insertion of the `types` param into the url on the next line, since retrieve data seems to win?

lfarrell: Hmm, yeah it would have had to have been added in both places, since one updates the url and the other retrieves the actual data.

PR #810, opened 2019-07-24
Structure browse

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2415


bbpennel:
Playing around with things a bit more, I'm also realizing that the Gallery Browse is only showing immediate children of the current object. It should show all children of the object. Meanwhile the new structure browse should only show immediate children.

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -172,21 +190,28 @@
...
             this.is_folder = document.getElementById('is-folder') !== null;
 
-            if (this.is_collection || this.is_folder) {
+            if (this.is_collection || this.is_folder || this.browse_type === 'structure-display') {
                 this.updateUrl();
             }
 
             this.retrieveData();
             window.addEventListener('resize', debounce(this.numberOfColumns), 300);
+            this.interval_check = window.setInterval(this.browseDisplayType, 1500);
```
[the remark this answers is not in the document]
lfarrell: Depending how far we wanted to go with this it might be better to use Vuex, https://vuex.vuejs.org/

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -147,46 +159,106 @@
...
+                /*
+                 * Show browse options for non admin unit pages
+                 * The options live outside Vue controlled DOM
+                 * See access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
+                 */
+                if (this.is_collection || this.is_folder) {
+                    this.displayBrowseButtons();
+                }
+
+                if (this.validPage) {
```
bbpennel: what is `validPage`? I can't find any other references to it

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -147,46 +159,106 @@
...
                     this.container_metadata = response.data.container;
+                    this.is_page_loading = false;
                 }).catch(function (error) {
                     console.log(error);
                 });
+            },
+
+            findPageType() {
+                // Small hack to check outside of Vue controlled DOM to see what page type we're on
+                this.is_admin_unit = document.getElementById('is-admin-unit') !== null;
```
bbpennel: I guess this is needed because we are getting the parent record in the same request as the children? I wonder if it'd be possible to pass the type of the parent in during initialization. Probably something we can think about later

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -66,26 +67,37 @@
...
+                is_page_loading: true,
                 record_count: 0,
                 record_list: [],
+                search_method: 'searchJson',
                 uuid: ''
             }
         },
 
         computed: {
+            applicablePage() {
```
bbpennel: Would it be possible to only include this component on pages where it is applicable, versus including it on all record pages and then disabling it? Or do you feel like it's better in the long run to do things this way?

static/js/vue-cdr-access/src/mixins/browseOptionUtils.js
```diff
@@ -0,0 +1,83 @@
...
+        displayBrowseButtons() {
+            document.getElementById('browse-display-type').classList.remove('hidden');
+        },
+
+        browseTypeFromUrl(is_admin_set = false) {
+            let current_url_params = this.urlParams();
+
+            if (is_admin_set) {
+                this.browse_type = null;
+                localStorage.removeItem('dcr-browse-type');
```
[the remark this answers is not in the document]
lfarrell: Yes, it should get populate local storage only if they go to collection/folder page. It seemed slightly easier to have everything sync

static/js/vue-cdr-access/src/components/browseDisplay.vue
lfarrell: At this point it's used mainly to show the buttons or not and set the browse query

static/js/vue-cdr-access/src/components/browseDisplay.vue
lfarrell: Not sure how we'd tell it's an applicable page otherwise. It matching the route /record/<uuid> in the Vue router, so it's going to think it needs to be loaded. As far as I know nothing in the url differentiates between a say a file vs a collection page.

static/js/vue-cdr-access/src/components/browseDisplay.vue
lfarrell: That got renamed. Fixed now.

PR #811, opened 2019-07-24
BXC-2401 - Edit label working

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2401

Gets basic edit label functionality working through the admin ui.

Could use some polish, in terms of how it should behave when MODS is present.
The refresh functionality is also not working, but that is a separate issue.


static/js/admin/src/AbstractForm.js
```diff
@@ -25,6 +25,7 @@ define('AbstractForm', [ 'jquery', 'jquery-ui', 'underscore', 'ModalCreate',
 			});
 			// Flag to track when the form has been submitted and needs to be locked
 			this.submitted = false;
+			this.submitSuccessful= false;
```
bbpennel: SPACE

PR #816, opened 2019-08-07
BXC-2283 - Cleanup

bbpennel:
* Removes fcrepo3 clients, constants, and supporting files
* Removes unused features and files, including deposit collectors, structure browse search page, etc.
* Refactors some classes to use current clients
* Removes all admin tabs except browse and status monitor, along with other shortcuts to the same pages
* Makes the browse page the landing page for the admin interface.


sword-server/src/main/java/edu/unc/lib/dl/cdr/sword/server/managers/ServiceDocumentManagerImpl.java
```diff
@@ -113,39 +124,39 @@ public ServiceDocument getServiceDocument(String sdUri, AuthCredentials auth, Sw
      */
     protected List<SwordCollection> getImmediateContainerChildren(PID pid, AuthCredentials auth,
             SwordConfigurationImpl config) throws IOException {
```
lfarrell: Are we still using SWORD for anything?

bbpennel: Yes, forms deposits, maybe some other types of deposits too, but i'd need to check

PR #819, opened 2019-08-13
Update links

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2412


bbpennel:
We'll need to do this on fcrepo4 as well

PR #824, opened 2019-08-15
BXC-2405 - MODS Export

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2405
* Set correct path to controller from XML export javascript
* Refactor exception handling for xml import and export, plus resolve warnings
* Add missing dependencies required for sending emails
* Remove unused params

### DIFF
--- a/services/src/main/java/edu/unc/lib/dl/cdr/services/processing/XMLExportJob.java
+++ b/services/src/main/java/edu/unc/lib/dl/cdr/services/processing/XMLExportJob.java
@@ -15,6 +15,8 @@
  */
 package edu.unc.lib.dl.cdr.services.processing;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.File;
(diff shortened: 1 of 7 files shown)


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/XMLExportJob.java
```diff
@@ -203,7 +199,7 @@ private File zipit(File mdExportFile) throws IOException {
         return mdExportZip;
     }
 
-    private void sendEmail(File mdExportFile, String toAddress) throws MessagingException {
+    private void sendEmail(File mdExportFile, String toAddress) {
```
lfarrell: Is there still an exception on error, now that MessagingException is gone?

bbpennel: No, the exception it was supposed to catch is actually swallowed here:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/69bcc5e865afc8e8c325e2b5ac3e81bc15e39505/persistence/src/main/java/edu/unc/lib/persist/services/EmailHandler.java#L62

PR #830, opened 2019-08-16
BXC-2437 - Change display order values for resource types

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2437

Change display order values for resource types so that works will appear before files, and to add more spacing in case more types are added in the future


smithjp:
The works are being listed before files in csv exports, but it seems like things are being listed in creation order and folders at the same level are mixed in with the works and files instead of being listed first.  I'm not sure if that is supposed to be happening or not.

bbpennel:
@smithjp The sort is specified here:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/admin/src/main/resources/search.properties#L117
What it should be doing is sorting first by the path leading up to (and sometimes including) the object. This is so that items in the same path get grouped together, making it a depth first result. Works and Files have the same ancestorIds value, for reasons I can't recall right now (and might be worth revisiting), which is why they were coming back in an unpredictable order when work and file had the same precedence.
So all that to say, resourceType is just used for tie breaking things in the same path at the moment.

PR #837, opened 2019-08-28
BXC-2334 - Service for retrieving info about ingest sources

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2334

* Updates clients for looking up ingest source info to work with current clients
* Moves ingest source related classes to persistence module
* Expands tests of ingest source info retrieval
* Refactors some of the private/assumed classes used by the ingest source manager into stand alone classes.
* Removes unused/non-functioning support for ingest of zip files.

Note: this feature is not usable from the user interface in this PR. That will follow in BXC-2439.


persistence/src/main/java/edu/unc/lib/dl/persist/services/ingest/IngestSourceManager.java
```diff
@@ -0,0 +1,281 @@
...
+                    Files.walkFileTree(Paths.get(base), new SimpleFileVisitor<Path>() {
+                        @Override
+                        public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)
+                                throws IOException {
+                            if (matcher.matches(dir)) {
+                                log.debug("Matched dir {} for source {}", dir, source.getId());
+                                addCandidate(candidates, dir, source, base);
+                                return FileVisitResult.SKIP_SUBTREE;
+                            }
+                            return FileVisitResult.CONTINUE;
```
bbpennel: Do we need to skip the subtree in more cases?

persistence/src/main/java/edu/unc/lib/dl/persist/services/ingest/IngestSourceManager.java
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     * Adds information about applicable packages to the list of candidates
+     *
+     * @param candidates
+     * @param filePath
+     * @param source
+     * @param base
+     * @throws IOException
+     */
+    private void addCandidate(List<IngestSourceCandidate> candidates, Path filePath,
```
bbpennel: Might be able to remove the base param

PR #842, opened 2019-09-30
BXC-2453 - Return id of the object a role is assigned to

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2453

* Returns the PID of the object where a role+principal is being assigned.
* After discussion, decided to leave multiple role assignments for the same principal in staff permission response since they now indicate where they are assigned.
* Explicitly disallow assigning multiple roles to a principal on the same object in the API, so it agrees with the UI.
* Some updates to java version setting and including javax stuff that causes problems if being built on system with java 11 installed.

### DIFF
--- a/static/js/admin/vue-permissions-editor/tests/unit/staffRoles.spec.js
+++ b/static/js/admin/vue-permissions-editor/tests/unit/staffRoles.spec.js
@@ -24,7 +24,15 @@ describe('staffRoles.vue', () => {
                 alertHandler: {
                     alertHandler: jest.fn() // This method lives outside of the Vue app
                 },
-                containerName: 'Test Unit',
+                objectPath: [{ 
+                    pid: 'collections',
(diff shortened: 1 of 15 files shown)


static/js/admin/vue-permissions-editor/tests/unit/staffRoles.spec.js
```diff
@@ -121,6 +129,59 @@ describe('staffRoles.vue', () => {
             done();
         });
     });
+    
+    it("displays names of containers that roles are assigned to in inherited table", (done) => {
+        wrapper = shallowMount(staffRoles, {
```
lfarrell: Instead of overriding the whole mounted wrapper you should be able to do this `wrapper.setProps({ my props here });`
This should just override the props you want to update/change while inheriting the rest of the beforeEach() wrapper setup.

PR #846, opened 2019-10-14
BXC-2445 - Update patron access

bbpennel:
* Adds service for updating patron access
* Adds API endpoint for updating patron access
* Refactor staff role update error handling to be more in line with patron role and to use the access control validator class
* Allowing for multiple events to be written to a log at once
* Removed unnecessary ParseException being thrown when formatting dates into UTC


persistence/src/main/java/edu/unc/lib/dl/persist/services/acl/PatronAccessAssignmentService.java
```diff
@@ -0,0 +1,342 @@
...
+        PremisLogger logger = repoObj.getPremisLog();
+
+        if (rolesEvent != null) {
+            if (embargoEvent != null) {
+                logger.writeEvents(rolesEvent, embargoEvent);
+            } else {
+                logger.writeEvents(rolesEvent);
+            }
+        } else {
+            logger.writeEvents(embargoEvent);
```
lfarrell: Should this be written only if embargoEvent isn't null?

bbpennel: This method only gets called if one or the other event variable is not null:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/8cc36d993bca3dbeef81d22bb5c2cfb4a55580ba/persistence/src/main/java/edu/unc/lib/dl/persist/services/acl/PatronAccessAssignmentService.java#L116

But there's a lot going on in this class, so I'm going to add a comment to explain why its not checking for this.

PR #854, opened 2019-11-01
Implement patron access control interface

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2475


static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -191,7 +191,11 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 
 		// Set/Update permission actions
 		items["seppermission"] = "";
-		items["patronPermissions"] = {name : 'Patron permissions'};
+
+		if (metadata.type !== 'AdminUnit') {
```
bbpennel: Realizing that both the patron and staff permissions need access control checks before adding to the menu. patron requires `changePatronAccess` and there is a similar one for staff permissions.

PR #860, opened 2019-11-14
Set conditions for displaying permission editors

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2495

### DIFF
--- a/static/js/admin/src/ResultObjectActionMenu.js
+++ b/static/js/admin/src/ResultObjectActionMenu.js
@@ -190,13 +190,18 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 		}
 
 		// Set/Update permission actions
-		items["seppermission"] = "";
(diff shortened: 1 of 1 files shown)


static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -190,13 +190,18 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 		}
 
 		// Set/Update permission actions
-		items["seppermission"] = "";
+		if ($.inArray('assignStaffRoles', metadata.permissions) !== -1 || (metadata.type !== 'AdminUnit' && $.inArray('changePatronAccess', metadata.permissions) !== -1)) {
```
bbpennel: this looks fine for the most part, but mind if I suggest determining permissions once in variables and then reusing those rather than duplicating the checks?
```
let canChangePatronAccess = // whatever
let canAssignStaffRoles = // whatever

if (canChangePatronAccess || canAssignStaffRoles) {
```

PR #863, opened 2019-11-25
Search facets

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2487


access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
```diff
@@ -106,6 +106,22 @@ private String formatQueryText(HttpServletRequest request) {
         searchRequest.setRootPid(RepositoryPaths.getContentRootPid().getURI());
         searchRequest.setApplyCutoffs(false);
         SearchResultResponse resultResponse = queryLayer.performSearch(searchRequest);
+
+        SearchState searchState = searchRequest.getSearchState();
```
bbpennel: was there a reason not to reuse `doSearch` from this same class here? It probably needs a bit of cleaning up since we don't need to populate jsp search results anymore, but it seems to do the same stuff.

Also, we don't need to retrieve the facets in all cases (such as the full record page), so it would be best not to issue the facet query unless it's needed. You might need to add a parameter to state whether facets should be retrieved or not

access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
```diff
@@ -117,6 +133,22 @@ private String formatQueryText(HttpServletRequest request) {
         searchRequest.setRootPid(pid);
```
bbpennel: The body of this method is basically the same as the other searchJson one, would it be possible to make one call the other?

static/js/vue-cdr-access/src/components/facets.vue
```diff
@@ -0,0 +1,205 @@
...
+                return {
+                    collection: collection_name,
+                    path: path,
+                    queryFacets: this._formatFacets(updated_facets)
+                };
+            },
+
+            facetInfo(facet) {
+                let facet_index = this.facet_info.findIndex((f) => {
+                   return f.displayValue === facet.displayValue;
```
bbpennel: might make sense to go by `value` rather than `displayValue` for lookups?

access/src/main/java/edu/unc/lib/dl/ui/controller/SearchActionController.java
bbpennel: Seems like we can probably clear out `search()` set of methods in this class, plus some of the private methods they use as well, since I don't think we are going to have any calls to them anymore.

lfarrell:
>     * Not really sure what's happening, but if I enable a facet, and then remove it, the results are different then when I started. It seems like all the results are there, but the sort has reversed maybe?
> 
>     * Seems like two requests are made each time I click on a facet
> 
>     * Search pages seem to be requiring identifiers to begin with "uuid", which isn't the case with other pages. Could it be changed to not add it in and require it? It seems like some of the routing is expecting it. The search widget on the record page isn't doing this.
> 
>     * When I select two facets at once (I tested collection and format at the same time) I get a blank page. It looks like things are being duplicated in the url
>       https://localhost:8080/search/uuid:f88ff51e-7e74-4e0e-9ab9-259444393aeb/uuid:f88ff51e-7e74-4e0e-9ab9-259444393aeb?a.setStartRow=0&rows=20&sort=title,normal&facetSelect=collection,format&anywhere=&format=text

Not sure why it's doing the first one. It's displaying results in the order they're returned from the server. The other issues should be fixed now.

lfarrell:
Not sure what to do about reloading the page with facets set. If the format is set to "Image" and you reload the server doesn't return any info about the image facet just it's children: png, jpg, whatever ... Could probably hard code part of the needed info, but not the record count.

PR #865, opened 2019-12-02
title edit

smithjp:
* https://jira.lib.unc.edu/browse/BXC-2440

### DIFF
--- a/persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditTitleService.java
+++ b/persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditTitleService.java
@@ -1,0 +1,135 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.dl.persist.services.edit;
+
+import edu.unc.lib.dl.acl.service.AccessControlService;
+import edu.unc.lib.dl.acl.util.AgentPrincipals;
+import edu.unc.lib.dl.acl.util.Permission;
+import edu.unc.lib.dl.fcrepo4.BinaryObject;
+import edu.unc.lib.dl.fcrepo4.ContentObject;
+import edu.unc.lib.dl.fcrepo4.RepositoryObjectLoader;
+import edu.unc.lib.dl.fedora.PID;
+import edu.unc.lib.dl.metrics.TimerFactory;
+import io.dropwizard.metrics5.Timer;
+import org.apache.commons.io.IOUtils;
+
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.logging.Logger;
+
+/**
+ * Service that manages editing of the mods:title property on an object
+ *
+ * @author smithjp
+ *
+ */
+public class EditTitleService {
+
+    private AccessControlService aclService;
+    private RepositoryObjectLoader repoObjLoader;
+
+    private static Logger log = Logger.getLogger(Logger.GLOBAL_LOGGER_NAME);
+
+    public EditTitleService() {
+    }
+
+    /**
+     * Changes an object's mods:title
+     *
+     * @param agent security principals of the agent making request
+     * @param pid the pid of the object whose label is to be changed
+     * @param title the new label (dc:title) of the given object
+     */
+    public void editTitle(AgentPrincipals agent, PID pid, String title) throws Exception {
+        log.info(pid.toString());
+        log.info(agent.getUsername());
+        log.info(agent.getPrincipals().toString());
+
+        aclService.assertHasAccess(
+                "User does not have permissions to edit titles",
+                pid, agent.getPrincipals(), Permission.editDescription);
+
+        log.info("access checked");
+
+        ContentObject obj = (ContentObject) repoObjLoader.getRepositoryObject(pid);
+        log.info("object loaded");
+        BinaryObject mods = obj.getDescription();
+        log.info("mods fetched");
+
+        String username = agent.getUsername();
+
+        if (mods != null) {
+            log.info(mods.toString());
+            String oldTitle = getOldTitle(mods);
+            // if old title is null, add titleInfo/title to mods
+            if (oldTitle != null) {
+                log.info("mods has title element: "+oldTitle);
+                // update title
+            } else {
+                log.info("mods needs title element");
+                // add titleInfo/title to mods
+            }
+        } else {
+            log.info("null mods");
+            // else create new mods stream
+            createMODS(title);
+        }
+
+        // Send message that the action completed
+//        operationsMessageSender.sendUpdateDescriptionOperation(agent.getUsername(), Arrays.asList(pid));
+    }
+
+    /**
+     * @param aclService the aclService to set
+     */
+    public void setAclService(AccessControlService aclService) {
+        this.aclService = aclService;
+    }
+
+    public RepositoryObjectLoader getRepoObjLoader() {
+        return repoObjLoader;
+    }
+
+    /**
+     *
+     * @param repoObjLoader
+     */
+    public void setRepoObjLoader(RepositoryObjectLoader repoObjLoader) {
+        this.repoObjLoader = repoObjLoader;
+    }
+
+    /**
+     *
+     * @param mods the mods record to be edited
+     * @return current title property
+     */
+    private String getOldTitle(BinaryObject mods) throws Exception {
+        String oldTitle = null;
+        InputStream modsStream = mods.getBinaryStream();
+        String modsString = IOUtils.toString(modsStream, StandardCharsets.UTF_8);
+        log.info(modsString);
+        // find title
+
+        return oldTitle;
+    }
+
+    private String createMODS(String title) {
+
+        // create mods
+
+        return "mods with new title";
+    }
+}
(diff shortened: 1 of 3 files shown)


persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditTitleService.java
```diff
@@ -0,0 +1,199 @@
...
+
+            if (modsStream != null) {
+                String modsString = IOUtils.toString(modsStream, StandardCharsets.UTF_8);
+                ByteArrayInputStream modsByteArray = new ByteArrayInputStream(modsString.getBytes());
+                SAXBuilder sb = new SAXBuilder(new XMLReaderSAX2Factory(false));
+                Document document = sb.build(modsByteArray);
+                Element rootEl = document.getRootElement();
+
+                String oldTitle = getOldTitle(rootEl);
+                if (oldTitle != "no mods:title") {
```
bbpennel: it won't matter if you switch over to `hasExistingTitle`, but just mentioning that in java string equality should use `!oldTitle.equals("text")` because java can't make it that easy on us

persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditTitleService.java
```diff
@@ -0,0 +1,199 @@
...
+                } else {
+                    newMods = addTitleToMODS(document, title);
+                }
+            } else {
+                Document document = new Document();
+                document.addContent(new Element("mods", MODS_V3_NS));
+                newMods = addTitleToMODS(document, title);
+            }
+
+            InputStream newModsStream = new ByteArrayInputStream(newMods.getBytes());
```
bbpennel: I'd recommend having `updateTitle` and `addTitleToMODS` return a Document rather than the result of `.toString()` (my understanding is that representation is intended for debugging), and pass the result through `XMLOutputter.outputString` like you are in the test.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -0,0 +1,182 @@
...
+        document.addContent(new Element("mods", MODS_V3_NS)
+                .addContent(new Element("titleInfo", MODS_V3_NS)
+                        .addContent(new Element("title", MODS_V3_NS).setText("original title"))));
+        when(binaryObj.getBinaryStream()).thenReturn(convertDocumentToStream(document));
+
+        service.editTitle(agent, pid, title);
+
+        verify(operationsMessageSender).sendUpdateDescriptionOperation(anyString(), pidsCaptor.capture());
+        Collection<PID> pids = pidsCaptor.getValue();
+        assertEquals(pids.size(), 1);
```
bbpennel: switch the order of the parameters, the expected value goes first. You'll need to do so in the other test methods too.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -0,0 +1,182 @@
...
+        assertTrue(pids.contains(pid));
```
bbpennel: you can `verify` that `obj.setDescription` was called, using an argument captor to get back the inputstream, and then deserialize it into a Document to verify that the title has changed. You'll need to do the same thing in many tests, so I'd suggest making a helper methods `getUpdatedDescriptionDocument` which does the argument captor stuff and deserializes it, and another named `assertTitleValue(document, expectedTitle)` to assert that there is a title with the expected value

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -0,0 +1,182 @@
...
+        doThrow(new AccessRestrictionException()).when(aclService)
+                .assertHasAccess(anyString(), eq(pid), any(AccessGroupSet.class), any(Permission.class));
+
+        service.editTitle(agent, pid, title);
+    }
+
+    @Test
+    public void noModsTest() throws Exception {
+        String title = "new title";
+        when(binaryObj.getBinaryStream()).thenReturn(null);
```
bbpennel: i think you want to do `when(contentObj.getDescription()).thenReturn(null)` to be closer to the expected behavior. The last stubbing wins, so there's no problem overriding the default from the test init method.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -0,0 +1,182 @@
...
+        service.editTitle(agent, pid, title);
+
+        verify(operationsMessageSender).sendUpdateDescriptionOperation(anyString(), pidsCaptor.capture());
+        Collection<PID> pids = pidsCaptor.getValue();
+        assertEquals(pids.size(), 1);
+        assertTrue(pids.contains(pid));
+    }
+
+    @Test
+    public void multipleTitlesInMods() throws Exception {
```
bbpennel: this one will be slightly harder to verify, but should be doable. `assertTitleValue` could either be written to search all titleInfo elements for the value (see `document.getRootElement().getChildren().stream().anyMatch(...)`), or you could handle this one as a separate case. But it is important for us to check that there are still two titles after the update.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
smithjp: What would the expected behavior be in this case?  An empty document?

bbpennel: Do you mean in terms of the return value from contentObj.getDescription, binaryObj.getBinaryStream, or what should the resulting that gets added to the record be? For the first, it should return null. For the second, it shouldn't reach that. And for the third it should create a new mods document with just the title in it.

persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditTitleService.java
```diff
@@ -153,47 +170,46 @@ public void setOperationsMessageSender(OperationsMessageSender operationsMessage
     /**
      *
      * @param mods the mods record to be edited
-     * @return current title value as a string
+     * @return true if mods has title
      */
-    private String getOldTitle(Element mods) {
-        String oldTitle;
+    private Boolean hasExistingTitle(Element mods) {
+        Boolean hasOldTitle;
```
bbpennel: you can skip going through a variable and immediately return true or false (its fine to return in the `catch` block). Also, the return type probably doesn't need to be a `Boolean` in this case, the primative `boolean` should be fine.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -106,8 +115,12 @@ public void editTitleTest() throws Exception {
 
         verify(operationsMessageSender).sendUpdateDescriptionOperation(anyString(), pidsCaptor.capture());
         Collection<PID> pids = pidsCaptor.getValue();
-        assertEquals(pids.size(), 1);
+        assertEquals(1, pids.size());
         assertTrue(pids.contains(pid));
+
+        verify(contentObj).setDescription(inputStreamCaptor.capture());
+        Document updatedDoc = getUpdatedDescriptionDocument(inputStreamCaptor.getValue());
+        assertTrue(assertTitleValue(updatedDoc, title));
```
bbpennel: I'd suggest renaming `assertTitleValue` to `hasTitleValue`. Alternatively, you could do the asserting inside of `assertTitleValue`, but you would need `assertTitleNotPresent` or similar for the negative case.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -162,21 +183,57 @@ public void multipleTitlesInMods() throws Exception {
...
+        Document document;
+        String modsString;
+
+        try {
+            modsString = IOUtils.toString(inputStream, StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new ServiceException("Unable to covert mods stream to string for " + pid, e);
+        }
+
+        ByteArrayInputStream modsByteArray = new ByteArrayInputStream(modsString.getBytes());
```
bbpennel: You don't need to pass the mods through a String before deserializing it into a document, `SAXBuilder` can take the original InputStream directly.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -162,21 +183,57 @@ public void multipleTitlesInMods() throws Exception {
...
+        SAXBuilder sb = new SAXBuilder(new XMLReaderSAX2Factory(false));
+        try {
+            document = sb.build(modsByteArray);
+        } catch (IOException| JDOMException e) {
+            throw new ServiceException("Unable to build mods document for " + pid, e);
```
bbpennel: for test methods/helpers, i generally just let the method `throw Exception` since there isn't going to be any other code relying on the API of the method to be more descriptive.

persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditTitleService.java
```diff
@@ -74,25 +79,33 @@ public void editTitle(AgentPrincipals agent, PID pid, String title) throws Excep
...
+            if (mods != null) {
                 modsStream = mods.getBinaryStream();
-            } catch (NullPointerException e) {
+            } else {
                 modsStream = null;
             }
 
             if (modsStream != null) {
-                String modsString = IOUtils.toString(modsStream, StandardCharsets.UTF_8);
+                String modsString;
```
bbpennel: You don't need to pass the mods through a String before deserializing it into a document, `SAXBuilder` can take the original InputStream directly.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditTitleServiceTest.java
```diff
@@ -162,21 +183,57 @@ public void multipleTitlesInMods() throws Exception {
...
 
         service.editTitle(agent, pid, title);
 
         verify(operationsMessageSender).sendUpdateDescriptionOperation(anyString(), pidsCaptor.capture());
         Collection<PID> pids = pidsCaptor.getValue();
-        assertEquals(pids.size(), 1);
+        assertEquals(1, pids.size());
         assertTrue(pids.contains(pid));
+
+        verify(contentObj).setDescription(inputStreamCaptor.capture());
```
bbpennel: you're welcome to put this line into `getUpdatedDescriptionDocument`, but that's just a style thing, its really up to you.

bbpennel:
Looks good, nice work!

PR #868, opened 2019-12-12
BXC-2468 - Original and FITS external binaries

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2468
https://jira.lib.unc.edu/browse/BXC-2470

* Original files and fits are stored as external binaries
* Adds deposit jobs for transferring binaries to external storage locations, ingested file objects updated to use them
* Binary apis updated to allow for content uris
* Add pid helper method to get the id with component path added (so object id plus file id)
* Add method that creates or updates binary objects with contenturis in fedora
* Add helper methods for generating datastream uris, and cleanup unused datastream constants
* Enhancements updated to work with content uris, no longer attempt to download binaries or build fedora binary path
* Remove replication routes so they don't need to be updated


deposit/src/main/java/edu/unc/lib/deposit/normalize/AssignStorageLocationsJob.java
```diff
@@ -0,0 +1,146 @@
...
+            return;
+        }
+
+        try {
+            while (iterator.hasNext()) {
+                Resource childResc = (Resource) iterator.next();
+                if (!needsStorageLocation(childResc)) {
+                    continue;
+                }
+                assignStorageLocation(childResc, storageId);
```
bbpennel: Maybe only need to figure out the storage location once

deposit/src/test/java/edu/unc/lib/deposit/normalize/AssignStorageLocationsJobTest.java
```diff
@@ -0,0 +1,239 @@
...
+
+        assertTrue(postDepBag.hasProperty(Cdr.storageLocation, LOC_ID));
+        assertTrue(postCollBag.hasProperty(Cdr.storageLocation, LOC_ID));
+        assertTrue(postFolderBag.hasProperty(Cdr.storageLocation, LOC_ID));
+        assertTrue(postWorkBag.hasProperty(Cdr.storageLocation, LOC_ID));
+        assertTrue(postFileResc.hasProperty(Cdr.storageLocation, LOC_ID));
+    }
+
+    @Test
+    public void resourceWithNoTypeDeposit() throws Exception {
```
bbpennel: add comment to clarify what this is testing

deposit/src/main/java/edu/unc/lib/deposit/normalize/AssignStorageLocationsJob.java
bbpennel: I misread this, the destination is only determined once, it is just assigned on all objects. So this should be fine.

PR #873, opened 2020-01-02
BXC-2469 - More external binaries

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2469

* Enables external binary ingest jobs
* Updates actions which change MODS to store to external binary locations
* Bulk metadata updates performed in camel
* Refactor of multi-destination transfer sessions so that they can be used in the same code as single destination sessions.
* Adds support for copying streams to external storage locations
* Deposit manifests stored as external


persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/UpdateDescriptionService.java
```diff
@@ -0,0 +1,185 @@
...
+        ContentObject obj = (ContentObject) repoObjLoader.getRepositoryObject(pid);
+
+        updateDescription(transferSession, agent, obj, modsStream);
+    }
+
+    private void updateDescription(BinaryTransferSession transferSession, AgentPrincipals agent,
+            ContentObject obj, InputStream modsStream) throws IOException {
+
+        log.debug("Updating description for {}", obj.getPid().getId());
+        try (Timer.Context context = timer.time()) {
```
lfarrell: Does the timer get used?

bbpennel: If I remember correctly, when the `try` block ends the timer gets closed which writes the result to the metrics client, which gets dumped to a log periodically

deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
```diff
@@ -91,9 +100,17 @@ public void runJob() {
...
+                .map(Statement::getResource).collect(toSet());
+
+        if (TYPES_ALLOWING_DESC.stream().anyMatch(rescTypes::contains)) {
+            transferModsFile(objPid, resc, transferSession);
+        }
+
+        if (rescTypes.contains(Cdr.FileObject)) {
             transferOriginalFile(objPid, resc, transferSession);
             transferFitsExtract(objPid, resc, transferSession);
+
```
lfarrell: Can probably get rid of this blank line

PR #877, opened 2020-01-08
BXC-2471 - Destroy action cleans up external files

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2471

* Destroy action deletes original file from external binary location
* Destroy action triggers clearing of destroyed records from search index
* Destroy execution moved to asynchronous job in services-camel
* Adds transfer client method for deleting external binary
* Normalizes file uris from file:/// to file:/ since fedora is doing this internally
* Fills in tests for destroy operations

### DIFF
--- a/persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
+++ b/persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
@@ -26,6 +26,7 @@ import static java.util.stream.Collectors.toList;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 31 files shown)


persistence/src/main/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJob.java
```diff
@@ -84,10 +112,18 @@ public void run() {
             // convert each destroyed obj to a tombstone
             for (PID pid : objsToDestroy) {
                 RepositoryObject repoObj = repoObjLoader.getRepositoryObject(pid);
+
+                assertCanDestroy(agent, repoObj, aclService);
```
bbpennel: Don't delete the binaries until the fedora cleanup has been successful, otherwise the rollback can result in objects with missing files.

PR #883, opened 2020-01-17
Allow staff only access when creating new collections or folders

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2502

### DIFF
--- a/services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/AddContainerIT.java
+++ b/services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/AddContainerIT.java
@@ -17,16 +17,14 @@ package edu.unc.lib.dl.cdr.services.rest.modify;
 
 import static edu.unc.lib.dl.acl.util.AccessPrincipalConstants.AUTHENTICATED_PRINC;
 import static edu.unc.lib.dl.acl.util.AccessPrincipalConstants.PUBLIC_PRINC;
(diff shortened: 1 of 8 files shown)


services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/AddContainerIT.java
```diff
@@ -128,6 +136,10 @@ public void testAddFolder() throws Exception {
         Map<String, Object> respMap = getMapFromResponse(result);
         assertEquals(collObj.getPid().getId(), respMap.get("pid"));
         assertEquals("create", respMap.get("action"));
+
+        PatronAccessDetails accessDetails = new PatronAccessDetails();
```
bbpennel: Would be good to update the test method name to something like `testAddFolderStaffOnly`

static/js/admin/src/CreateContainerForm.js
```diff
@@ -17,6 +17,7 @@ define('CreateContainerForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStat
 	CreateContainerForm.prototype.open = function(resultObject) {
 		this.options.containerType = this.getContainerType(resultObject);
 		this.options.title = 'Create ' + this.options.containerType;
+		this.options.privateOnlyBox = this.options.containerType !== "AdminUnit";
```
bbpennel: lets make this `staffOnly` or `staffOnlyBox` for the sake of using consistent language.

PR #888, opened 2020-01-23
Header for interior pages

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2512


lfarrell:
Added to https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/885, as I kept getting confused about which styles should be displaying

PR #890, opened 2020-01-28
Implement user interface to edit work titles

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2430

### DIFF
--- a/services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/EditTitleIT.java
+++ b/services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/EditTitleIT.java
@@ -1,0 +1,133 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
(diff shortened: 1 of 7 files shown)


services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/EditTitleIT.java
```diff
@@ -0,0 +1,133 @@
...
+        assertEquals("editTitle", respMap.get("action"));
+        hasTitleValue(getUpdatedDescriptionDocument(work.getDescription()), title);
+    }
+
+    @Test
+    public void testReplaceTitle() throws Exception {
+        PID pid = makePid();
+        String oldTitle = "old_work_title";
+        Model workModel = ModelFactory.createDefaultModel();
+        workModel.add(workModel.createResource(pid.getRepositoryPath()), DcElements.title,
```
bbpennel: You probably want to be testing replacing a MODS title rather than a DC title

PR #899, opened 2020-02-12
BXC-2526 - Generate content structure

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2526

* Adds command to transform a content tree into a bxc5 deposit structure in jena-tdb
* Includes progress display
* Adds some supporting commands
  * export file paths for a set of object ids, for generating pull lists to make test deposit sets
  * command display a deposit structure
* Adds a transformation manager which tracks progress and constructs new transform jobs
* Adds utility for managing concurrent access to tdb deposit dataset
* Reintroduces some bxc3 constants useful for the migration
* Path index can now retrieve all versions of a datastream if asked for, and by default only returns the most recent
* Adds test utility for generating foxml files


migration-util/src/main/java/edu/unc/lib/dcr/migration/ViewDepositModelCommand.java
```diff
@@ -0,0 +1,78 @@
...
+            description = "Serialize the RDF as turtle instead of n-triples")
+    private boolean asTurtle;
+
+    @Override
+    public Integer call() throws Exception {
+        PID depositPid = PIDs.get(DEPOSIT_RECORD_BASE, depositId);
+
+        DepositModelManager depositModelManager = new DepositModelManager(depositPid, parentCommand.tdbDir);
+
+        depositModelManager.commit();
```
bbpennel: remove this

PR #911, opened 2020-03-02
longleaf registration

smithjp:
* https://jira.lib.unc.edu/browse/BXC-2424

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJob.java
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.deposit.fcrepo4;
+
+import edu.unc.lib.deposit.work.AbstractDepositJob;
+import edu.unc.lib.dl.rdf.Cdr;
+import edu.unc.lib.dl.rdf.CdrDeposit;
+import org.apache.jena.rdf.model.*;
+import org.apache.jena.vocabulary.RDF;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Set;
+
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * Job which registers deposited files to longleaf
+ *
+ * @author smithjp
+ *
+ */
+public class RegisterToLongleafJob extends AbstractDepositJob {
+    private static final Logger log = LoggerFactory.getLogger(RegisterToLongleafJob.class);
+
+    public RegisterToLongleafJob() {
+        super();
+    }
+
+    public RegisterToLongleafJob(String uuid, String depositUUID) {
+        super(uuid, depositUUID);
+    }
+
+    @Override
+    public void runJob() {
+        log.debug("Registering files from deposit {} to longleaf", getDepositPID());
+
+        Model model = getReadOnlyModel();
+        Bag depositBag = model.getBag(getDepositPID().getRepositoryPath());
+
+        registerFilesToLongleaf(depositBag);
+    }
+
+    public void registerFilesToLongleaf(Resource resc) {
+        Set<Resource> rescTypes = resc.listProperties(RDF.type).toList().stream()
+                .map(Statement::getResource).collect(toSet());
+
+        System.out.println(rescTypes);
+
+        if (rescTypes.contains(Cdr.FileObject)) {
+            System.out.println("object is file");
+            if (resc.hasProperty(CdrDeposit.storageUri)) {
+                System.out.println("file has been transferred");
+            } else if (resc.hasProperty(CdrDeposit.descriptiveStorageUri)) {
+                System.out.println("mods file has been transferred");
+            } else {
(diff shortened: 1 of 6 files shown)


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJob.java
```diff
@@ -0,0 +1,166 @@
...
+                Statement currentStatement = statementIterator.nextStatement();
+
+                // find storageUri, descriptiveStorageUri, fitsStorageUri, premisStorageUri
+                if (currentStatement.getPredicate().toString().matches(".*Uri")) {
+                    URI storageUri = URI.create(currentStatement.getString());
+                    if (currentStatement.getString().matches(".*original_file")) {
+                        String checksum = resc.getProperty(CdrDeposit.md5sum).getString();
+                        registerFile(storageUri, checksum);
+                    } else {
+                        registerFile(storageUri, "");
```
bbpennel: one minor suggestion, i'd recommend using `null` to represent the absence of something rather than an empty string

deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
```diff
@@ -103,6 +103,10 @@ private void transferBinaries(Resource resc, BinaryTransferSession transferSessi
             transferDepositManifests(objPid, resc, transferSession);
         }
 
+        // Add premis log location to deposit model
+        Path premisLogPath = getPremisFile(objPid).toPath();
```
bbpennel: Could you verify that this is the right path? My memory is that this is the pre-ingest location where the premis file is stored, rather than post-ingest. I don't think the path of the post-ingest log exists until after IngestContentObjectJob and IngestDepositRecordJob have completed. 

The following might be a way to get the path in `RegisterToLongleafJob`, but i haven't tested it:
```
BinaryObject premisBin = repositoryObjectLoader.getBinaryObject(DatastreamPids.getMdEventsPid(pid));
URI premisStorageUri = premisBin.getContentUri();
```
But only after the ingest content/deposit jobs have completed, meaning it wouldn't work in this job since the objects haven't been ingested into fedora yet.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJobTest.java
```diff
@@ -0,0 +1,240 @@
...
+        job.run();
+
+        String registrationArguments = "register -f " + LOC3_ID + "\n";
+        String output = FileUtils.readFileToString(new File(outputPath));
+
+        assertEquals(registrationArguments, output);
+    }
+
+    @Test
+    public void registerFilesWhichHaveNotBeenIngestedToLongleaf() throws Exception {
```
bbpennel: Just to make sure i understand this test, its to verify that a file that is in the deposit but has no storage uri is not registered to longleaf?

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJobTest.java
```diff
@@ -0,0 +1,240 @@
...
+        Resource fileResc = addFileObject(workBag, FILE_CONTENT1, true);
+        fileResc.addProperty(CdrDeposit.storageUri, FILE_SCHEME + LOC2_ID + "/original_file");
+        workBag.addProperty(Cdr.primaryObject, fileResc);
+
+        job.closeModel();
+
+        job.run();
+
+        String registrationArguments = "register -f " + LOC2_ID  + "/original_file --checksums 'md5:" + MD5 + "'\n";
+        String output = FileUtils.readFileToString(new File(outputPath));
```
bbpennel: add a StandardCharsets.UTF-8 to each of these `readFileToString`, my IDE gets grumpy without it :)  
Also, just a suggestion, but if you hold onto the `File` object representation of the outputPath when its first created you could save a bit of effort on instantiating new file objects.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJobTest.java
```diff
@@ -0,0 +1,240 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
+import java.util.Set;
+
+import static edu.unc.lib.dl.test.TestHelpers.setField;
+import static org.junit.Assert.assertEquals;
+
+public class RegisterToLongleafJobTest extends AbstractDepositJobTest {
```
bbpennel: These tests look good. Could you add one more which verifies things work properly if the same object has multiple types of file? So mods, fits, premis, original file, or some subset of them.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJobTest.java
smithjp: Yes, I think this case probably would never happen since anything that gets this far would have a storageUri value.  I can remove it if it seems unnecessary.

bbpennel: No i think this is fine, just wanted to make sure i understood

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/RegisterToLongleafJobTest.java
```diff
@@ -0,0 +1,285 @@
...
+        fileResc.addProperty(CdrDeposit.storageUri, FILE_SCHEME + LOC2_ID + "/original_file");
+        workBag.addProperty(Cdr.primaryObject, fileResc);
+        fileResc.addProperty(CdrDeposit.fitsStorageUri, FILE_SCHEME + LOC3_ID + "/techmd_fits");
+        fileResc.addProperty(CdrDeposit.descriptiveStorageUri, FILE_SCHEME + LOC3_ID + "/md_descriptive");
+
+        job.closeModel();
+
+        job.run();
+
+        String registrationArguments = "register -f " + LOC1_ID + "/event_log\n" + // bag
```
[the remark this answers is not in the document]
bbpennel: Not sure if you saw this comment

PR #914, opened 2020-03-04
BXC-2390 - Links between apps and Header fixes

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2390

* Fixes double slash after application name in links between applications
* Resolves most cases where "access" shows up in urls. The Log on and Contact Us links are still incorrect until a separate ansible ticket is merged
* Removes the "Home" link from the front page (it was already gone from interior pages)


admin/src/main/webapp/WEB-INF/jsp/common/header.jsp
```diff
@@ -18,6 +18,8 @@
 <%@ page language="java" contentType="text/html; charset=UTF-8" pageEncoding="UTF-8"%>
 <%@ page trimDirectiveWhitespaces="true" %>
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ taglib prefix="s" uri="http://www.springframework.org/tags" %>
```
bbpennel: delete

PR #924, opened 2020-03-12
Update workflow for running enhancements

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2583


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/RunEnhancementsService.java
```diff
@@ -0,0 +1,132 @@
...
+                        pid, agent.getPrincipals(), Permission.runEnhancements);
+
+                LOG.debug("sending solr update message for {} of type {}", pid, "runEnhancements");
+
+                String objectType = (String) objectPid.get("objectType");
+
+                if (!objectType.equals("File")) {
+                    SearchRequest searchRequest = new SearchRequest();
+                    searchRequest.setAccessGroups(agent.getPrincipals());
+                    searchRequest.setSearchState(new SearchState());
```
bbpennel: Since all of our enhancements are for files at the moment, you could limit the search to just return file objects. You can do that by adding a facet filter, something like:
`searchState.getFacets().put(SearchFieldKeys.RESOURCE_TYPE.name(),
                    ResourceType.File.name());`

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/RunEnhancementsService.java
```diff
@@ -0,0 +1,132 @@
...
+                    for (BriefObjectMetadata metadata : resultResponse.getResultList()) {
+                        createMessage(metadata, pid, agent.getUsername(), force);
+                    }
+                } else {
+                    SimpleIdRequest searchRequest = new SimpleIdRequest(uuid, agent.getPrincipals());
+                    BriefObjectMetadata metadata = queryLayer.getObjectById(searchRequest);
+                    createMessage(metadata, pid, agent.getUsername(), force);
+                }
+            }
+        } catch (Exception e) {
```
bbpennel: Hmm, I think you can remove this catch and leave it up to the thing that calls it to decide whether or not to log the error. Similarly, you might want to add some logging to the controller to log the error in the catch block it already has.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/RunEnhancementsService.java
```diff
@@ -0,0 +1,132 @@
...
+        Document msg = new Document();
+        Element entry = new Element("entry", ATOM_NS);
+        entry.addContent(new Element("author", ATOM_NS)
+                .addContent(new Element("name", ATOM_NS).setText(userid)));
+        entry.addContent(new Element("pid", ATOM_NS).setText(filePath));
+        entry.addContent(new Element("mimeType", ATOM_NS).setText(mimeType));
+
+        if (force) {
+            Element paramForce = new Element("force", CDR_MESSAGE_NS);
+            paramForce.setText("force");
```
bbpennel: doesn't matter too much, but maybe submit the value "true"?

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/RunEnhancementsController.java
```diff
@@ -0,0 +1,90 @@
...
+                log.error("Failed to run enhancements for {}", data.getPids().toString(), e);
+                return new ResponseEntity<>(result, HttpStatus.INTERNAL_SERVER_ERROR);
+            }
+        }
+
+        result.put("timestamp", System.currentTimeMillis());
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    public static class RunEnhancementsRequest {
```
bbpennel: Totally optional, but wanted to mention that you could pass RunEnhancementsRequest into the service as the input rather than PIDS and force separately. I don't really have that much of an opinion on this though. If you do, you might want to move the declaration of this class into the service though.

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/RunEnhancementsController.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import edu.unc.lib.dl.acl.util.GroupsThreadStore;
+import edu.unc.lib.dl.cdr.services.processing.RunEnhancementsService;
+
+/**
+ * @author bbpennel
```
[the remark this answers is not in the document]
lfarrell: Apparently so. In 2015 :)

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/RunEnhancementsController.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author lfarrell
+ */
+@Controller
+public class RunEnhancementsController {
+    private static final Logger log = LoggerFactory.getLogger(RunEnhancementsController.class);
+
+    @Autowired
+    private RunEnhancementsService enhService;
+
+    @RequestMapping(value = "runEnhancements", method = RequestMethod.POST, produces = "application/json")
```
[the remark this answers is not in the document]
lfarrell: Was wondering about that. Seems some of our controllers pass in request, response, but don't always use them.

static/js/admin/src/EditThumbnailForm.js
```diff
@@ -0,0 +1,40 @@
+define('EditThumbnailForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStateChangeMonitor', 'tpl!../templates/admin/editThumbnailForm',
```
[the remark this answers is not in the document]
lfarrell: Ugh, yeah forgot to remove these

services/src/test/java/edu/unc/lib/dl/cdr/services/rest/RunEnhancementsIT.java
```diff
@@ -108,10 +127,12 @@ public void setup() throws Exception {
 
     @Test
     public void runEnhancementsFileObject() throws Exception {
+        FileObject fileObj = repositoryObjectFactory.createFileObject(PIDs.get(UUID_TEST), null);
```
bbpennel: If you call `createFileObject(null)` then it will make a PID for you and you don't have to worry about collisions between test runs. You can get the pid back by calling `fileObj.getPid()`. This would allow you to remove hard coded pids too.

PR #931, opened 2020-03-16
Fix display of file size and other metadata

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2605


static/js/admin/src/ResultObject.js
```diff
@@ -220,12 +220,13 @@ define('ResultObject', [ 'jquery', 'jquery-ui', 'underscore', 'ModalLoadingOverl
...
 					return {
 						name : fields[0],
 						mimeType : fields[1],
-						extension : fields[2],
-						fileSize : fields[3],
-						checksum : fields[4],
+						extension : fields[3],
+						fileSize : fields[4],
+						checksum : fields[5],
 						defaultWebObject : fields.length > 5? fields[5] : null
```
bbpennel: seems like this last entry needs to be updated too?

PR #932, opened 2020-03-17
Update display of items marked for deletion in public UI

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2563

### DIFF
--- a/access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
@@ -21,7 +21,7 @@
 <%@ taglib prefix="cdr" uri="http://cdr.lib.unc.edu/cdrUI" %>
 
 <c:if test="${markedForDeletion}">
-	<c:set var="isDeleted" value="deleted" scope="session"/>
+	<c:set var="isDeleted" value="deleted" scope="request"/>
 </c:if>
 
 <c:set var="allowsPublicAccess" value="${permsHelper.allowsPublicAccess(briefObject)}" />
--- a/access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
@@ -20,9 +20,10 @@
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
 <%@ taglib prefix="cdr" uri="http://cdr.lib.unc.edu/cdrUI" %>
 
-<c:if test="${not empty briefObject && cdr:contains(briefObject.status, 'Deleted') || cdr:contains(briefObject.status, 'Parent Deleted')}">
-	<c:set var="isDeleted" value="deleted" scope="page"/>
+<c:if test="${markedForDeletion}">
+	<c:set var="isDeleted" value="deleted" scope="session"/>
 </c:if>
+
 <c:set var="allowsPublicAccess" value="${permsHelper.allowsPublicAccess(briefObject)}" />
 <c:if test="${not empty briefObject && (not allowsPublicAccess || not empty briefObject.activeEmbargo)}">
 	<c:set var="isProtected" value="protected" scope="page"/>
--- a/access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
@@ -29,6 +29,17 @@
 	<c:set var="isProtected" value="protected" scope="page"/>
 </c:if>
 
+<c:set var="badgeIcon" scope="request">
+	<c:choose>
+		<c:when test="${markedForDeletion}">
+			trash
+		</c:when>
+		<c:when test="${not empty briefObject && (not permsHelper.allowsPublicAccess(briefObject) || not empty briefObject.activeEmbargo)}">
+			lock
+		</c:when>
+	</c:choose>
+</c:set>
+
 <div class="content-wrap full_record ${isDeleted}${' '}${isProtected}">
 
 <c:choose>
--- a/access/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
@@ -49,15 +49,16 @@
                 	<c:when test="${markedForDeletion}"><h2 class="deleted"></c:when>
                 	<c:otherwise><h2></c:otherwise>
                 </c:choose>
-                <i class="fa fa-university" aria-hidden="true"></i> <c:out value="${briefObject.title}" />
-                <c:if test="${not empty badgeIcon}">
-                    <div class="thumbnail-badge thumbnail-badge-${badgeIcon}">
-                        <div class="fa-stack">
-                            <i class="fas fa-circle fa-stack-2x background"></i>
-                            <i class="fas fa-${badgeIcon} fa-stack-1x foreground"></i>
+                <i class="fa fa-university" aria-hidden="true">
+                    <c:if test="${not empty badgeIcon}">
+                        <div class="thumbnail-badge thumbnail-badge-${badgeIcon}">
+                            <div class="fa-stack">
+                                <i class="fas fa-circle fa-stack-2x background"></i>
+                                <i class="fas fa-${badgeIcon} fa-stack-1x foreground"></i>
+                            </div>
                         </div>
-                    </div>
-                </c:if>
+                    </c:if>
+                </i> <c:out value="${briefObject.title}" />
             </h2>
             <p><strong>Subjects:</strong>
                 <c:choose>
--- a/access/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
@@ -50,6 +50,14 @@
                 	<c:otherwise><h2></c:otherwise>
                 </c:choose>
                 <i class="fa fa-university" aria-hidden="true"></i> <c:out value="${briefObject.title}" />
+                <c:if test="${not empty badgeIcon}">
+                    <div class="thumbnail-badge thumbnail-badge-${badgeIcon}">
+                        <div class="fa-stack">
+                            <i class="fas fa-circle fa-stack-2x background"></i>
+                            <i class="fas fa-${badgeIcon} fa-stack-1x foreground"></i>
+                        </div>
+                    </div>
+                </c:if>
             </h2>
             <p><strong>Subjects:</strong>
                 <c:choose>
--- a/access/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
@@ -45,7 +45,12 @@
     <div id="is-admin-unit" class="columns browse-header">
         <div class="column is-12">
             <c:import url="fullRecord/navigationBar.jsp" />
-            <h2><i class="fa fa-university" aria-hidden="true"></i> <c:out value="${briefObject.title}" /></h2>
+            	<c:choose>
+                	<c:when test="${markedForDeletion}"><h2 class="deleted"></c:when>
+                	<c:otherwise><h2></c:otherwise>
+                </c:choose>
+                <i class="fa fa-university" aria-hidden="true"></i> <c:out value="${briefObject.title}" />
+            </h2>
             <p><strong>Subjects:</strong>
                 <c:choose>
                     <c:when test="${not empty briefObject.subject}">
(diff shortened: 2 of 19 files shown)


static/css/sass/cdr-ui.scss
```diff
@@ -457,6 +457,19 @@ $border-style: 1px solid $light-gray;
 $box-shadow-color: rgba(211,211,211,0.5);
 $box-shadow: inset 3px 3px 10px -1px $box-shadow-color, 0px 16px 22px -15px $box-shadow-color;
 
+/* Items marked for deletion */
+.marked-delete {
```
bbpennel: there are still existing styles for deleted items, they are over in here:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/46e5d59c48bde7288c6c80f348d6480ce86cada5/static/css/cdrui_styles.css#L200-L218
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/46e5d59c48bde7288c6c80f348d6480ce86cada5/static/css/cdrui_styles.css#L971-L986

the class name is `deleted`. I don't know if these are still helpful or not, but there seems to be a bit more happening there than in this rule, so its worth comparing. It covers record and result pages. We should probably consider if there are more styles there that we want to retain, and cleanup the old rules once we have everything we want in here. 

Also, `marked-delete` reads a little weird to me, maybe because its an adjective plus verb. Maybe `deleted` or `marked-deleted`?

One other thing, in bxc3 we get a trash icon over the thumbnail when something is deleted. I haven't deployed this branch yet, but i don't see anything in the PR that would add that in.

For an example of a deleted resource, see:
https://dcr-qa.lib.unc.edu/search?anywhere=doomed_from_the_start
you might have to spoof yourself some credentials, I can't remember your permission setup.

lfarrell: No, there, currently, isn't any trash icon added.

lfarrell: I saw the exisiting classes, but I think they weren't being applied for some reason in the Vue components if a link had the "deleted" class. I'll poke around

lfarrell: I was thinking at some point we might want to find/run a tool to find "dead" css.

static/js/public/fullRecord.js
```diff
@@ -156,6 +182,19 @@ define('fullRecord', ['module', 'jquery', 'JP2Viewer', 'StructureView', 'dataTab
...
+		function showBadge(data) {
+			var markedForDeletion = false;
+			var restrictedAccess = false;
+
+			if (data.status !== undefined) {
+				var restrictions = data.status.join(',').toLowerCase();
+				markedForDeletion = /marked.*?deletion/.test(restrictions);
+				restrictedAccess = /embargoed|staff-only/.test(restrictions);
+			}
+console.log({ markDeleted: markedForDeletion, restricted: restrictedAccess })
```
bbpennel: can remove this

static/js/public/fullRecord.js
```diff
@@ -156,6 +182,19 @@ define('fullRecord', ['module', 'jquery', 'JP2Viewer', 'StructureView', 'dataTab
 			return "";
 		}
 
+		function showBadge(data) {
+			var markedForDeletion = false;
+			var restrictedAccess = false;
+
+			if (data.status !== undefined) {
+				var restrictions = data.status.join(',').toLowerCase();
```
bbpennel: `status` is an array of known values, feels a little weird/roundabout to transform it into a string for searching

static/js/vue-cdr-access/src/components/browseDisplay.vue
```diff
@@ -4,9 +4,25 @@
...
                             <i v-else class="fa" :class="recordType(record.type)"></i>
+                            <div v-if="markedForDeletion(record)" class="thumbnail-badge-trash"
+                                 :class="{'deleted-image-icon': thumbnailPresent(record.thumbnail_url)}">
+                                <div class="fa-stack">
+                                    <i class="fa fa-circle fa-stack-2x background"></i>
+                                    <i class="fa fa-trash fa-stack-1x foreground"></i>
+                                </div>
+                            </div>
+                            <div v-else-if="isRestricted(record)" class="thumbnail-badge-lock"
+                                 :class="{'deleted-image-icon': thumbnailPresent(record.thumbnail_url)}">
```
bbpennel: is this the right class for something that is restricted rather than deleted?

static/js/vue-cdr-access/src/components/listDisplay.vue
```diff
@@ -5,14 +5,31 @@
...
+                                <div v-if="markedForDeletion(record)" class="thumbnail-badge-trash"
+                                     :class="{'deleted-image-icon': thumbnailPresent(record.thumbnail_url),
+                                     'thumbnail-badge-trash-search ': !isRecordBrowse}">
+                                    <div class="fa-stack">
+                                        <i class="fa fa-circle fa-stack-2x background"></i>
+                                        <i class="fa fa-trash fa-stack-1x foreground"></i>
+                                    </div>
+                                </div>
+                                <div v-else-if="isRestricted(record)" class="thumbnail-badge-lock"
+                                     :class="{'deleted-image-icon': thumbnailPresent(record.thumbnail_url),
```
bbpennel: same question as in the previous file

access/src/main/webapp/WEB-INF/jsp/fullRecord/neighborList.jsp
```diff
@@ -29,7 +29,12 @@
 			<c:set var="currentItemClass" scope="page">
 				<c:if test="${briefObject.id == neighbor.id}"> current_item</c:if>
 			</c:set>
-			<div class="relateditem ${currentItemClass}">
+			<c:choose>
+				<c:when test="${markedForDeletion}">
```
bbpennel: Wouldn't this mark all the neighbors as deleted if the current record is deleted, even if they aren't deleted?

static/js/vue-cdr-access/src/mixins/displayUtils.js
```diff
@@ -22,6 +22,22 @@ export default {
...
+        isRestricted(record) {
+            return /embargoed|staff-only/.test(this.restrictions(record));
+        },
+
+        restrictions(record) {
+            if (record.status === undefined) {
+                return '';
+            }
+
+            return record.status.join(',').toLowerCase();
```
bbpennel: guess I have the same question here about searching as an array

static/js/public/fullRecord.js
lfarrell: Yeah, I was using $.array, but it didn't seem to work. I can double check it.

lfarrell: Hmm, it does work with $.inArray, but I'm inclined to leave it as is, because with inArray you'd have to test for, I think three conditions for restricted status: Embargoed, Staff-Only Access and Parent Has Staff-Only Access.

static/js/vue-cdr-access/src/mixins/displayUtils.js
lfarrell: See, note above

access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
```diff
@@ -20,9 +20,10 @@
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
 <%@ taglib prefix="cdr" uri="http://cdr.lib.unc.edu/cdrUI" %>
 
-<c:if test="${not empty briefObject && cdr:contains(briefObject.status, 'Deleted') || cdr:contains(briefObject.status, 'Parent Deleted')}">
-	<c:set var="isDeleted" value="deleted" scope="page"/>
+<c:if test="${markedForDeletion}">
+	<c:set var="isDeleted" value="deleted" scope="session"/>
```
bbpennel: i think "session" will stick around past the current request. If you need a broader scope than "page", i think you can use "request" to avoid some potentially weird interference

access/src/main/webapp/WEB-INF/jsp/fullRecord/neighborList.jsp
```diff
@@ -30,7 +30,7 @@
 				<c:if test="${briefObject.id == neighbor.id}"> current_item</c:if>
 			</c:set>
 			<div class="relateditem ${currentItemClass}">
-				<div class="relatedthumb">
+				<div class="relatedthumb ${isDeleted} ${currentItemClass}">
```
bbpennel: Why is `currentItemClass` being added in here now? It doesn't look like anything has changed with the class. Was there a bug?

static/js/public/fullRecord.js
bbpennel: okay, we'll just have to be careful to make sure the regexes can't match across statuses in the future, since they aren't paying attention to the commas at present.

access/src/main/webapp/WEB-INF/jsp/fullRecord/neighborList.jsp
lfarrell: Nope, got copy/pasted

access/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
```diff
@@ -45,7 +45,21 @@
     <div id="is-admin-unit" class="columns browse-header">
         <div class="column is-12">
             <c:import url="fullRecord/navigationBar.jsp" />
-            <h2><i class="fa fa-university" aria-hidden="true"></i> <c:out value="${briefObject.title}" /></h2>
+            	<c:choose>
```
bbpennel, line 48: can't you use the `isDeleted` variable here? I don't think having any empty `class` attribute is against any rules

access/src/main/webapp/WEB-INF/jsp/fullRecord/neighborList.jsp
```diff
@@ -29,8 +29,18 @@
 			<c:set var="currentItemClass" scope="page">
 				<c:if test="${briefObject.id == neighbor.id}"> current_item</c:if>
 			</c:set>
+			<c:set var="markedForDeletion"  value="${cdr:contains(neighbor.status, 'Marked For Deletion')}" scope="request"/>
```
bbpennel, line 32: i think you can use the default page scope for this, you don't need the variable past this point

access/src/main/webapp/WEB-INF/jsp/fullRecord/neighborList.jsp
```diff
@@ -29,8 +29,18 @@
...
+			<c:choose>
+				<c:when test="${markedForDeletion}">
+					<c:set var="isDeleted" value="deleted" scope="request"/>
```
bbpennel, line 35: this is overriding the other `isDeleted` value, I'd suggest using a different variable name, like `neighborIsDeleted` so we don't end up with unintended side effects. Also you can use the page scope for this variable.

access/src/main/webapp/WEB-INF/jsp/fullRecord.jsp
```diff
@@ -20,14 +20,26 @@
...
+<c:if test="${markedForDeletion}">
+	<c:set var="isDeleted" value="deleted" scope="request"/>
 </c:if>
+
 <c:set var="allowsPublicAccess" value="${permsHelper.allowsPublicAccess(briefObject)}" />
 <c:if test="${not empty briefObject && (not allowsPublicAccess || not empty briefObject.activeEmbargo)}">
 	<c:set var="isProtected" value="protected" scope="page"/>
 </c:if>
 
+<c:set var="badgeIcon" scope="request">
```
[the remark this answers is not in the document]
lfarrell: Unfortunately, yes. It doesn't show up otherwise for things like the collection pages.

access/src/main/webapp/WEB-INF/jsp/fullRecord/neighborList.jsp
lfarrell: Changing the scope here causes the current record's icon to show for all records

bbpennel: Changing it to `page` does? Weird, my understanding is `page` is for the current file, while `request` is for all pages in the request:
https://javapapers.com/jsp/explain-the-scope-of-jsp-objects/
It would likely be safest to change the name of the variable to something that doesn't overlap with the `markedForDeletion` variable from the controller, like `neighborMarkedForDeletion`.

lfarrell: Seems like it needs the current setup so the current neighbor can override the controller set icon, changing the variable name results in the issue of all related items getting the same icon, even if they shouldn't. I'm guessing it needs to stay "request" so thumbnail.jsp gets the right value.

bbpennel: Hmm, yeah i guess that makes sense since thumbnail.jsp is receiving this variable from the request rather than as a parameter passed to it explicitly, or a specially named variable just for the thumbnail component. Hopefully we won't have any issues where the state of the last neighbor changes content for the record the neighbor list is embedded in. Luckily its basically the last part of the page right now, so that reduces the chance of this issue. So we can go with this, and switch to passing in a parameter if we end up moving things around later.

PR #944, opened 2020-03-27
Check if download is a work. If so, use its primary object's PID

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2519

### DIFF
--- a/access/src/main/java/edu/unc/lib/dl/ui/controller/FedoraContentController.java
+++ b/access/src/main/java/edu/unc/lib/dl/ui/controller/FedoraContentController.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import edu.unc.lib.dl.fcrepo4.FileObject;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


access/src/main/java/edu/unc/lib/dl/ui/controller/FedoraContentController.java
```diff
@@ -87,6 +92,11 @@ private void streamData(String pidString, String datastream, boolean asAttachmen
         PID pid = PIDs.get(pidString);
         AccessGroupSet principals = getAgentPrincipals().getPrincipals();
 
+        RepositoryObject updatedPid = repoObjLoader.getRepositoryObject(pid);
+        if (updatedPid.getClass() == WorkObject.class) {
```
bbpennel: use `instanceof`

access/src/main/java/edu/unc/lib/dl/ui/controller/FedoraContentController.java
```diff
@@ -87,6 +92,11 @@ private void streamData(String pidString, String datastream, boolean asAttachmen
         PID pid = PIDs.get(pidString);
         AccessGroupSet principals = getAgentPrincipals().getPrincipals();
 
+        RepositoryObject updatedPid = repoObjLoader.getRepositoryObject(pid);
```
bbpennel: since this isn't really a pid object, i'd suggest a different variable name

access/src/main/java/edu/unc/lib/dl/ui/controller/FedoraContentController.java
```diff
@@ -87,6 +92,11 @@ private void streamData(String pidString, String datastream, boolean asAttachmen
...
+        if (updatedPid.getClass() == WorkObject.class) {
+            pid = ((WorkObject) updatedPid).getPrimaryObject().getPid();
```
bbpennel: not all works are guaranteed to have a primary object, so you should check that its null before getting the pid

PR #945, opened 2020-03-30
Change editLabel to edit Filename

lfarrell:
* Change editLabel to edit Filename
* Update Premis event and tests

https://jira.lib.unc.edu/browse/BXC-2616


persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditFilenameService.java
```diff
@@ -67,17 +69,21 @@ public void editLabel(AgentPrincipals agent, PID pid, String label) {
...
 
             RepositoryObject obj = repoObjLoader.getRepositoryObject(pid);
 
+            if (!(obj instanceof FileObject)) {
+                throw new Exception("Failed to edit filename for " + obj.getPid());
+            }
+
             String oldLabel = getOldLabel(obj);
 
             repoObjFactory.createExclusiveRelationship(obj, DcElements.title, label);
```
bbpennel: Its occurring to me now that we should change the filename instead of the label. Could you change this to setting the ebucore:filename? We retain the original filename in the PREMIS, so that should be okay.

static/templates/admin/editFilenameForm.html
```diff
@@ -1,9 +1,8 @@
 <iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
-<form id="edit_filename_form" method="post" action="editlabel/<%= pid %>" enctype="multipart/form-data" target="upload_file_container" class="edit_form edit_filename_form">
-	
+<form id="edit_filename_form" class="edit_form edit_filename_form">
 	<div class="form_field">
 		<label>Filename</label>
-		<input name="label" size="40" value="<%= metadata.label %>"></input>
```
bbpennel: Looking through the solr indexing code, it appears we aren't populating `label` anymore, so removing that was a good thing, and perhaps we should make a followup to remove the field.

Unfortunately the title and the filename for a file object are not necessarily going to be the same... The actual filename is stored in the solr datastream field, for example:
`original_file|image/jpeg|IMG_20200324_174901.jpg|jpg|3890854|urn:md5:788a541160b26039299647c9005b0767|`
it is the 3rd field in there. Perhaps pull the value from there? I believe this field will get updated after edit filename completes, but you might want to verify that.

bbpennel:
Also travis failed https://travis-ci.org/github/UNC-Libraries/Carolina-Digital-Repository/jobs/668890259

persistence/src/main/java/edu/unc/lib/dl/persist/services/edit/EditFilenameService.java
```diff
@@ -67,17 +68,21 @@ public void editLabel(AgentPrincipals agent, PID pid, String label) {
...
             RepositoryObject obj = repoObjLoader.getRepositoryObject(pid);
 
+            if (!(obj instanceof FileObject)) {
+                throw new IllegalArgumentException("Failed to edit filename for " + obj.getPid());
+            }
+
             String oldLabel = getOldLabel(obj);
 
-            repoObjFactory.createExclusiveRelationship(obj, DcElements.title, label);
+            repoObjFactory.createExclusiveRelationship(obj, Ebucore.filename, label);
```
bbpennel, line 82: actually I'm realizing that that this relationship needs to be set on the binary rather than the FileObject. You can get the binary by calling `fileObj.getOriginalFile()`. You'll also need to get the old label from the same place. Sorry I didn't notice that sooner, I haven't gotten to manually test yet.

static/js/admin/src/AbstractForm.js
```diff
@@ -8,6 +8,11 @@ define('AbstractForm', [ 'jquery', 'jquery-ui', 'underscore', 'ModalCreate',
 
 		AbstractForm.prototype.open = function(resultObject) {
 			var dialogBox = new ModalCreate(this.options);
+
+			if (this.options.showCurrentFilename) {
```
bbpennel, line 12: Hmm, I'd rather avoid adding the new filename related code into this class. You should be able to work this functionality into `EditFilenameForm.js`, maybe override the `open` method to set the filename and then call the parent method.

persistence/src/test/java/edu/unc/lib/dl/persist/services/edit/EditFilenameServiceTest.java
```diff
@@ -154,6 +151,7 @@ public void editFilenameTest() {
         service.editLabel(agent, pid, label);
 
         verify(repoObjFactory).createExclusiveRelationship(eq(binaryObj), eq(Ebucore.filename), any(Resource.class));
```
bbpennel, line 153: You shouldn't need to verify this method was called twice in a row, so you should be able to get rid of this line. Also I don't think you're verify the value before it is getting overwritten by line 156 before you verify it. Have you tried just doing `verify(repoObjFactory).createExclusiveRelationship(eq(binaryObj), eq(Ebucore.filename), eq(label))` ? If that doesn't work, then you'll need another assertEquals for the label right after this capture.

You'll want to make the same changes in the other test.

lfarrell: eq(label) seems to work fine. Updated

PR #946, opened 2020-03-30
Change marked for deletion premis event

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2619


persistence/src/main/java/edu/unc/lib/dl/persist/services/delete/RestoreDeletedJob.java
```diff
@@ -77,7 +77,7 @@ public void run() {
 
         sparqlUpdateService.executeUpdate(repoObj.getMetadataUri().toString(), updateString);
 
-        repoObj.getPremisLog().buildEvent(Premis.Deletion)
+        repoObj.getPremisLog().buildEvent(Premis.Deaccession)
```
bbpennel: Looking at the spreadsheet, it actually looks like this should be an Accession event. I'm not sure if that's already listed in the `Premis` class, if its not, the URI is `http://id.loc.gov/vocabulary/preservation/eventType/acc`

PR #953, opened 2020-04-03
Add message to confirm that user wants to update MODS record without adding a title

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2625

### DIFF
--- a/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilter.java
+++ b/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilter.java
@@ -97,23 +97,18 @@ public class SetDescriptiveMetadataFilter implements IndexDocumentFilter {
     }
 
     private String getAlternativeTitle(DocumentIndexingPackage dip) throws FedoraException, IndexingException {
-        String title = null;
         Resource resc = dip.getContentObject().getResource();
 
         // Use dc:title as a default
         if (resc.hasProperty(DcElements.title)) {
             Statement dcTitle = resc.getProperty(DcElements.title);
             return dcTitle.getString();
-        }
-
-        // fall back to filename if one is present
-        if (title == null && resc.hasProperty(Ebucore.filename)) {
+        } else if (resc.hasProperty(Ebucore.filename)) { // fall back to filename if one is present
             Statement filename = resc.getProperty(Ebucore.filename);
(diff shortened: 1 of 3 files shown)


solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -97,23 +97,18 @@ public void filter(DocumentIndexingPackage dip) throws IndexingException {
...
 
         // Use dc:title as a default
         if (resc.hasProperty(DcElements.title)) {
             Statement dcTitle = resc.getProperty(DcElements.title);
             return dcTitle.getString();
-        }
-
-        // fall back to filename if one is present
-        if (title == null && resc.hasProperty(Ebucore.filename)) {
+        } else if (resc.hasProperty(Ebucore.filename)) { // fall back to filename if one is present
```
bbpennel, line 106: Should we do empty checks before returning these values? `StringUtils.isBlank` catches null, blank, or all whitespace. If the title is present but empty, we'd want to fall back to the other options

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -131,7 +126,11 @@ private void extractTitles(Element mods, IndexDocumentBean idb) {
                 }
             }
         }
-        idb.setTitle(mainTitle);
+
+        if (mainTitle != null && !mainTitle.equals("")) {
```
bbpennel, line 130: you can use StringUtils.isBlank in place of these two checks (the apache commons lang3 version)

static/js/xmleditor/jquery.xmleditor.js
```diff
@@ -905,6 +905,18 @@ $.widget( "xml.xmlEditor", {
 		}
 		// convert XML DOM to string
 		var xmlString = this.xml2Str(this.xmlState.xml);
+
+		if (/mods:title.*?\/>/.test(xmlString)) {
+			if (window.confirm("The current MODS document has no title. " +
+				"This will result in no title displaying for the record. Would you like to save anyway?")) {
```
bbpennel, line 911: This statement probably isn't accurate anymore. Maybe we can remove this warning, assuming the indexing part works correctly?

PR #955, opened 2020-04-07
Add collection thumbnails

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2518

Depends on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/47
Probably need to add/merge https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/48 too to get the VM to provision


services/src/main/java/edu/unc/lib/dl/cdr/services/processing/ImportThumbnailService.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public void run(InputStream importStream, AgentPrincipals agent, String uuid) throws Exception {
+        PID pid = PIDs.get(uuid);
+
+        aclService.assertHasAccess("User does not have permission to add/update collection thumbnails",
+                pid, agent.getPrincipals(), Permission.createCollection);
+
+        try {
+            Tika tika = new Tika();
+            String mimeType = tika.detect(importStream);
```
bbpennel, line 73: One thing I'm not sure about with this approach, is the importStream getting consumed? When you write the importStream out to file, is it writing anything? Normally inputstreams can only be read once, but some types of input streams can be reset and read multiple times. I don't know if that is happening here or not.

I guess if you used the user supplied mimetype from the upload request it might not be an issue. But if necessary you can also duplicate the inputstream, similiar to what is happening in the edit description service.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/ImportThumbnailService.java
```diff
@@ -0,0 +1,124 @@
...
+
+        // Create the directory if it doesn't already exist
+        Files.createDirectories(storagePath);
+    }
+
+    public void run(InputStream importStream, AgentPrincipals agent, String uuid) throws Exception {
+        PID pid = PIDs.get(uuid);
+
+        aclService.assertHasAccess("User does not have permission to add/update collection thumbnails",
+                pid, agent.getPrincipals(), Permission.createCollection);
```
bbpennel, line 69: I'd like to avoid locking this workflow down to collections, since I think we'll want to be able to use it for admin units too, and maybe even folders. `EditDescription` permission might be reasonable for this.

More generally, it'd be good to go with naming like "EditThumbnail" rather than including "collection" in the names of stuff.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/ImportThumbnailService.java
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            Tika tika = new Tika();
+            String mimeType = tika.detect(importStream);
+
+            if (!containsIgnoreCase(mimeType, "image")) {
+                throw new IllegalArgumentException("Uploaded file is not an image");
+            }
+
+            String thumbnailBasePath = idToPath(uuid, HASHED_PATH_DEPTH, HASHED_PATH_SIZE);
+            String filePath = Paths.get(storagePath.toString(), thumbnailBasePath, uuid).toString();
```
bbpennel, line 80: Another option is to do `storagePath.resolve(thumbnailBasePath).resolve(uuid)`, might not make much difference but i guess if you'd rather not have to convert to a string

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/ImportThumbnailService.java
```diff
@@ -0,0 +1,124 @@
...
+
+            log.info("Job to to add thumbnail to collection {} has been queued by {}",
+                    uuid, agent.getUsername());
+        } catch (IllegalArgumentException e) {
+            log.error("Uploaded file for collection {} is not an image file", uuid);
+            throw e;
+        }
+    }
+
+    private Document createEnhancementMsg(String userid, String filePath, String mimeType) {
```
bbpennel, line 95: If we can get this to just be a regular RunEnhancements call, then you should be able to use the helper method for generating enhancement messages.

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/ImportThumbnailService.java
lfarrell: This seems to have been the issue with the thumbnail generation. Switching to the uploaded mime-type seems to have resolved it.

services-camel/src/main/java/edu/unc/lib/dl/services/camel/BinaryEnhancementProcessor.java
```diff
@@ -52,23 +60,35 @@ public void process(final Exchange exchange) throws Exception {
...
             String pidValue = body.getChild("pid", ATOM_NS).getTextTrim();
-            Element isEditThumbnail = body.getChild("editThumbnail", ATOM_NS);
 
-            if (isEditThumbnail != null) {
-                String mimeType = body.getChild("mimeType", ATOM_NS).getTextTrim();
+            String[] collThumbPath = pidValue.split("/");
+            String uuid = collThumbPath[collThumbPath.length - 1];
+            PID objPid = PIDs.get(uuid);
+
+            if (repoObjLoader.getCollectionObject(objPid) != null) {
```
bbpennel, line 68: I think this will throw an error if the object pid isn't a collection. You would want to use `getRepositoryObject` and then check the type of the result. Also, if these methods don't find an object, they throw an exception rather than returning null.

However, I don't think i'd recommend the line `in.setHeader(FCREPO_RESOURCE_TYPE, Binary.getURI());` below since it's effectively declaring that a collection/folder/work/whatever is a binary for the sake of getting through the router.

My suggestion would be to populate `FCREPO_RESOURCE_TYPE` with the comma separated join of `repoObj.getTypes()`. Then add a new route here:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/65d5e3c120736859fb9c397d6b9ad4151213232d/services-camel/src/main/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouter.java#L71

for processing non-binaries. In that route, there can be a new processor that does most of what you have in this `if` statement. This way, we don't need to give an inaccurate resource type, and it clearly separates the behavior for binaries vs others where they diverge. We also shouldn't need the `CdrEditThumbnail` header in this case. I think all the changes to BinaryEnhancementProcessor could be backed out except for the new stuff to set the resource types header. 

The non-binary route would be similar to the binary one. After it goes through the new processor (Maybe NonBinaryEnhancementProcessor), if enhancements are needed it should have CdrBinaryPath set, then you can set CdrEnhancementSet and call `"direct:process.enhancements", "direct-vm:solrIndexing"`. Otherwise, you'd just want to call solrIndexing.

A few other things:
1. the run enhancement message pid file should have a value you can just pass into `PIDs.get` without any extra splitting.
2. `thumbnailBasePath` should actually be the path for where the source image is, not where thumbnails would be. This should be updated in the ImportThumbnailService as well, and we should double check that it is using the correct directory.
3. Only set CdrBinaryPath for non-binaries if a file exists at the generated source image path. That way the router can decide whether to perform binary specific enhancements on if there is a binary.

services/src/main/java/edu/unc/lib/dl/cdr/services/rest/EditThumbnailController.java
```diff
@@ -0,0 +1,80 @@
...
+        try (InputStream importStream = thumbnailFile.getInputStream()) {
+            service.run(importStream, agent, pid, mimeType);
+        } catch (IOException e) {
+            log.error("Failed to get submitted file", e);
+            result.put("error", e.getMessage());
+            return new ResponseEntity<>(result, HttpStatus.INTERNAL_SERVER_ERROR);
+        } catch (IllegalArgumentException e) {
+            log.error("Error queueing the job", e);
+            result.put("error", e.getMessage());
+            return new ResponseEntity<>(result, HttpStatus.INTERNAL_SERVER_ERROR);
```
bbpennel, line 73: What kind of operation would throw IllegalArgumentException? I'd usually expect a 4xx for bad input.

services/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -311,6 +311,13 @@
       <property name="modsValidator" ref="modsValidator" />
       <property name="versionedDatastreamService" ref="versionedDatastreamService" />
   </bean>
+
+    <bean id="editThumbnailService" class="edu.unc.lib.dl.cdr.services.processing.ImportThumbnailService"
+          init-method="init">
+        <property name="aclService" ref="aclService" />
+        <property name="messageSender" ref="runEnhancementMessageSender" />
+        <property name="dataDir" value="${data.editThumb.dir}" />
```
bbpennel, line 319: would need to update this property name too

services/src/test/resources/add-thumb-it-servlet.xml
```diff
@@ -0,0 +1,52 @@
...
+        <property name="messageSender" ref="messageSender" />
+        <property name="dataDir" value="dataDir" />
+    </bean>
+
+    <bean id="aclServices" class="org.mockito.Mockito" factory-method="mock">
+        <constructor-arg value="edu.unc.lib.dl.acl.fcrepo4.AccessControlServiceImpl"/>
+    </bean>
+
+    <bean id="dataDir" class="java.lang.String">
+        <constructor-arg value="src/test/resources/temp"/>
```
bbpennel, line 50: use the `target` directory for temp space, its used for throw away and build stuff that shouldn't get checked into git.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDatastreamFilter.java
```diff
@@ -60,22 +62,27 @@ public void filter(DocumentIndexingPackage dip) throws IndexingException {
...
+            // Retrieve list of derivatives associated with the object
+            List<Datastream> derivatives = getDerivatives(fileObj.getPid(), ownedByOtherObject);
+            datastreams.addAll(derivatives);
+
+            IndexDocumentBean doc = dip.getDocument();
+
+            doc.setDatastream(getDatastreamStrings(datastreams));
+            doc.setFilesizeTotal(getFilesizeTotal(datastreams));
+            doc.setFilesizeSort(getFilesize(datastreams));
+        } else if (contentObj instanceof CollectionObject) {
```
bbpennel, line 78: indexing derivatives is sufficiently generic that i think you can remove the collection object restriction

services-camel/src/test/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouterIT.java
```diff
@@ -194,6 +212,26 @@ public void testImageFile() throws Exception {
         verify(solrIngestProcessor).process(any(Exchange.class));
     }
 
+   /* @Test
```
bbpennel, line 215: Cant this be revived as a non-binary with source image test?

services-camel/src/test/java/edu/unc/lib/dl/services/camel/BinaryEnhancementProcessorTest.java
```diff
@@ -101,11 +116,23 @@ public void testExistingUriHeader() throws Exception {
         verify(message, never()).setHeader(FCREPO_RESOURCE_TYPE, Binary.getURI());
     }
 
+    @Test
+    public void testEditThumbnail() throws Exception {
```
bbpennel, line 120: I think this test is out of date? It should probably just be "testNonBinary". I'm not sure the `setMessageBody("image/*");` bit is doing anything either?

lfarrell: Looks like it needs the message body, otherwise the processor will give a NPE

persistence/src/main/java/edu/unc/lib/dl/services/RunEnhancementsMessageHelpers.java
```diff
@@ -47,12 +47,20 @@ public static Document makeEnhancementOperationBody(String userid, PID pid, Bool
...
+        Element paramForce = new Element("force", CDR_MESSAGE_NS);
 
         if (force) {
-            Element paramForce = new Element("force", CDR_MESSAGE_NS);
             paramForce.setText("true");
+        } else {
+            paramForce.setText("false");
         }
+
+        Element enhancements = new Element(ENHANCEMENTS.getName(), ATOM_NS);
```
bbpennel, line 59: use the cdr message namespace for the enhancement element and the pid element

metadata/src/main/java/edu/unc/lib/dl/util/JMSMessageUtil.java
```diff
@@ -110,6 +110,7 @@ public static FedoraActions getAction(String value) {
         ADD("add"),
         EDIT_ACCESS_CONTROL("editAccess"),
         EDIT_TYPE("editType"),
+        ENHANCEMENTS("enhancements"),
```
bbpennel, line 113: good idea to add a constant. My only suggestion is to make it a verb/action since all the other ones are

services-camel/src/main/java/edu/unc/lib/dl/services/camel/BinaryEnhancementProcessor.java
```diff
@@ -47,11 +53,28 @@ public void process(final Exchange exchange) throws Exception {
             Document msgBody = MessageUtil.getDocumentBody(in);
             Element body = msgBody.getRootElement();
 
-            String pidValue = body.getChild("pid", ATOM_NS).getTextTrim();
+            Element hasEnhancementPid = body.getChild("enhancements", ATOM_NS);
```
bbpennel, line 56: has usually implies a boolean and isn't really what this is storing, maybe just enhancementEl or something like that?

bbpennel: also, use the constant from JMSElementUtil for the name of this element

services-camel/src/test/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouterIT.java
```diff
@@ -162,14 +179,32 @@ public void testFolderEnhancements() throws Exception {
...
+
+        NotifyBuilder notify1 = new NotifyBuilder(cdrEnhancements)
+                .whenCompleted(1)
+                .create();
+
+        boolean result1 = notify1.matches(5l, TimeUnit.SECONDS);
+        assertTrue("Enhancement route not satisfied", result1);
+
+        verify(addSmallThumbnailProcessor, never()).process(any(Exchange.class));
+        verify(addLargeThumbnailProcessor, never()).process(any(Exchange.class));
```
bbpennel, line 196: Should check `verify(solrIngestProcessor).process(any(Exchange.class));`  too

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetDatastreamFilter.java
```diff
@@ -60,22 +60,20 @@ public void filter(DocumentIndexingPackage dip) throws IndexingException {
         ContentObject contentObj = dip.getContentObject();
 
         FileObject fileObj = getFileObject(contentObj);
-        if (fileObj == null) {
-            return;
+        if (fileObj != null) {
```
bbpennel, line 63: No datastreams are going to get indexed for objects other than works and files, i think? Did you need to get rid of the other side of this conditional? We still need to index other objects, we just don't need to restrict it to CollectionObjects, which is what you had before.

lfarrell: So are we back to this?
`   if (fileObj != null) {
            boolean ownedByOtherObject = contentObj instanceof WorkObject;
            // Retrieve list of datastreams associated with this object
            List<Datastream> datastreams = getDatastreams(fileObj, ownedByOtherObject);
            // Retrieve list of derivatives associated with the object
            List<Datastream> derivatives = getDerivatives(fileObj.getPid(), ownedByOtherObject);
            datastreams.addAll(derivatives);

            IndexDocumentBean doc = dip.getDocument();

            doc.setDatastream(getDatastreamStrings(datastreams));
            doc.setFilesizeTotal(getFilesizeTotal(datastreams));
            doc.setFilesizeSort(getFilesize(datastreams));
        } else {
            List<Datastream> derivatives = getDerivatives(contentObj.getPid(), false);
            List<Datastream> datastreams = new ArrayList<>(derivatives);

            IndexDocumentBean doc = dip.getDocument();

            doc.setDatastream(getDatastreamStrings(datastreams));
        }`

bbpennel: yes almost, my last comment was that in the second block, `derivatives` and `datastreams` have the exact same stuff in them, so there's no reason to make the second variable. We do need the `else` block which covers objects besides works and files.

services-camel/src/main/java/edu/unc/lib/dl/services/camel/BinaryEnhancementProcessor.java
```diff
@@ -53,9 +53,9 @@ public void process(final Exchange exchange) throws Exception {
             Document msgBody = MessageUtil.getDocumentBody(in);
             Element body = msgBody.getRootElement();
 
-            Element hasEnhancementPid = body.getChild("enhancements", ATOM_NS);
-            if (hasEnhancementPid != null) {
-                String pidValue = hasEnhancementPid.getChild("pid", ATOM_NS).getTextTrim();
+            Element enhancementsEl = body.getChild("runEnhancements", CDR_MESSAGE_NS);
```
bbpennel, line 56: use the `JMSMessageUtil.CDRActions.RUN_ENHANCEMENTS` constant here

services-camel/src/test/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouterIT.java
bbpennel: The solr ingest processor should run in this case, if its not then that would be a bug

services-camel/src/test/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouterIT.java
```diff
@@ -137,11 +154,11 @@ public void init() throws Exception {
     }
 
     @Test
-    public void testFolderEnhancements() throws Exception {
+    public void nonBinaryWithSourceImages() throws Exception {
```
bbpennel, line 157: This test should verify that thumbnail enhancements are being called. You could also check that the access copy is not called.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/enhancements/EnhancementRouterIT.java
```diff
@@ -268,13 +304,20 @@ public void testProcessFilterOutDescriptiveMDSolr() throws Exception {
...
 
-    private static Map<String, Object> createEvent(PID pid, String... type) {
-
+    private Map<String, Object> createEvent(boolean editThumb, PID pid, String... type) {
         final Map<String, Object> headers = new HashMap<>();
         headers.put(IDENTIFIER, pid.getRepositoryPath());
         headers.put(EVENT_TYPE, "ResourceCreation");
         headers.put("CamelFcrepoUri", pid.getRepositoryPath());
         headers.put(RESOURCE_TYPE, String.join(",", type));
+        headers.put(CdrBinaryMimeType, "text/plain");
```
bbpennel, line 313: I think this and the `if (editThumb)` should be removed, since the router and its processes should be setting these headers. You should also be able to remove the `boolean editThumb` parameter from this method

PR #959, opened 2020-04-23
Display MODs data

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2592


access/src/main/webapp/WEB-INF/jsp/fullRecord/aggregateRecord.jsp
```diff
@@ -174,5 +174,6 @@
 		</table>
 	</div>
 </c:if>
-
+<h2 class="full-metadata">Detailed Metadata</h2>
```
bbpennel, line 177: The inclusion of this section should be conditional on there being a MODS datastream. You might need to modify `PermissionsHelper` to add a `hasDescriptionAccess` method like the other ones in that file.

static/js/public/fullRecord.js
```diff
@@ -73,6 +74,24 @@ define('fullRecord', ['module', 'jquery', 'JP2Viewer', 'StructureView', 'dataTab
 		});
 	}
 
+	if ($modsDisplay.length > 0) {
+		$.ajax({
+			url: "/record/" + $modsDisplay.attr("data-pid") + "/metadataView",
+			dataType: "html",
+			success: function(data) {
+				if (/^no.metadata/i.test($.trim(data))) {
```
bbpennel, line 82: Might make more sense to return a response status of 404, but if it doesn't seem worth while at this time, we can live with it.

access-common/src/test/java/edu/unc/lib/dl/ui/util/ModsUtilTest.java
```diff
@@ -0,0 +1,250 @@
...
+                "    <title>  \n  \n</title>" +
+                "  </titleInfo>" +
+                "</mods>";
+
+        Document modsDoc = buildMods(modsString);
+        assertEquals(0, modsDoc.getRootElement().getChildren().size());
+    }
+
+    @Test
+    public void emptyDeepNested() throws JDOMException, IOException {
```
bbpennel, line 215: maybe rename this to "emptyNestedWithAttributes"?

access-common/src/test/java/edu/unc/lib/dl/ui/util/ModsUtilTest.java
```diff
@@ -0,0 +1,250 @@
...
+import org.jdom2.Document;
+import org.jdom2.Element;
+import org.jdom2.JDOMException;
+import org.jdom2.input.SAXBuilder;
+import org.junit.Test;
+
+/**
+ * @author lfarrell
+ */
+public class ModsUtilTest {
```
bbpennel, line 36: One more case to test, make sure that if all elements are populated, nothing gets removed.

PR #961, opened 2020-04-23
Style counts as badge

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2643

### DIFF
--- a/static/css/sass/cdr-ui.scss
+++ b/static/css/sass/cdr-ui.scss
@@ -659,10 +659,6 @@ p.spacing {
   width: 100%;
 }
 
-.record-count {
-
-}
-
 .iteminfo {
   float: none;
   margin-top: 5px;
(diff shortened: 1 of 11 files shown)


static/css/sass/cdr-ui.scss
```diff
@@ -647,18 +647,20 @@ p.spacing {
...
+  border-radius: 5px;
+  color: white;
+  line-height: 18px;
+  margin-bottom: 4px;
+  padding: 3px 8px;
+  white-space: nowrap;
+  width: 100%;
+}
+
+.record-count {
```
bbpennel, line 662: This rule doesn't appear to do anything

static/css/sass/cdr-ui.scss
```diff
@@ -1000,6 +1006,10 @@ p.no-search-results {
   }
 }
 
+.item_container_count {
```
bbpennel, line 1009: Hmm, this rule is also empty, is it needed?

lfarrell: Hmm, sorry thought I removed the empty rule. Maybe need to run static again to generate the updates

PR #967, opened 2020-04-28
BXC-2641/2642 - Tombstone updates

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2641
https://jira.lib.unc.edu/browse/BXC-2642

* Retain link from child to parent when transforming objects to tombstones
* Treat tombstones as if they were marked for deletion during ACL checks
* Tombstones are now ContentObjects, with indexing disabled
* Tombstone PREMIS should be accessible via API now


fcrepo-clients/src/main/java/edu/unc/lib/dl/acl/fcrepo4/ObjectAclFactory.java
```diff
@@ -200,7 +205,8 @@ public void setRepositoryObjectCacheLoader(RepositoryObjectCacheLoader repositor
             while (it.hasNext()) {
                 Statement stmt = it.next();
                 Property property = stmt.getPredicate();
-                if (CdrAcl.NS.equals(property.getNameSpace())) {
+                if (CdrAcl.NS.equals(property.getNameSpace())
+                        || RDF.type.equals(property) && Cdr.Tombstone.equals(stmt.getResource())) {
```
lfarrell, line 209: Does this need another set of () around it?

bbpennel: hmm, yeah that would be safest. In java `&&` takes precedence over `||` so it'd be evaluated first, which I guess is why this worked, but it'd be safer to group it

PR #969, opened 2020-04-29
Upgrade Vue.js CLI to 4.x

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2523


static/js/vue-cdr-access/src/components/browseSearch.vue
```diff
@@ -41,7 +41,12 @@
         methods: {
             getResults() {
                 let update_params = { anywhere: encodeURIComponent(this.search_query) };
-                this.$router.push({ name: 'displayRecords', query: this.urlParams(update_params) });
+                this.$router.push({ name: 'displayRecords', query: this.urlParams(update_params) })
+                    .catch((e) => {
+                        if (e.name !== 'NavigationDuplicated') {
```
bbpennel, line 46: what triggers NavigationDuplicated errors? Something like clicking twice on a link?

static/js/vue-cdr-access/tests/unit/modalMetadata.spec.js
```diff
@@ -9,7 +10,7 @@ localVue.use(VueRouter);
 
 const updated_uuid = 'c03a4bd7-25f4-4a6c-a68d-fedc4251b680';
 const title = 'Test Collection';
-const response = `<table><tbody><tr><th>Creator</th> <td><p>Real Dean</p></td></tr></tbody></table>`;
+const response = pretty(`<table><tbody><tr><th>Creator</th><td><p>Real Dean</p></td></tr></tbody></table>`);
```
bbpennel, line 13: Real dean is getting prettified!

Makefile
```diff
@@ -14,18 +14,13 @@ endif
...
 
 	# Add new line so app*.js doesn't get commented out
 	echo >> static/js/vue-permissions.js
 
-ifeq ($(DEPLOY_TYPE), prod)
-	cat static/js/admin/vue-permissions-editor/dist/js/app*.js >> static/js/vue-permissions.js
-else
-	cat static/js/admin/vue-permissions-editor/dist/app.js >> static/js/vue-permissions.js
-endif
+	cat static/js/admin/vue-permissions-editor/dist/js/app*js >> static/js/vue-permissions.js
```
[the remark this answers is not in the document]
lfarrell: Correct, the JS files should be the same now across deployment types. In the dev build the css still get bundled into the JS files.

static/js/vue-cdr-access/tests/unit/modalMetadata.spec.js
lfarrell: Yeah, vue test utils changed how they return HTML. It used to have no spaces, but they switched to formatting the HTML with the pretty package before returning it.

static/js/vue-cdr-access/src/components/browseSearch.vue
lfarrell: I think that or trying to redirect to the same url you're at. If I read the ticket mentioned in the commit correctly, it shouldn't give this error if you update parameters. It's possible that doesn't include parameters's specified via the ?. It might only refer to params in the route, like :uuid in our case.

lfarrell:
The big change was vue-test-utils stopped running actions within tests synchronously. Hence all the async awaits scattered around in the tests.

PR #975, opened 2020-05-01
Consider any admin ui work done if an error is returned

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2614


static/js/admin/src/RemoteStateChangeMonitor.js
```diff
@@ -52,6 +52,9 @@ define('RemoteStateChangeMonitor', ['jquery'], function($) {
 				clearInterval(this.pingId);
 				this.pingId = null;
 			}
+
+			// Consider any work done if an error is returned
+			this.options.statusChanged.call(this.options.statusChangedTarget, null);
```
bbpennel, line 57: Not 100% sure what all actions are going to do when confronted with a null, but this seems like a reasonable approach

PR #979, opened 2020-05-04
Allow images to get through access filter

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2657

### DIFF
--- a/access-common/src/main/java/edu/unc/lib/dl/ui/controller/LorisContentController.java
+++ b/access-common/src/main/java/edu/unc/lib/dl/ui/controller/LorisContentController.java
@@ -62,8 +62,8 @@ public class LorisContentController extends AbstractSolrSearchController {
      * @return
      */
     private boolean hasAccess(PID pid, String datastream) {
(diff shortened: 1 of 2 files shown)


access-common/src/main/java/edu/unc/lib/dl/ui/controller/LorisContentController.java
```diff
@@ -62,8 +62,8 @@
      * @return
      */
     private boolean hasAccess(PID pid, String datastream) {
-        // Defaults to jp2 surrogate if no datastream specified
-        if (datastream == null) {
+        // Defaults to jp2 surrogate
+        if (datastream.equals("IMAGE_JP2000") || datastream == null) {
```
bbpennel, line 66: hmm, IMAGE_JP2000 is the boxc3 name for the datastream. Might be better to update `JP2Viewer.js` to request datastreams `jp2` (I think that's the current datastream name).
Its kind of funny that `hasAccess` has this default datastream behavior but the actual controller methods both require the datastream name in the path.

PR #983, opened 2020-05-05
Only run enhancements if they don't already exist or are "forced"

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2607

### DIFF
--- a/services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextRouterTest.java
+++ b/services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextRouterTest.java
@@ -15,14 +15,20 @@
  */
 package edu.unc.lib.dl.services.camel.fulltext;
 
+import static edu.unc.lib.dl.fcrepo4.RepositoryPathConstants.HASHED_PATH_DEPTH;
+import static edu.unc.lib.dl.fcrepo4.RepositoryPathConstants.HASHED_PATH_SIZE;
+import static edu.unc.lib.dl.fcrepo4.RepositoryPaths.idToPath;
 import static edu.unc.lib.dl.services.camel.JmsHeaderConstants.EVENT_TYPE;
 import static edu.unc.lib.dl.services.camel.JmsHeaderConstants.IDENTIFIER;
 import static edu.unc.lib.dl.services.camel.JmsHeaderConstants.RESOURCE_TYPE;
 import static edu.unc.lib.dl.services.camel.util.CdrFcrepoHeaders.CdrBinaryMimeType;
 import static edu.unc.lib.dl.rdf.Fcrepo4Repository.Binary;
+import static org.fcrepo.camel.FcrepoHeaders.FCREPO_URI;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.verify;
 
+import java.io.File;
+import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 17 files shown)


services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextRouterTest.java
```diff
@@ -60,14 +74,60 @@ protected AbstractApplicationContext createApplicationContext() {
...
+        FileUtils.writeStringToFile(existingFile, "extracted text", "utf-8");
+
+        getMockEndpoint("mock:direct:fulltext.extraction").expectedMessageCount(1);
+        createContext(ENHANCEMENT_ROUTE);
 
+        Map<String, Object> headers = createEvent();
+        headers.put("force", "true");
+        template.sendBodyAndHeaders("", headers);
         assertMockEndpointsSatisfied();
+        existingFile.delete();
```
bbpennel, line 130: You can probably skip these delete steps since you have the `@After` method which recursively deletes the dir. If you didn't, this delete could be problematic since if the test failed it wouldn't delete the file.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/ImageEnhancementsRouterTest.java
```diff
@@ -123,40 +150,146 @@ public void testThumbDisallowedImageType() throws Exception {
...
+        createContext(smallThumbRoute);
+
+        getMockEndpoint("mock:exec:/bin/sh").expectedMessageCount(1);
+
+        Map<String, Object> headers = createEvent(fileID, eventTypes, "true");
+        template.sendBodyAndHeaders("", headers);
+
+        verify(addSmallThumbnailProcessor).process(any(Exchange.class));
+        assertMockEndpointsSatisfied();
+        existingFile.delete();
```
bbpennel, line 211: Can remove these in this class as well

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/ImageEnhancementsRouterTest.java
```diff
@@ -123,40 +150,146 @@ public void testThumbDisallowedImageType() throws Exception {
...
+    public void testThumbLargeRouteForceNoFileExists() throws Exception {
+        createContext(largeThumbRoute);
+
+        getMockEndpoint("mock:exec:/bin/sh").expectedMessageCount(1);
+
+        Map<String, Object> headers = createEvent(fileID, eventTypes, "true");
+
+        template.sendBodyAndHeaders("", headers);
+
+        doThrow(Exception.class).when(addLargeThumbnailProcessor).process(any(Exchange.class));
```
bbpennel, line 240: Is this throw serving a purpose here? Doesn't seem like what you'd expect in a successful case

services-camel/src/test/java/edu/unc/lib/dl/services/camel/images/ImageEnhancementsRouterTest.java
```diff
@@ -123,40 +150,146 @@ public void testThumbDisallowedImageType() throws Exception {
...
+        FileUtils.writeStringToFile(existingFile, "extracted text", "utf-8");
+        createContext(largeThumbRoute);
+
+        getMockEndpoint("mock:exec:/bin/sh").expectedMessageCount(0);
+
+        Map<String, Object> headers = createEvent(fileID, eventTypes, "false");
+
+        template.sendBodyAndHeaders("", headers);
+
+        doThrow(Exception.class).when(addLargeThumbnailProcessor).process(any(Exchange.class));
```
bbpennel, line 259: same question for this throw, plus you can probably remove the delete below

services/src/main/java/edu/unc/lib/dl/cdr/services/processing/RunEnhancementsService.java
```diff
@@ -110,11 +115,12 @@ public void setAclService(AccessControlService aclService) {
     private void createMessage(BriefObjectMetadata metadata, String username, Boolean force) {
         PID pid = metadata.getPid();
         Datastream originalDs = metadata.getDatastreamObject(ORIGINAL_FILE.getId());
-        if (originalDs == null) {
+        Datastream hasThumbnail = metadata.getDatastreamObject(THUMBNAIL_SMALL.getId());
```
bbpennel, line 118: Hmm, I would rather leave it up to the enhancement processors to decide if anything needs to be done to an object rather than do it here, which we would need to remember to update in the future if more types of derivatives are added for non-files. So I'd suggest removing this and the conditional at 119. The shortcut to jump to the original ds is probably fine to keep.

PR #987, opened 2020-05-11
Allow admins to pause/resume deposits

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2653

### DIFF
--- a/services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DepositController.java
+++ b/services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DepositController.java
@@ -250,8 +250,9 @@ public class DepositController {
         result.put("jobsURI", "/api/status/deposit/" + depositUUID + "/jobs");
         result.put("eventsURI", "/api/status/deposit/" + depositUUID + "/eventsXML");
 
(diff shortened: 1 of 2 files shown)


services/src/main/java/edu/unc/lib/dl/cdr/services/rest/DepositController.java
```diff
@@ -250,6 +250,9 @@ public void init() {
         result.put("jobsURI", "/api/status/deposit/" + depositUUID + "/jobs");
         result.put("eventsURI", "/api/status/deposit/" + depositUUID + "/eventsXML");
 
+        AccessGroupSet groups = getAgentPrincipals().getPrincipals();
+        result.put("isAdmin", groups.contains("unc:app:lib:cdr:admin"));
```
bbpennel, line 254: We should avoid hardcoding this.
You already have access to `GlobalPermissionEvaluator` in this controller, which has `hasGlobalPermission` and `getGlobalUserRoles`, either of which can probably be used for this purpose. With the latter, I think you could just check the result to see if it contains the Admin role, although it looks like you're already using `globalPermissionEvaluator.hasGlobalPermission(principals, Permission.createAdminUnit)` down below. Either is fine.

PR #988, opened 2020-05-12
Have page scroll to the top when user goes to a new route

lfarrell:
Have page scroll to the top when user goes to a new route. Most notably, clicking to the next page from the pagination component would leave you at the bottom of the page.


dstukes:
tested. works as expected.

PR #992, opened 2020-05-18
Upgrade leaflet and leaflet-IIIF

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2652
Depends on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/53


lfarrell:
No, I don't think we need the src files. Removed

PR #994, opened 2020-05-19
BXC-2537/2638 - Deposit migrated timestamps and deposit links

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2638
https://jira.lib.unc.edu/browse/BXC-2537

* Applies created/modified timestamps supplied in a deposit if the overrideTimestamps flag if provided
* Overrides the deposit record link on objects in an ingest if one was set in the deposit model
    * fixes a bug where the deposit id had the wrong pid type
* Fixes issue with longleaf registration which occurs when a deposit does not create a deposit record

### DIFF
--- a/deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
+++ b/deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
@@ -728,6 +728,60 @@ public class IngestContentObjectsJobIT extends AbstractFedoraDepositJobIT {
         assertTrue(logModel.contains(null, RDF.type, Premis.Ingestion));
     }
 
+    @Test
+    public void fromMultipleDepositsTest() throws Exception {
+
+        PID folderObj1Pid = pidMinter.mintContentPid();
+        PID folderObj2Pid = pidMinter.mintContentPid();
+        PID folderObj3Pid = pidMinter.mintContentPid();
+
+        PID deposit2Pid = pidMinter.mintDepositRecordPid();
+        PID deposit3Pid = pidMinter.mintDepositRecordPid();
+
+        // Create the deposit records since the references must resolve
+        repoObjFactory.createDepositRecord(deposit2Pid, null);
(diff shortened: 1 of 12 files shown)


deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -728,6 +734,135 @@ public void onlyIngestionEventAddedTest() throws Exception {
...
+        Bag collBag = model.createBag(collPid.getRepositoryPath());
+        collBag.addProperty(RDF.type, Cdr.Collection);
+        collBag.addLiteral(CdrDeposit.lastModifiedTime, LAST_MODIFIED_STRING);
+        collBag.addLiteral(CdrDeposit.createTime, CREATED_STRING);
+        unitBag.add(collBag);
+
+        PID folderPid = pidMinter.mintContentPid();
+        Bag folderBag = model.createBag(folderPid.getRepositoryPath());
+        folderBag.addProperty(RDF.type, Cdr.Folder);
+        folderBag.addLiteral(CdrDeposit.lastModifiedTime, LAST_MODIFIED_STRING);
```
lfarrell, line 824: does .addLiteral() overwrite the previous value, or is the last modified on the deposit not set by default?

bbpennel: it doesn't overwrite the previous value, unless its exactly the same value. The last modified is not normally set in the deposit pipeline because it is automatically generated by fedora. Fedora actually would normally throw an error if you make a PUT or a POST request with the created or lastModified timestamp, you have to put it into the special "relaxed" mode to allow this, which we will only have enabled when we need it (such as the migration). You can see me turning it on for this IT in this PR.

lfarrell: ah, so it's something you generally don't need to worry about.

PR #995, opened 2020-05-20
Add note about max file size and throw an error if file is larger than the max allowed

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2620


static/js/admin/src/AbstractFileUploadForm.js
```diff
@@ -24,20 +24,29 @@ define('AbstractFileUploadForm', [ 'jquery', 'jquery-ui', 'underscore', 'Abstrac
...
 		
-		$("input[type='file']", this.$form).change(function(){
+		$("input[type='file']", this.$form).change(function() {
+			errors = []; // reset errors on change
 			self.ingestFile = this.files[0];
 			if (self.ingestFile) {
+				var fileSize = self.ingestFile.size
+
+				if (fileSize > maxFileSize) {
+					errors.push("The uploaded file is too large. The maximum allowed file size is: 500mb");
```
bbpennel, line 43: this message doesn't match the size elsewhere

PR #996, opened 2020-05-26
Fix display of bulk action buttons from user actions

lfarrell:
* Fix display of bulk action buttons
* Remove publish/unpublish actions

https://jira.lib.unc.edu/browse/BXC-2672


static/js/admin/src/ResultView.js
```diff
@@ -30,14 +30,10 @@ define('ResultView', [ 'jquery', 'jquery-ui', 'ResultObjectList', 'URLUtilities'
...
 						{
-							actions : [
-								{action : 'PublishBatch', label : 'Publish'},
-								{action : 'UnpublishBatch', label : 'Unpublish'}
-							]
-						}, {
 							actions : [
 								{action : 'RestoreBatch', label : 'Restore'},
-								{action : 'OpenDeleteBatch', label : 'Delete'}
+								{action : 'OpenDeleteBatch', label : 'Delete'},
```
bbpennel, line 35: Looking around a bit, it seems that this set of resultActions is passed into ResultObjectActionMenu and used for a bulk actions menu. My recollection is that when multiple things are selected, right clicking to bring up the context menu was supposed to list these actions rather than the usual menu options for individual objects. That doesn't appear to be working here or in production, but the code appears to still be around. See the `batchActions` references in `ResultObjectActionMenu`. 

All that to say, it might be good to check on why that menu isn't working anymore while we're already looking at the bulk actions.

lfarrell: Do you think it should be a separate ticket?

bbpennel: https://jira.lib.unc.edu/browse/BXC-2688

bbpennel:
Deseree reported an issue on the ticket
> Once you complete the action on the items, the buttons in the upper right don't automatically disappear until you reload the page.

do you know if that has been resolved?

static/js/admin/src/ResultView.js
lfarrell: It should work now.

lfarrell:
It depends. Export mods and delete always seem to be there. restore/destroy go away after a restore, but not after a destroy

PR #997, opened 2020-05-26
Fix issues with display showing the wrong user prinicipal in some situations

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2675

### DIFF
--- a/static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
+++ b/static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
@@ -225,6 +225,12 @@
                 return inherited.length > 1 && inherited[0].role === inherited[1].role;
             },
 
+            hasSameOrEmptyInheritedRoles() {
+                let inherited = this.display_roles.inherited.roles;
+                let roleSize = inherited.length;
+                return roleSize === 0 || (roleSize > 1 && inherited[0].role === inherited[1].role);
+            },
+
             hasSameAssignedRoles() {
                 let assigned = this.display_roles.assigned.roles;
                 return assigned.length > 1 && assigned[0].role === assigned[1].role;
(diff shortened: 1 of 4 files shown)


static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
```diff
@@ -378,7 +378,8 @@
 
                 if (users.length === 0) {
                     return users;
-                } else if (this.hasSameAssignedRoles && users[0].role === 'none' && users[1].role === 'none') {
+                } else if (this.hasSameAssignedRoles && users[0].role === 'none' && users[1].role === 'none'
```
bbpennel, line 381: I'm having a hard time following this. If i'm understanding correctly, it will display a single "staff only" line if all of the following are try 1) both the assigned roles are the same, 2) the computed user roles are both "none", 3) either both the inherited roles are the same or there are no inherited roles + the assigned roles at the same.

I'm wondering why both 1 and 2 are required? Seems like the computed roles on the object would be what determines the display here, unless its to avoid rows coming and going too frequently when toggling options? Could 1 & 2 be replaced with `users[0].role === users[1].role && users[0].role === 'none'`? Guess I'm just not sure why the assigned roles matter, but I probably am missing something.

Could `hasSameInheritedRoles` return true if there are no inherited roles? That might simplify things some.

static/js/admin/vue-permissions-editor/tests/unit/patronRoles.spec.js
```diff
@@ -535,6 +558,25 @@ describe('patronRoles.vue', () => {
         });
     });
 
+    it("does not merge display permissions if assigned roles are different but inherited roles are the same", (done) => {
```
bbpennel, line 561: Should this test be verifying that the parents aren't merged? And that the correct rows are being marked as active? (both the parent roles I believe).

lfarrell: Yes. I think I copied another test and apparently only partly updated it.

PR #998, opened 2020-05-27
Filter and display tags

lfarrell:
* Filter and display tags
* Remove unpublished tag

Supersedes #991

https://jira.lib.unc.edu/browse/BXC-2226

### DIFF
--- a/static/js/admin/src/ResultObject.js
+++ b/static/js/admin/src/ResultObject.js
@@ -31,7 +31,7 @@ define('ResultObject', [ 'jquery', 'jquery-ui', 'underscore', 'ModalLoadingOverl
 
 		if (tags.length > 0) {
 			this.metadata.tags = tags.filter(function(d) {
-				return !/^(has|not|no.primary)/i.test(d);
+				return !/^(has|not|no.primary|public.access)/i.test(d);
 			}).map(function(d) {
 				var tagValue;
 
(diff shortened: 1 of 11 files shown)


static/templates/admin/resultEntry.html
```diff
@@ -28,7 +28,7 @@
 			<% for (var i in metadata.tags) { 
 				var tag = metadata.tags[i];
 				%>
-				<a class="status" title="<%= tag.text %>" ><%= _.escape(tag.label) %></a>
+				<a class="status" title="<%= tag.text %>" ><%= _.escape(tag) %></a>
```
bbpennel, line 31: Is `tag.text` still a thing? Seems like the tags are just strings now. It would probably be a good idea to have help text for each tag though, which could be passed in as a map. You're welcome to add that, or create a follow up ticket for it

static/js/admin/src/ResultObject.js
```diff
@@ -60,6 +66,38 @@ define('ResultObject', [ 'jquery', 'jquery-ui', 'underscore', 'ModalLoadingOverl
...
+			case 'embargoed':
+				helpText = 'Object has an active embargo set';
+				break;
+			case 'patron-settings':
+				helpText = 'Patron access settings for this object have been added';
+				break;
+			case 'primary-object':
+				helpText = 'This file is the representative object for the work which contains it';
+				break;
+			case 'public-access':
```
bbpennel, line 89: Based on the list from the ticket, I don't think `public-access` should be showing as a tag, only as a facet

PR #999, opened 2020-05-28
Warn user if moving objects between admin units

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2666

### DIFF
--- a/static/js/admin/src/MoveDropLocation.js
+++ b/static/js/admin/src/MoveDropLocation.js
@@ -35,7 +35,7 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
 
 				var destTitle = self._formatTitle(metadata.title);
 				var targetAdminUnit = self._getAdminUnit(metadata.objectPath);
-				var adminUnitsSize = 1;
+				var adminUnitList = [];
 
 				// Check that we are not moving an object to itself
 				try {
@@ -49,7 +49,9 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
 
 						if (currentAdminUnit !== undefined && targetAdminUnit !== undefined &&
 							currentAdminUnit.pid !== targetAdminUnit.pid) {
-							adminUnitsSize += 1;
+							if (adminUnitList.indexOf(currentAdminUnit.pid) === -1) {
+								adminUnitList.push(currentAdminUnit.pid);
+							}
 						}
 					});
 				} catch (e) {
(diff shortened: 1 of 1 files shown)


static/js/admin/src/MoveDropLocation.js
```diff
@@ -50,27 +50,40 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
...
-				if (destTitle.length > 50) {
-					destTitle = destTitle.substring(0, 50) + "...";
+				var promptText = '';
+				var moveSingleObj = self.manager.dragTargets.length === 1;
+
+				var destTitle = self._formatTitle(metadata.title);
+				var targetAdminUnit = self._getAdminUnit(metadata.objectPath);
+
+				var repTitle = self._formatTitle(representative.metadata.title);
+				var currentAdminUnit = self._getAdminUnit(representative.metadata.objectPath);
```
bbpennel, line 60: I think you'll need to check every object being move to see if it is moving to a different admin unit. You can have objects from multiple admin units in the same search result, in which case we'd particularly want to be able to warn users. If objects are coming from multiple admin units, maybe we can just say "from multiple admin units" rather than showing a title

static/js/admin/src/MoveDropLocation.js
```diff
@@ -50,27 +50,40 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
...
+
+				if (currentAdminUnit.pid !== targetAdminUnit.pid) {
+					if (moveSingleObj) {
+						promptText += " &quot;" + repTitle + "&quot; is ";
```
bbpennel, line 64: I might be missing something, but i think there's an extra space being added before the prompt in this case

static/js/admin/src/MoveDropLocation.js
```diff
@@ -50,27 +50,40 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
...
+				var promptText = '';
+				var moveSingleObj = self.manager.dragTargets.length === 1;
+
+				var destTitle = self._formatTitle(metadata.title);
+				var targetAdminUnit = self._getAdminUnit(metadata.objectPath);
+
+				var repTitle = self._formatTitle(representative.metadata.title);
+				var currentAdminUnit = self._getAdminUnit(representative.metadata.objectPath);
+
+				if (currentAdminUnit.pid !== targetAdminUnit.pid) {
```
bbpennel, line 62: I think this would throw an error if the objectPath didn't have a length > 1 since it returns undefined. I think this could occur if you were moving a collection to a different unit

static/js/admin/src/MoveDropLocation.js
```diff
@@ -32,13 +32,25 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
...
 				try {
 					$.each(self.manager.dragTargets, function() {
 						if (this.pid == metadata.id) {
 							throw "Invalid destination.  Object " + this.pid + " cannot be move into itself.";
 						}
+
+						// Check if the object is being moved to another admin unit
+						var currentAdminUnit = self._getAdminUnit(this.metadata.objectPath);
+
+						if (currentAdminUnit !== undefined && targetAdminUnit !== undefined &&
```
bbpennel, line 50: I haven't verified this, but I am guessing that if you move two items from the same admin unit to a different admin unit, this will result in an adminUnitsSize of 2? Which could be a bit confusing to users, particularly if they are moving a lot more items.

Should add a test to verify this use case.

PR #1002, opened 2020-06-01
BXC-2622 - Register updates to longleaf 

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2622

* Support registration of binary updates and creations to longleaf via camel routes.
* Registration to longleaf performed in batches
* Deposit longleaf registration job switched into camel processor

Depends on:
https://github.com/UNC-Libraries/longleaf-preservation/pull/26

### DIFF
--- a/services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/LongleafRouter.java
+++ b/services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/LongleafRouter.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.dl.services.camel.longleaf;
+
+import org.apache.camel.BeanInject;
+import org.apache.camel.LoggingLevel;
+import org.apache.camel.builder.RouteBuilder;
+
+/**
+ * Router for longleaf operations
+ *
+ * @author bbpennel
+ */
+public class LongleafRouter extends RouteBuilder {
+
+    @BeanInject(value = "registerLongleafProcessor")
+    private RegisterToLongleafProcessor registerProcessor;
+
+    @Override
+    public void configure() throws Exception {
+        from("direct-vm:filter.longleaf")
+            .filter().method(RegisterToLongleafProcessor.class, "registerableBinary")
+            .log(LoggingLevel.DEBUG, "Queuing ${headers[CamelFcrepoUri]} for registration to longleaf")
+            .to("sjms:register.longleaf?transacted=true");
+
+        from("sjms-batch:register.longleaf?completionTimeout={{longleaf.register.completionTimeout}}"
+                + "&completionSize={{longleaf.register.completionSize}}"
+                + "&consumerCount={{longleaf.register.consumers}}"
+                + "&aggregationStrategy=#longleafAggregationStrategy"
+                + "&connectionFactory=jmsFactory")
+            .log(LoggingLevel.DEBUG, "Processing batch of longleaf registrations")
+            .bean(registerProcessor);
+    }
+
+}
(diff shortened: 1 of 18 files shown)


services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/LongleafRouter.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @BeanInject(value = "registerLongleafProcessor")
+    private RegisterToLongleafProcessor registerProcessor;
+
+    @Override
+    public void configure() throws Exception {
+        from("direct-vm:filter.longleaf")
+            .filter().method(RegisterToLongleafProcessor.class, "registerableBinary")
+            .log(LoggingLevel.DEBUG, "Queuing ${headers[CamelFcrepoUri]} for registration to longleaf")
+            .to("sjms:register.longleaf?transacted=true");
```
lfarrell, line 37: How is this getting to the next route, the only from() goes to sjms-batch?

services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/RegisterToLongleafProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+        );
+
+    private String longleafBaseCommand;
+
+    private RepositoryObjectLoader repoObjLoader;
+
+    private FcrepoClient fcrepoClient;
+
+    /**
+     * The exchange here is expected to be a batch message with one or more
```
lfarrell, line 78: What is this "one or more" of?

services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/RegisterToLongleafProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Executes longleaf register command for a batch of files with digests
+     *
+     * @param digestsMap mapping of digest algorithms to paths plus digest values
+     */
+    private void registerFiles(Map<String, Map<URI, String>> digestsMap) {
+        long start = System.currentTimeMillis();
+
+        StringBuilder sb = new StringBuilder();
+        final int[] cntArray = { 0 };
```
lfarrell, line 146: Why isn't this just a regular int?

services-camel/src/test/java/edu/unc/lib/dl/services/camel/routing/MetaServicesRouterTest.java
```diff
@@ -77,8 +80,7 @@ public void testRouteStartContainer() throws Exception {
 
     @Test
     public void testEventTypeFilter() throws Exception {
-        getMockEndpoint("mock:{{cdr.enhancement.stream.camel}}").expectedMessageCount(0);
```
lfarrell, line 80: Should there still be a test for this endpoint somewhere?

services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/LongleafRouter.java
bbpennel: its a quirk of how sjms-batch works, you have to set destinations as `sjms:`, but the route configuration has to be setup with `sjms-batch`. You can see it scattered around this document:
https://camel.apache.org/components/latest/sjms-batch-component.html
My impression is that from the outside, it just looks like an sjms queue, while internally it is configured to do the batching (there might even be two routes behind the scenes)

services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/RegisterToLongleafProcessor.java
bbpennel: we may never know...
I'll fix that

services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/RegisterToLongleafProcessor.java
bbpennel: because its being incremented inside of nested `forEach` blocks. Variables used inside of these blocks must be final, but for primatives like ints you can't modify them if they are final. So instead its incrementing a value inside of a final array. Other methods would be to use an AtomicInteger or a MutableInt, perhaps one of those would be less confusing. I don't know if there is an performance difference.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/routing/MetaServicesRouterTest.java
bbpennel: The route being tested here doesn't go to that destination anymore. There is a test that checks for messages going to `{{cdr.enhancement.stream.camel}}` down in the test of the `process.creation` route though:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/1002/files/810b15210ea2ce06832543eff1aa46f34fe52367#diff-490c86aa5ca6fd996958ce730fc402a0R132

PR #1003, opened 2020-06-05
Improve UI for illegal move operations

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2670

### DIFF
--- a/static/js/admin/src/MoveDropLocation.js
+++ b/static/js/admin/src/MoveDropLocation.js
@@ -45,7 +45,7 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
 						}
 
 						// Return error message if dropTarget is invalid for the object being moved
-						if (self._invalidTarget(this.metadata, metadata)) {
+						if (!self._validTarget(this.metadata, metadata)) {
 							throw "Invalid move location for " + self._formatTitle(this.metadata.title)
 							+ " object" + (self.manager.dragTargets.length > 1 ? "s" : "")
 							+ " to " + destTitle;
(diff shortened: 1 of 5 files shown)


static/js/admin/src/MoveDropLocation.js
```diff
@@ -140,21 +139,81 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
...
-			this.element.off("click.dropClickBlocking").off("mouseenter.dropTargetHover").off("mouseleave.dropTargetLeave");
+		var targets = $('.structure_content a.res_link, .result_table a.res_link');
+
+		try {
+			if (active) {
+				var self = this;
+				var adminUnits = [];
+
+				// Highlight valid drop locations
+				targets.each(function() {
```
bbpennel, line 150: A few suggestions for this loop:
1. Loop through all the dragTargets before this loop to determine if there are any admin units being moved. That would allow things to fail quickly, and its not something that will vary by potential destination. You could then move the adminUnitsMoving check out of the `targets` loop as well.
2. Initialize `destInfo` once per target, probably right after `var selector = $(this);` since it will not vary by dragTarget.
3. You can move the setting of `targets` down to after adminUnit check. Also, could you rename it to something like "destinations" so that the name isn't so similar to `dragTargets`?
4. If the current `target` is an invalid destination for a `dragTarget`, then there's no need to check anymore `dragTargets` against that destination.

static/js/admin/src/MoveDropLocation.js
```diff
@@ -140,21 +139,81 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
...
+
+				this.element.on("click.dropClickBlocking", "a", function(e) {
+					e.preventDefault();
+				}).on("mouseenter.dropTargetHover", this.options.dropTargetSelector, function() {
+					$(this).addClass("drop_hover");
+				}).on("mouseleave.dropTargetLeave", this.options.dropTargetSelector, function() {
+					$(this).removeClass("drop_hover");
+				});
+			} else {
+				targets.removeClass("moving invalid_target"); // 3.3+ syntax for this changes to ["moving", "invalid_target"]
```
bbpennel, line 196: 3.3+ of what?

PR #1006, opened 2020-06-08
Objects with embargoes failing migration

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2691

### DIFF
--- a/migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ACLTransformationHelpers.java
+++ b/migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ACLTransformationHelpers.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.Property;
@@ -84,8 +85,15 @@ public class ACLTransformationHelpers {
 
         // Migrate existing embargoes
         if (bxc3Resc.hasProperty(CDRProperty.embargoUntil.getProperty())) {
-            destResc.addLiteral(CdrAcl.embargoUntil,
-                    bxc3Resc.getProperty(CDRProperty.embargoUntil.getProperty()).getString());
+            String embargoDate = bxc3Resc.getProperty(CDRProperty.embargoUntil.getProperty()).getString();
+            String regex = "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}$"; // ISO date without milliseconds
+            Pattern pattern = Pattern.compile(regex);
+
+            if (pattern.matcher(embargoDate).matches()) {
+                embargoDate += ".000Z";
+            }
+
+            destResc.addLiteral(CdrAcl.embargoUntil, embargoDate);
         }
 
         // Calculate the most restrictive roles assigned to each patron group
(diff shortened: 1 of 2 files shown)


migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ACLTransformationHelpers.java
```diff
@@ -72,9 +75,9 @@ private ACLTransformationHelpers() {
      */
     public static void transformPatronAccess(Resource bxc3Resc, Resource bxc5Resc, PID parentPid) {
         // For admin units, cache patron access settings so they can be used for children instead
+        Model unitModel = createDefaultModel();
```
bbpennel, line 78: you shouldn't need to create a new model to create a literal here, since you have existing models. You can do `bxc3Resc.getModel()` or `bxc5Resc.getModel()`. There's also a static method for creating typed literals, `ResourceFactory.createTypedLiteral()` which would probably work, although I haven't tried it

migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ACLTransformationHelpers.java
```diff
@@ -195,8 +200,10 @@ private static void mergeParentPatronAcls(PID parentPid, Resource destResc, Prop
         if (parentUnitModel != null) {
             Resource parentUnitResc = parentUnitModel.getResource(parentPid.getRepositoryPath());
             if (!destResc.hasProperty(CdrAcl.embargoUntil) && parentUnitResc.hasProperty(CdrAcl.embargoUntil)) {
-                destResc.addLiteral(CdrAcl.embargoUntil,
-                        parentUnitResc.getProperty(CdrAcl.embargoUntil).getLiteral().getString());
+                String embargoDate = formatEmbargoDate(parentUnitResc.getProperty(
```
bbpennel, line 203: Since this is just carrying over an embargo from a parent object that already went throught the formatting method, I think you can skip most of these steps and just do something like `destResc.addLiteral(CdrAcl.embargoUntil, parentUnitResc.getProperty(CdrAcl.embargoUntil).getObject())`. If that doesn't work, you can probably at least skip the formatEmbargoDate call here

migration-util/src/test/java/edu/unc/lib/dcr/migration/content/ACLTransformationHelpersTest.java
```diff
@@ -335,7 +355,16 @@ private void setInheritFromParent(Resource bxc3Resc, boolean inherit) {
     }
 
     private void addEmbargo(Resource bxc3Resc, String embargoEndDate) {
-        bxc3Resc.addLiteral(CDRProperty.embargoUntil.getProperty(), embargoEndDate);
+        String regex = "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}$"; // ISO date without milliseconds
```
bbpennel, line 358: Adding the milliseconds in here probably means that the actual code is never receiving the embargo without milliseconds during the test run. It should be fine to leave off adding the milliseconds here, which would also be more true to the actual data coming from boxc3.

Also, a note that you shouldn't need to create a new model to create the typed literal down below, as in my previous comment.

migration-util/src/test/java/edu/unc/lib/dcr/migration/content/ACLTransformationHelpersTest.java
```diff
@@ -378,7 +407,7 @@ private void assertNoRolesAssignedforAuthenticated(Resource bxc5Resc) {
 
     private void assertHasEmbargo(String embargoEndDate, Resource bxc5Resc) {
         assertTrue("Resource " + bxc5Resc.getURI() + " did not have expected embargo with end date " + embargoEndDate,
-                bxc5Resc.hasLiteral(CdrAcl.embargoUntil, embargoEndDate));
```
bbpennel, line 381: this isn't verifying that the right embargo end date is set anymore, could you add that back in?

PR #1015, opened 2020-06-15
trigger unchecking all selected items after running a batch action

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2688

### DIFF
--- a/static/js/admin/src/action/AjaxCallbackAction.js
+++ b/static/js/admin/src/action/AjaxCallbackAction.js
@@ -125,9 +125,21 @@ define('AjaxCallbackAction', [ 'jquery', 'jquery-ui', 'RemoteStateChangeMonitor'
 					op.context.target.setState("idle");
 				op.complete(data);
 			}
-		}).fail(function(data) {
-			op.alertHandler.alertHandler('error', textStatus + ", " + error);
-			console.error(textStatus, error);
+
+			// Trigger unchecking all checked objects when action completes
+			$(".select_all").trigger('click', function(){
+				var checkbox = $(this).children("input").prop("checked", false);
+				var toggleFn = checkbox.prop("checked") ? "select" : "unselect";
+				var resultObjects = op.resultObjectList.resultObjects;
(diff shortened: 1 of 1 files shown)


static/js/admin/src/action/AjaxCallbackAction.js
```diff
@@ -125,9 +125,21 @@ define('AjaxCallbackAction', [ 'jquery', 'jquery-ui', 'RemoteStateChangeMonitor'
 					op.context.target.setState("idle");
 				op.complete(data);
 			}
-		}).fail(function(data) {
-			op.alertHandler.alertHandler('error', textStatus + ", " + error);
-			console.error(textStatus, error);
+
+			// Trigger unchecking all checked objects when action completes
+			$(".select_all").trigger('click', function(){
```
bbpennel, line 130: Why does the behavior of the select_all button need to be redefined here? It is already defined here to do the exact same thing:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/static/js/admin/src/ResultTableView.js#L397

lfarrell: In the init method it's only bound to the select_all checkbox. So, the user needs to manually click the checkbox for it to run. Maybe we can break it out into a helper?

bbpennel: Not sure I follow, its being called on the same selector in both cases, `$(".select_all")`. Could we update ResultTableView to cover both cases? It'd be nice to be able to trigger an "unselect all" event/action without having to go through hitting the checkbox, but that may not be practical

I guess this version does have one difference, which is `var checkbox = $(this).children("input").prop("checked", false);` I'm kind of unclear on what that's doing right before toggling the checkbox

lfarrell: I reduced it a bit. resultObjects and and selectionUpdated only seem to be in scope when wrapped in an event.

PR #1016, opened 2020-06-15
Check if file has already been ingested

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2699

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
@@ -63,8 +63,6 @@ public class TransferBinariesToStorageJob extends AbstractDepositJob {
     private static final Set<Resource> TYPES_ALLOWING_DESC = new HashSet<>(asList(
             Cdr.Folder, Cdr.Work, Cdr.Collection, Cdr.AdminUnit, Cdr.FileObject));
 
-    private final JobStatusFactory jobStatusFactory = getJobStatusFactory();
-
     /**
      *
      */
@@ -124,6 +122,7 @@ public class TransferBinariesToStorageJob extends AbstractDepositJob {
     }
 
     private void transferOriginalFile(PID objPid, Resource resc, BinaryTransferSession transferSession) {
+        JobStatusFactory jobStatusFactory = getJobStatusFactory();
         // add storageUri if doesn't already exist. It will exist in a resume scenario.
         if (resc.hasProperty(CdrDeposit.stagingLocation) && !resc.hasProperty(CdrDeposit.storageUri) &&
                 !jobStatusFactory.objectIsIngested(jobUUID, objPid.getId())) {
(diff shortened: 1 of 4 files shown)


metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
```diff
@@ -161,6 +163,18 @@ public void setTotalCompletion(String jobUUID, int totalClicks) {
         }
     }
 
+    public boolean objectIsIngested(String depositId, String objectId) {
```
bbpennel, line 166: Could you change the name of the two new methods to use the "Completed" language instead of "Ingested"? And add javadoc to describe what they are for

deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
```diff
@@ -164,9 +191,15 @@ private void transferDepositManifests(PID objPid, Resource resc, BinaryTransferS
                         manifestPath, getDepositUUID());
                 continue;
             }
+
+            JobStatusFactory jobStatusFactory = getJobStatusFactory();
             PID manifestPid = getDepositManifestPid(objPid, manifestFile.getName());
-            URI storageUri = transferSession.transfer(manifestPid, manifestUri);
-            resc.addLiteral(CdrDeposit.storageUri, storageUri.toString());
+
+            if (!jobStatusFactory.objectIsIngested(depositId, manifestPid.getQualifiedId())) {
```
bbpennel, line 198: This seems like an operation we are going to be doing in many places. It would likely make sense and save some repetition to add protected methods to the parent class for the deposit jobs, `isObjectCompleted` and `markObjectCompleted`. They can both take a PID, the parent class should already have access to the deposit id and jobStatusFactory.

The parent class can also setup the identifier used for the progress tracking field. You should be able to do something like:
`String depositJobId = depositUUID + ":" + this.getClass().getName();`
to get the name of the job running rather than the name of parent class.

deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
```diff
@@ -74,6 +77,7 @@ public TransferBinariesToStorageJob() {
      */
     public TransferBinariesToStorageJob(String uuid, String depositUUID) {
         super(uuid, depositUUID);
+        this.depositId = depositUUID + ":" + TransferBinariesToStorageJob.class.getName();
```
bbpennel, line 80: this would probably be clearer as `depositJobId`

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
```diff
@@ -161,6 +163,31 @@ public void setTotalCompletion(String jobUUID, int totalClicks) {
         }
     }
 
+    /**
+     * Checks if a binary is already in the list of deposited objects for the current deposit
```
bbpennel, line 167: change `binary` to `object` since this isn't limited to binaries

metadata/src/main/java/edu/unc/lib/dl/util/JobStatusFactory.java
```diff
@@ -161,6 +163,31 @@ public void setTotalCompletion(String jobUUID, int totalClicks) {
...
+     * @return boolean
+     */
+    public boolean objectIsCompleted(String depositId, String objectId) {
+        try (Jedis jedis = getJedisPool().getResource()) {
+            return jedis.sismember(JOB_COMPLETED_OBJECTS + depositId, objectId);
+        }
+    }
+
+    /**
+     * A list of the values of the current deposit's objects that have been ingested
```
bbpennel, line 180: change `ingested` to `completed` here and for `objectId`

PR #1020, opened 2020-06-18
Add check to see if virus check has already been run.

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2710

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/validate/VirusScanJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/validate/VirusScanJob.java
@@ -81,36 +81,39 @@ public class VirusScanJob extends AbstractDepositJob {
             verifyRunning();
             PID objPid = href.getKey();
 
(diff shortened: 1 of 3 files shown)


deposit/src/main/java/edu/unc/lib/deposit/validate/VirusScanJob.java
```diff
@@ -79,35 +79,38 @@ public void runJob() {
...
-            case FAILED:
-                failures.put(manifestURI.toString(), result.getSignature());
-                break;
-            case ERROR:
-                throw new Error(
-                        "Virus checks are producing errors: "
-                                + result.getException()
+            PID objPid = href.getKey();
+
+            if (!isObjectCompleted(objPid)) {
```
bbpennel, line 84: Just a style/minimizing changes thing, but instead of indenting everything, you can do an early continue
```
if (isObjectCompleted(objPid)) {
    continue;
}
```

PR #1022, opened 2020-06-18
BXC-2687 - Deposit SHA1s

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2687

* Adds a specialized job for verifying fixity and generating a default digest for original files. 
* Removes fixity checking from technical metadata job
* Adds mechanism for preventing rollback of changes to the deposit model when a job fails. Enabled this for the transfer binaries job.

### DIFF
--- a/deposit/src/test/java/edu/unc/lib/deposit/validate/FixityCheckJobTest.java
+++ b/deposit/src/test/java/edu/unc/lib/deposit/validate/FixityCheckJobTest.java
@@ -67,8 +67,6 @@ public class FixityCheckJobTest extends AbstractDepositJobTest {
 
     private File stagingDir;
 
-
-
     @Before
     public void setup() throws Exception {
         pidMinter = new RepositoryPIDMinter();
--- a/deposit/src/test/java/edu/unc/lib/deposit/validate/FixityCheckJobTest.java
+++ b/deposit/src/test/java/edu/unc/lib/deposit/validate/FixityCheckJobTest.java
@@ -1,0 +1,271 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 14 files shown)


persistence/src/main/java/edu/unc/lib/dl/util/MultiDigestInputStreamWrapper.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    private void calculateDigests() {
+        if (computedDigests != null) {
+            return;
+        }
+
+        if (!streamRetrieved) {
+            // Stream not previously consumed, consume it now in order to calculate digests
+            try (final InputStream is = getInputStream()) {
```
lfarrell, line 151: What is the purpose of reading through the file here? I don't see "is used anywhere else.

deposit/src/test/java/edu/unc/lib/deposit/validate/FixityCheckJobTest.java
```diff
@@ -63,6 +67,8 @@
 
     private File stagingDir;
 
+
```
lfarrell, line 70: There are multiple spaces here.

persistence/src/main/java/edu/unc/lib/dl/util/MultiDigestInputStreamWrapper.java
bbpennel: Its because the input stream is wrapped in a DigestInputStream
https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/1022/files#diff-7e1c346f26b46834efa5d9ed91646b62R83-R90
When you read the contents of the stream, it calculates one or more digests. You could consume the contents of the stream for some other purpose, like writing the stream to a file or something, and it would still calculate the digests. That is why this block is wrapped in `if (!streamRetrieved)`, so that it won't try to re-read the stream. But if we had not already read the stream at the time that digests are needed, then we have to read it. This is the case for the FixityCheckJob since it is loading the files just for the sake of checksumming.

PR #1028, opened 2020-06-24
Allow for file rollback if transfer is interrupted or fails

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2709

### DIFF
--- a/persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClient.java
+++ b/persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClient.java
@@ -15,14 +15,8 @@
  */
 package edu.unc.lib.dl.persist.services.transfer;
 
-import static java.nio.file.StandardCopyOption.COPY_ATTRIBUTES;
-import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
-
 import java.io.IOException;
 import java.net.URI;
-import java.net.URISyntaxException;
-import java.nio.file.CopyOption;
-import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -48,11 +42,6 @@ public class FSToFSTransferClient implements BinaryTransferClient {
     private IngestSource source;
     protected StorageLocation destination;
 
-    private static final CopyOption[] COPY_NO_OVERWRITE = { COPY_ATTRIBUTES };
-    private static final CopyOption[] COPY_ALLOW_OVERWRITE = { COPY_ATTRIBUTES, REPLACE_EXISTING };
-    private static final CopyOption[] MOVE_NO_OVERWRITE = { };
-    private static final CopyOption[] MOVE_ALLOW_OVERWRITE = { REPLACE_EXISTING };
-
     public FSToFSTransferClient(IngestSource source, StorageLocation destination) {
         this.source = source;
         this.destination = destination;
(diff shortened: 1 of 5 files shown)


persistence/src/test/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClientTest.java
```diff
@@ -163,6 +163,22 @@ public void transferReplaceFileAlreadyExists() throws Exception {
         assertFalse("Source file should not exist after transfer", sourceFile.toFile().exists());
     }
 
+    @Test
+    public void rollbackOnTransferInterruption() throws Exception {
+        String existingContent = "I exist";
+
+        Files.createDirectories(binDestPath.getParent());
```
bbpennel, line 170: There's actually already a `transferFileAlreadyExists` test which you could extend to check that no extra files are left over. For a test that could fail part way through, from my previous comment
> One test would be to write a file to a directory, then call file.setReadOnly() on that file, then try to overwrite the file (with replace allowed). It should fail partway through, then you will want to verify that the current file still exists, and that there are no extra files in the directory. You'll might need to set file.setWriteable(true) afterwards so that you can delete the file when the test ends, but i haven't verified that.

persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClient.java
```diff
@@ -69,30 +70,77 @@ public URI transferReplaceExisting(PID binPid, URI sourceFileUri) {
...
+            if (!allowOverwrite && destFileExists) {
+                throw new BinaryAlreadyExistsException("Failed to transfer " + sourceFileUri
+                        + ", a binary already exists in " + destination.getId() + " at path " + destUri);
+            }
+
+            // Copy/move new file
             if (source.isReadOnly()) {
-                Files.copy(Paths.get(sourceFileUri), Paths.get(destUri),
+                Files.copy(Paths.get(sourceFileUri), newFilePath,
                         allowOverwrite ? COPY_ALLOW_OVERWRITE : COPY_NO_OVERWRITE);
```
bbpennel, line 93: I think we only need `COPY_NO_OVERWRITE` at this point, an overwrite shouldn't occur and if it does we would want it to fail, which is what would happen in the `move` case

persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClient.java
```diff
@@ -49,8 +50,8 @@
 
     private static final CopyOption[] COPY_NO_OVERWRITE = { COPY_ATTRIBUTES };
     private static final CopyOption[] COPY_ALLOW_OVERWRITE = { COPY_ATTRIBUTES, REPLACE_EXISTING };
-    private static final CopyOption[] MOVE_NO_OVERWRITE = { };
-    private static final CopyOption[] MOVE_ALLOW_OVERWRITE = { REPLACE_EXISTING };
+
+    private static final Logger log = LoggerFactory.getLogger(FSToFSTransferClient.class);
 
     public FSToFSTransferClient(IngestSource source, StorageLocation destination) {
```
bbpennel, line 56: You will also need to update the `StreamToFSTransferClient` with the roughly same safe write procedure.

persistence/src/test/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClientTest.java
```diff
@@ -163,6 +164,30 @@ public void transferReplaceFileAlreadyExists() throws Exception {
...
+        parentDir.setReadOnly();
+
+        try {
+            client.transferReplaceExisting(binPid, sourceFile.toUri());
+        } catch (BinaryTransferException e) {
+            assertTrue("Original file should be present", destFile.exists());
+            assertEquals(1, binDestPath.getParent().toFile().listFiles().length);
+        } finally {
+            parentDir.setWritable(true);
+            destFile.deleteOnExit();
```
bbpennel, line 187: you can just delete the file here rather than waiting for exit

persistence/src/test/java/edu/unc/lib/dl/persist/services/transfer/StreamToFSTransferClientTest.java
```diff
@@ -147,4 +175,9 @@ protected void assertContent(Path path, String content) throws Exception {
     protected InputStream toStream(String content) {
         return new ByteArrayInputStream(content.getBytes());
     }
+
+    private Path createFile(Path filePath, String content) throws Exception {
```
bbpennel, line 179: since you're adding this helper method, you should probably update the existing `ExistingFile` tests to use it since they are all doing it themselves with stuff like `FileUtils.copyInputStreamToFile(toStream(ORIGINAL_CONTENT), binDestPath.toFile());`

PR #1029, opened 2020-06-25
BXC-2718 - Generate missing migrated deposit records

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2718

* Transformation option to generate referenced deposit records which don't exist as fedora objects
* Option for performing dry run of transformation
* Integration tests
* Minor updates for fedora fault resolution and new packaging type for migrations

### DIFF
--- a/migration-util/src/main/java/edu/unc/lib/dcr/migration/TransformContentCommand.java
+++ b/migration-util/src/main/java/edu/unc/lib/dcr/migration/TransformContentCommand.java
@@ -21,6 +21,8 @@ import static org.slf4j.LoggerFactory.getLogger;
 import java.util.concurrent.Callable;
 
 import org.slf4j.Logger;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
 
 import edu.unc.lib.dcr.migration.content.ContentObjectTransformerManager;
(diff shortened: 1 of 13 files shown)


migration-util/src/main/java/edu/unc/lib/dcr/migration/TransformContentCommand.java
```diff
@@ -71,27 +74,39 @@
...
-        output.info("Populating deposit:");
-        output.info(depositPid.getId());
+        output.info("Populating deposit: " + depositPid.getId());
 
         DepositModelManager depositModelManager = new DepositModelManager(depositPid, parentCommand.tdbDir);
         DepositDirectoryManager depositDirectoryManager = new DepositDirectoryManager(
                 depositPid, parentCommand.depositBaseDir, hashNesting);
 
-        PathIndex pathIndex = new PathIndex();
+        PathIndex pathIndex = (PathIndex) context.getBean("pathIndex");
```
lfarrell, line 105: Why are context files used?

bbpennel: I converted this command over to using the spring context stuff because of adding in the fedora interactions, which required a whole bunch of extra dependencies and config options. The context stuff was already in use in the TransformDepositRecordCommand, so it was a pretty easy switch.

PR #1030, opened 2020-06-30
Add pid locking for premis files

smithjp:
https://jira.lib.unc.edu/browse/BXC-2623
https://jira.lib.unc.edu/browse/BXC-2722


smithjp:
I forgot to open this pr last week.  I do have some questions.

How can I reuse the pid locking code without duplicating it or creating a circular dependency between the fcrepo-clients and persistence packages?

The only other external binary log or metadata file I was able to find was the datastream history log in the persistence package which already had pid locking.  Are there any others that I missed?

I wasn't sure how to test these changes.  Is there a good way to test pid locking?

bbpennel:
> How can I reuse the pid locking code without duplicating it or creating a circular dependency between the fcrepo-clients and persistence packages?
The approach you've taken of moving the PidLockManager up to the fcrepo-clients module is good, you can remove the instances of it in persistence. You can see that we've needed to do that with a few other things with all the interfaces in fcrepo-clients in `edu.unc.lib.dl.persist.api` packages. You could add this class to that package. After the 5.0 release I want to revisit our modules to refocus them and to split interfaces from implementations.

> The only other external binary log or metadata file I was able to find was the datastream history log in the persistence package which already had pid locking. Are there any others that I missed?

Since it seems like MODS and its history are already covered, I think that is all of them that can be modified at the moment. There are other datastreams, like FITS and the original file, but they are only updated during creation at the moment.

> I wasn't sure how to test these changes. Is there a good way to test pid locking?
You could test for it with the PREMIS logger by performing concurrent modifications to the log and verifying that all the changes make it into the final version. Here's a recent example of doing something like this (the generating threads, starting them all at once, joining, etc) https://github.com/UNC-Libraries/fedora-repair-siblings/blob/main/src/test/java/edu/unc/lib/fedora/repairsiblings/RepairIT.java#L166
Try the test without your changes as well to verify that its making a difference.

fcrepo-clients/src/main/java/edu/unc/lib/dl/event/FilePremisLogger.java
```diff
@@ -47,6 +49,8 @@
     private PID objectPid;
     private Model model;
 
+    private static final PidLockManager lockManager = PidLockManager.getDefaultPidLockManager();
```
bbpennel, line 52: It may not be necessary to do locking in this client since these changes are happening outside of Fedora in a temp file, so it seems unlikely for there to be conflicts within here.

fcrepo-clients/src/main/java/edu/unc/lib/dl/event/RepositoryPremisLogger.java
```diff
@@ -133,6 +137,8 @@ public PremisLogger writeEvents(Resource... eventResources) {
             log.debug("Adding events to PREMIS log for {}", objPid);
 
             PID logPid = getMdEventsPid(objPid);
+
+            Lock logLock = lockManager.awaitWriteLock(logPid);
```
bbpennel, line 141: At some point we should also add in read locking, so that it won't try to read while a write is being performed. I've created a followup ticket for doing this with the premis logger: https://jira.lib.unc.edu/browse/BXC-2722

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -166,4 +169,65 @@ public void getEventsModelForObjectWithoutLog() throws Exception {
 
         assertTrue(eventsModel.isEmpty());
     }
+
+    @Test
+    public void makeMultipleChangesSimultaneously() throws InterruptedException {
```
bbpennel, line 174: Good work on this test. Do you know if it fails without the new pid locking code? It can sometimes be hard to trigger concurrency cases with a small number of requests. 

It might be good to check the premis log after all the threads have completed to verify that all the expected events are present at once, just to make sure we end up with a merger of the two change sets at the end.

Its too bad we're using a mock binary transfer session in all these tests, so we aren't getting a test of currency in the write code, but judging by the comment on the mock we're probably stuck with it.

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -166,4 +169,70 @@ public void getEventsModelForObjectWithoutLog() throws Exception {
...
+        initPremisLogger(parentObject);
+
+        // make sure that there are no events in premis log before the writes
+        PremisLogger retrieveLogger = new RepositoryPremisLogger(parentObject, mockSession,
+                pidMinter, repoObjLoader, repoObjFactory);
+        Model initialLogModel = retrieveLogger.getEventsModel();
+        assertFalse("New premis already contains events", initialLogModel.listObjects().hasNext());
+
+        // add original event
+        Resource originalEvent = logger.buildEvent(Premis.note)
```
bbpennel, line 185: It would be more valuable to check that this event exists after the concurrent operations below take place, to show that the whole log makes it through. We can count on the other tests to already be making sure that normal addition of events works

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -166,4 +181,221 @@ public void getEventsModelForObjectWithoutLog() throws Exception {
...
+        events.add(event1Resc);
+
+        Date ingestDate = Date.from(Instant.parse("2010-01-02T12:00:00Z"));
+        Resource event2Resc = logger.buildEvent(null, Premis.Ingestion, ingestDate)
+                .addEventDetail("Ingested")
+                .create();
+        events.add(event2Resc);
+
+        for (int i = 1; i <= 200; i++) {
+            Resource anotherEvent = logger.buildEvent(new PID("event" + System.currentTimeMillis() + i), Premis.note,
```
bbpennel, line 218: At some point PID is going to become a interface, but this should be fine for now.

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -166,4 +181,221 @@ public void getEventsModelForObjectWithoutLog() throws Exception {
...
+        Resource originalEventResc = logger.buildEvent(Premis.note)
+                .addEventDetail("first premis event")
+                .write();
+
+        // add new events
+        List<Thread> threads = new ArrayList<>();
+        List<Resource> events = new ArrayList<>();
+        List<String> eventUris = new ArrayList<>();
+
+        Resource event1Resc = logger.buildEvent(Premis.VirusCheck)
```
bbpennel, line 206: event1 and 2 here seem a little superfluous, seems like they are testing the same thing as the events from the loop. Maybe we should just have originalEvent and then all the events from the loop? It'd make the test a bit shorter and easier to understand what's being tested.

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -166,4 +181,221 @@ public void getEventsModelForObjectWithoutLog() throws Exception {
...
+        initPremisLogger(parentObject);
+
+        // make sure that there are no events in premis log before the writes
+        PremisLogger retrieveLogger = new RepositoryPremisLogger(parentObject, mockSession,
+                pidMinter, repoObjLoader, repoObjFactory);
+        Model initialLogModel = retrieveLogger.getEventsModel();
+        assertFalse("New premis already contains events", initialLogModel.listObjects().hasNext());
+
+        // add event
+        Resource event = logger.buildEvent(Premis.note)
```
bbpennel, line 317: since the test isn't ever checking for the presence of this event, seems like we could probably remove it

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -166,4 +181,221 @@ public void getEventsModelForObjectWithoutLog() throws Exception {
...
+        List<String> premisNotes = new ArrayList<>();
+        Runnable readThreadRunnable = new Runnable() {
+            @Override
+            public void run() {
+                Model logModel = retrieveLogger.getEventsModel();
+                Resource logEventResc = logModel.getResource(anotherEvent.getURI());
+                premisNotes.add(logEventResc.getProperty(Premis.note).getString());
+            }
+        };
+        Thread readThread = new Thread(readThreadRunnable);
```
bbpennel, line 344: i'm not sure this is showing that the read waited for the write, since we aren't verifying the read hasn't completed before the write

I think you would need to start the read and write threads at the same time, so set up threads in advance and then `writeThread.start();` immediately followed by `readThread.start();`. You'd want to show that the read thread had not completed yet with `premisNotes.size() == 0` before the `writeThread.join();`, and then that the event was present after. 

In order to get the timing to work out reliably, in the writeThread you could do:
1. get a write lock for the object
2. start the writeEvents operation
3. sleep for 25ms
4. release the write lock

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -166,4 +181,221 @@ public void getEventsModelForObjectWithoutLog() throws Exception {
...
+        initPremisLogger(parentObject);
+
+        // make sure that there are no events in premis log before the writes
+        PremisLogger retrieveLogger = new RepositoryPremisLogger(parentObject, mockSession,
+                pidMinter, repoObjLoader, repoObjFactory);
+        Model initialLogModel = retrieveLogger.getEventsModel();
+        assertFalse("New premis already contains events", initialLogModel.listObjects().hasNext());
+
+        // add an event
+        logger.buildEvent(Premis.note)
```
bbpennel, line 367: This test looks good, my only suggestion is that this event seems unnecessary

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
```diff
@@ -81,7 +93,10 @@ public void init() throws Exception {
                     @Override
                     public URI answer(InvocationOnMock invocation) throws Throwable {
                         InputStream contentStream = invocation.getArgumentAt(1, InputStream.class);
-                        copyInputStreamToFile(contentStream, path.toFile());
+                        Path tempFilePath = createTempFile("temp_content", null);
+                        copyInputStreamToFile(contentStream, tempFilePath.toFile());
+                        contentStream.close();
```
bbpennel, line 98: Its not doing any harm, but you don't need to close the stream in this case. Most InputStream read operations close the stream, and this particular one specifies that it does so https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#copyInputStreamToFile(java.io.InputStream,%20java.io.File)

fcrepo-clients/src/main/java/edu/unc/lib/dl/event/RepositoryPremisLogger.java
```diff
@@ -101,54 +105,60 @@ public PremisEventBuilder buildEvent(Resource eventType) {
...
+                    InputStream mergedStream = new SequenceInputStream(
+                            existingLogStream,
+                            newContentStream);
+
+                    updateOrCreateLog(mergedStream);
+                } catch (IOException e) {
+                    throw new RepositoryException("Failed to close log existing stream", e);
+                }
             }
+        } catch (Exception e) {
```
bbpennel, line 158: It should be fine to remove this `catch` and just go directly to the `finally`

fcrepo-clients/src/test/java/edu/unc/lib/dl/event/RepositoryPremisLoggerIT.java
bbpennel: When running this test locally I was curious to see what order threads were starting and stopping, so I added System.out.printlns at the beginning and end of each thread. The tiny extra delay seems to cause the readThread to perform its read before the write lock goes into place most of the time, so I get the error below, and the test fails. Did you run into this at all or on travis? I'm just a bit worried the test will flap from time to time.

```
Starting write
Starting read
Exception in thread "Thread-6" java.lang.NullPointerException
	at edu.unc.lib.dl.event.RepositoryPremisLoggerIT$5.run(RepositoryPremisLoggerIT.java:320)
	at java.lang.Thread.run(Thread.java:821)
Finished write
```

what I was running
```
        Runnable writeThreadRunnable = new Runnable() {
            @Override
            public void run() {
                System.out.println("Starting write");
                logger.writeEvents(anotherEvent);
                System.out.println("Finished write");
            }
        };
        Thread writeThread = new Thread(writeThreadRunnable);

        // start read thread
        List<String> premisNotes = new ArrayList<>();
        Runnable readThreadRunnable = new Runnable() {
            @Override
            public void run() {
                System.out.println("Starting read");
                Model logModel = retrieveLogger.getEventsModel();
                Resource logEventResc = logModel.getResource(anotherEvent.getURI());
                premisNotes.add(logEventResc.getProperty(Premis.note).getString());
                System.out.println("Finished read");
            }
        };
```

smithjp: I forgot to add the sleep between starting the write and read threads.  It is no longer flapping locally.

bbpennel: hmm, they are no longer happening concurrently are they? I'd guess the write thread finishes before read starts, and the 0 size check passes because its reading the list faster than the readThread reads from the logger. I guess I was thinking the threads should start at the same time, but there would be a sleep within the write thread instead:
```
Runnable writeThreadRunnable = new Runnable() {
            @Override
            public void run() {
                ReadWriteLock writeLock = pidLockManager.awaitWriteLock();
                logger.writeEvents(anotherEvent);
                Thread.sleep(25);
                writeLock.unlock();
            }
        };
```

PR #1031, opened 2020-06-30
Allow child object MODS to be exported

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2651


services/src/test/java/edu/unc/lib/dl/cdr/services/rest/modify/ExportXMLIT.java
```diff
@@ -140,7 +194,7 @@ public void testNoUsernameProvided() throws Exception {
         assertEquals("User must have a username to export xml", respMap.get("error"));
     }
 
-    private String createObjectsAndMakeJSON(boolean exportChildren) throws Exception {
```
bbpennel, line 143: this methods not doing what it says anymore since it no longer makes json. Maybe rename it to `createObjects` and have it just return the pids `List<String>`? Then have another method named `makeExportJson` which makes the request and turns it into json. You don't need the request object in most of the tests

PR #1032, opened 2020-06-30
Add retry if work object ingest gives a checksum mismatch

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2698

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
@@ -292,24 +292,7 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
                 } else if (childResc.hasProperty(RDF.type, Cdr.Folder)) {
                     ingestFolder(destObj, parentResc, childResc);
                 } else if (childResc.hasProperty(RDF.type, Cdr.Work)) {
-                    boolean ingested = false;
-                    int retryCount = 0;
-
-                    while (!ingested) {
-                        try {
-                            ingestWork(destObj, parentResc, childResc);
-                            ingested = true;
-                        } catch (ChecksumMismatchException e) {
-                            if (retryCount < 3) {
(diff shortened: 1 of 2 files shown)


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -630,6 +630,28 @@ private void ingestWork(ContentContainerObject parent, Resource parentResc, Reso
 
                 // Cease refreshing the transaction
                 txRefresher.stop();
+            } catch (ChecksumMismatchException e) {
```
bbpennel, line 633: what about overriding the method signature to `private void ingestWork(ContentContainerObject parent, Resource parentResc, Resource childResc, int retries)` and the same thing without `retries`. Most stuff would call the version without retries, which would supply the starting number of retries at 3.

Then you should be able to do something like:
```
} catch (ChecksumMismatchException e) {
     // Rollback changes for the work so far
     txRefresher.interrupt();
     tx.cancel();
     if (retries > 0) {
         retries--;
         ingestWork(parent, parentResc, childResc, retries);
     } else {
         failJob(...);
     }
```
I think if a work had more than one file and failed partway through, it would try to reingest those other files and run into a conflict since they'd already exist.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -630,6 +630,16 @@ private void ingestWork(ContentContainerObject parent, Resource parentResc, Reso
...
+                txRefresher.interrupt();
+                tx.cancelAndIgnore();
+
+                if (retries > 0) {
+                    retries--;
+                    ingestWork(parent, parentResc, childResc, retries);
```
bbpennel, line 639: Could you add a `warn` level message stating that there was a checksum mismatch, probably include the message from the error? So that we see if this issue is occurring

PR #1034, opened 2020-07-01
Accessibility updates

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2496
Does not address contrast issues. Probably needs a larger discussion on color choices.

### DIFF
--- a/static/css/sass/cdr-ui.scss
+++ b/static/css/sass/cdr-ui.scss
@@ -1,7 +1,6 @@
 @import url('https://fonts.googleapis.com/css?family=Libre+Baskerville:400i|Open+Sans:400,600');
 
 $unc-blue: #4B9CD3;
-$link-blue: #007FAE;
 $unc-navy: #13294B;
 $light-gray: #E1E1E1;
 $dark-gray: #767676;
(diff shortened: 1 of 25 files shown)


static/css/sass/cdr-ui.scss
```diff
@@ -664,8 +673,8 @@ p.spacing {
 }
 
 .item_container_count {
-  background-color: $unc-blue;
-  border: 1px solid $unc-blue;
+  background-color: #1A698C;
```
bbpennel, line 676: Why not use a variable for these anymore?

static/css/sass/cdr_ui_styles.scss
```diff
@@ -32,6 +33,18 @@ $box-shadow: inset 3px 3px 10px -1px $box-shadow-color, 0px 16px 22px -15px $box
   }
 }
 
+/* Class to display content for screen readers */
+.sr-only {
```
bbpennel, line 37: is this all standard for `sr-only` elements?

lfarrell: I stole it from bootstrap. It's what they use.

static/css/sass/cdr-ui.scss
lfarrell: $unc-blue still gets used in other places. This is a different, darker color, though it does get used a few times. I can add it as a variable

PR #1037, opened 2020-07-06
Skip transformation of deposit records for items in the CDR

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2726


migration-util/src/test/java/edu/unc/lib/dcr/migration/deposit/DepositRecordTransformerIT.java
```diff
@@ -225,6 +226,80 @@ public void transform_DepositRecord_WithDepositProperties() throws Exception {
         assertTrue(recResc.hasProperty(Cdr.depositPackageProfile, "package subtype"));
     }
 
+    @Test(expected = NotFoundException.class)
```
bbpennel, line 229: This all looks good. Two minor things:
1. all three of these tests start the same way, it might become more readable if (or at least easier to identify the differences) if you moved lines 231 through 245 into a helper method that takes the packaging type as a parameter
2. Do you know if its possible for `transformer.compute();` to throw a `NotFoundException`? If so, that test might incorrectly pass without reaching `repoObjLoader.getDepositRecord(bxc5Pid);`. That might not be an issue though

lfarrell: Looks like it already has the foxml at that point and there aren't any read operations against fedora.

PR #1038, opened 2020-07-06
Update deposited count after adding children

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2728

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
@@ -614,9 +614,6 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
 
                 addDescription(obj, childResc);
 
-                // Increment the count of objects deposited prior to adding children
-                addClicks(1);
-
                 log.info("Created work object {} for deposit {}", childPid, getDepositPID());
 
                 ingestChildren(obj, childResc);
@@ -628,6 +625,9 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
                 addPremisEvents(obj);
 
                 overrideModifiedTimestamp(obj, childResc);
+
+                // Increment the count of objects deposited after adding children
+                addClicks(1);
 
                 // Cease refreshing the transaction
                 txRefresher.stop();
--- a/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
@@ -438,17 +438,16 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
                 parent.addMember(obj);
 
                 addDescription(obj, childResc);
-
-                // Increment the count of objects deposited prior to adding children
-                addClicks(1);
-                getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
-
                 log.info("Created folder object {} for deposit {}", childPid, getDepositPID());
             } catch (Exception e) {
                 tx.cancel(e);
             } finally {
                 tx.close();
             }
+
+            // Increment the count of objects deposited prior to adding children
+            addClicks(1);
+            getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
         }
 
         // ingest children of the folder
@@ -489,17 +488,16 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
                 parent.addMember(obj);
 
                 addDescription(obj, childResc);
-
-                // Increment the count of objects deposited prior to adding children
-                addClicks(1);
-                getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
-
                 log.info("Created admin unit {} for deposit {}", childPid, getDepositPID());
             } catch (Exception e) {
                 tx.cancel(e);
             } finally {
                 tx.close();
             }
+
+            // Increment the count of objects deposited prior to adding children
+            addClicks(1);
+            getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
         }
 
         // ingest children of the admin unit
@@ -540,17 +538,16 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
                 parent.addMember(obj);
 
                 addDescription(obj, childResc);
-
-                // Increment the count of objects deposited prior to adding children
-                addClicks(1);
-                getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
-
                 log.info("Created collection {} for deposit {}", childPid, getDepositPID());
             } catch (Exception e) {
                 tx.cancel(e);
             } finally {
                 tx.close();
             }
+
+            // Increment the count of objects deposited prior to adding children
+            addClicks(1);
+            getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
         }
 
         // ingest children of the admin unit
@@ -623,9 +620,6 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
 
                 overrideModifiedTimestamp(obj, childResc);
 
-                // Increment the count of objects deposited after adding children
-                addClicks(1);
-                getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
                 // Cease refreshing the transaction
                 txRefresher.stop();
             } catch (ChecksumMismatchException e) {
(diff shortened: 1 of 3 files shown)


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -629,6 +626,9 @@ private void ingestWork(ContentContainerObject parent, Resource parentResc, Reso
 
                 overrideModifiedTimestamp(obj, childResc);
 
+                // Increment the count of objects deposited after adding children
+                addClicks(1);
```
bbpennel, line 630: We probably need to do this for other object types too? Would it make sense to do this at the same time `incrIngestedObjects` happens below?

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -441,15 +441,14 @@ private void ingestFolder(ContentContainerObject parent, Resource parentResc, Re
 
                 // Increment the count of objects deposited prior to adding children
                 addClicks(1);
+                getDepositStatusFactory().incrIngestedObjects(getDepositUUID(), 1);
```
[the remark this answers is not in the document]
bbpennel: Looking at the stuff that happens after "ingestChildren" calls in most of these methods, it seems like some of that stuff should not be happening during resume if all the children finished (adding events particularly). That doesn't need to be part of this ticket though.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -441,15 +441,14 @@ private void ingestFolder(ContentContainerObject parent, Resource parentResc, Re
```
bbpennel, line 441: I think `ingestFileObject` could still be problematic since its happening inside of the transaction for a work. If it fails after a few files the count would be off. Handling this might be a bit tricky, you might have to keep a running count of the number of file objects added rather than committing to redis immediately, and then include that count in the addClicks and incrIngestedObjects when the transaction ends for the work object. It'd definitely be good to have a test to ensure the counts come out right in a failure case. Is there an existing test that might be used for this?

lfarrell: Wouldn't this cover that case?
```java
if (skipResumed(childResc)) {
     return;
}
```

There is this test https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java#L444. Maybe something like it would work?

bbpennel: If the work transaction got rolled back, then the file wouldn't be in fedora, so we'd need to reingest the object. skipResumed is checking against objects in fedora, so it wouldn't prevent the fileObject from getting ingested and the counter from getting incremented.

I don't see a test that verifies resuming a work deposit with multiple files partway through, but you could make a new test from the one you linked, adding 2 or 3 files to the work and having it fail on one of the files, then fix the error and re-run it, then make sure the count is correct

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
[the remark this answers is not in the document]
bbpennel: I've created a followup ticket for skipping the `ingestChildren` and stuff after it during resume https://jira.lib.unc.edu/browse/BXC-2730

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -153,6 +153,8 @@
 
     private boolean overrideTimestamps;
 
+    private int filesIngested = 0;
```
bbpennel, line 156: could you rename this to something like `filesIngestInWork` and/or add a comment explaining that it is tracking the file count within one work, not across the whole deposit? Just seems worth clarifying since its a class level variable

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -629,6 +620,10 @@ private void ingestWork(ContentContainerObject parent, Resource parentResc, Reso
 
                 overrideModifiedTimestamp(obj, childResc);
 
+                // Add counts for files deposited for the work
```
bbpennel, line 623: Wrap these two calls in an `if (filesIngested > 0)` block in case there were no files.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -476,6 +476,50 @@ public void resumeCompletedWorkIngestTest() throws Exception {
...
+
+        PID mainPid = addFileObject(workBag, FILE1_LOC, FILE1_MIMETYPE, FILE1_SHA1, FILE1_MD5);
+        PID mainPid2 = addFileObject(workBag, FILE1_LOC, FILE1_MIMETYPE, null, FILE1_MD5_BAD);
+
+        depBag.add(workBag);
+
+        workBag.asResource().addProperty(Cdr.primaryObject,
+                model.getResource(mainPid.getRepositoryPath()));
+
+        workBag.asResource().addProperty(Cdr.primaryObject,
```
bbpennel, line 501: there should only be one primaryObject per work (that's enforced elsewhere). Also, you don't need the `.asResource()` part

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -476,6 +476,50 @@ public void resumeCompletedWorkIngestTest() throws Exception {
...
+
+        job.closeModel();
+
+        try {
+            job.run();
+            fail("Test should not reach this line. Job should have thrown exception");
+        } catch (JobFailedException e) {
+            Map<String, String> jobStatus = jobStatusFactory.get(jobUUID);
+            assertNull(jobStatus.get(JobField.num.name()));
+            Resource fileResc =job.getWritableModel().getResource(mainPid2.getRepositoryPath());
```
bbpennel, line 512: space after `=`

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -476,6 +476,50 @@ public void resumeCompletedWorkIngestTest() throws Exception {
...
+            fileResc.addProperty(CdrDeposit.md5sum, FILE1_MD5);
```
bbpennel, line 513: You also need to remove the bad md5sum property, otherwise the test might flap since there will be two md5sums and it might randomly pick the bad one.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobTest.java
```diff
@@ -307,7 +307,7 @@ public void ingestWorkObjectTest() throws Exception {
                 eq(supMime), anyString(), anyString(), any(Model.class));
         verify(work).setPrimaryObject(mainPid);
 
-        verify(jobStatusFactory, times(3)).incrCompletion(eq(jobUUID), eq(1));
+        verify(jobStatusFactory, times(1)).incrCompletion(eq(jobUUID), eq(2));
```
bbpennel, line 310: it should also be called one time with a value of 1

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobTest.java
```diff
@@ -434,7 +434,7 @@ public void resumeIngestWorkObjectTest() throws Exception {
         verify(work).setPrimaryObject(mainPid);
 
         verify(jobStatusFactory).setCompletion(eq(jobUUID), eq(2));
-        verify(jobStatusFactory).incrCompletion(eq(jobUUID), eq(1));
+        verify(jobStatusFactory, never()).incrCompletion(eq(jobUUID), eq(1));
```
bbpennel, line 437: `incrCompletion` is still getting called, isn't it? Just not with the value 1

lfarrell: No, but it looks like setTotalCompletion() is being set

PR #1039, opened 2020-07-07
BXC-2426 - Deregister destroyed files from longleaf

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2426

* Adds deregistration from longleaf of destroyed binaries.
* Implements bulk deregistration of binaries in longleaf
* Destroy job triggers deregistration from longleaf. Add more debug statements during destroy
* Allow longleaf message aggregator to get uri from body or header

### DIFF
--- a/services-camel/src/test/java/edu/unc/lib/dl/services/camel/longleaf/DeregisterLongleafRouteTest.java
+++ b/services-camel/src/test/java/edu/unc/lib/dl/services/camel/longleaf/DeregisterLongleafRouteTest.java
@@ -1,0 +1,162 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.dl.services.camel.longleaf;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.net.URI;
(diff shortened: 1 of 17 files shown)


services-camel/src/test/java/edu/unc/lib/dl/services/camel/longleaf/DeregisterLongleafRouteTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        output = LongleafTestHelpers.readOutput(outputPath);
+        assertDeregisterCalled(1);
+        assertDeregisterPaths(contentUri);
+    }
+
+    @Test
+    public void deregisterSingleBatch() throws Exception {
+        NotifyBuilder notify = new NotifyBuilder(cdrLongleaf)
+                .whenCompleted(1 + 3)
```
lfarrell, line 95: Why the 1 + 3?

services-camel/src/test/java/edu/unc/lib/dl/services/camel/longleaf/DeregisterLongleafRouteTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        output = LongleafTestHelpers.readOutput(outputPath);
+        assertDeregisterCalled(1);
+        assertDeregisterPaths(contentUris);
+    }
+
+    @Test
+    public void deregisterMultipleBatches() throws Exception {
+        NotifyBuilder notify = new NotifyBuilder(cdrLongleaf)
+                .whenCompleted(2 + 10)
```
lfarrell, line 112: Same here with the 2 + 10?

services-camel/src/test/java/edu/unc/lib/dl/services/camel/longleaf/DeregisterLongleafRouteTest.java
bbpennel: I wrote it that was since there are two types of messages being completed, 1 batch message and 3 individual messages, since the notifyBuilder tracks the whole camel context rather than a specific route within it. Perhaps a comment would be better though

services-camel/src/test/java/edu/unc/lib/dl/services/camel/longleaf/DeregisterLongleafRouteTest.java
bbpennel: I've added a comment to clarify

PR #1040, opened 2020-07-10
BXC-2711 - Improve deposit pausing

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2711

* Adds breakpoints in longer running deposit jobs, where the job checks to see if it has been paused or interrupted, and ends early if so.
* Switch binary transfer client to write using utility method which can be interrupted
* Adjustments to exception throwing, fix for virus scans, and cleanup of extraneous code

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
@@ -92,9 +92,13 @@ public class BagIt2N3BagJob extends AbstractFileServerToBagJob {
             MandatoryVerifier.checkBagitFileExists(bagReader.getRootDir(), bagReader.getVersion());
 
             try (BagVerifier verifier = new BagVerifier()) {
+                interruptJobIfStopped();
                 verifier.isComplete(bagReader, false);
+                interruptJobIfStopped();
                 verifier.isValid(bagReader, false);
             }
+
+            interruptJobIfStopped();
 
             Set<Manifest> payloadManifests = bagReader.getPayLoadManifests();
 
--- a/deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobTest.java
+++ b/deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobTest.java
@@ -20,6 +20,7 @@ import static edu.unc.lib.dl.acl.util.AccessPrincipalConstants.PUBLIC_PRINC;
 import static edu.unc.lib.dl.persist.services.storage.StorageLocationTestHelper.LOC1_ID;
 import static edu.unc.lib.dl.test.TestHelpers.setField;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyCollectionOf;
 import static org.mockito.Matchers.anyString;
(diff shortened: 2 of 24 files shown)


deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
```diff
@@ -92,10 +92,14 @@ public void runJob() {
             MandatoryVerifier.checkBagitFileExists(bagReader.getRootDir(), bagReader.getVersion());
 
             try (BagVerifier verifier = new BagVerifier()) {
+                interruptJobIfStopped();
                 verifier.isComplete(bagReader, false);
+                interruptJobIfStopped();
                 verifier.isValid(bagReader, false);
             }
 
+            interruptJobIfStopped();
```
lfarrell, line 101: Would this cause any noticeable slowdown in ingest, checking for interruptions so often in succession?

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobTest.java
```diff
@@ -441,6 +450,52 @@ public void resumeIngestWorkObjectTest() throws Exception {
...
+        job.closeModel();
+
+        when(work.addDataFile(any(PID.class), any(URI.class),
+                anyString(), anyString(), anyString(), anyString(), any(Model.class)))
+                .thenReturn(mockFileObj);
+        when(mockFileObj.getPid()).thenReturn(mainPid);
+
+        when(depositStatusFactory.getState(depositUUID))
+                .thenReturn(DepositState.running)
+                .thenReturn(DepositState.running)
```
lfarrell, line 474: Does DepositState.running need to be here twice?

deposit/src/test/java/edu/unc/lib/deposit/validate/VirusScanJobTest.java
```diff
@@ -233,6 +234,49 @@ public void errorScanTest() throws Exception {
...
+
+        File pdfFile = new File(depositDir, "pdf.pdf");
+        File textFile = new File(depositDir, "text.txt");
+        PID file1Pid = addFileObject(depBag, pdfFile);
+        PID file2Pid = addFileObject(depBag, textFile);
+
+        // Should be running for the first file, then paused
+        when(depositStatusFactory.getState(depositUUID))
+                .thenReturn(DepositState.running)
+                .thenReturn(DepositState.running)
```
lfarrell, line 252: Same here, does it need to be here twice?

deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
bbpennel: I tried to balance that by not adding too many of these checks in, and not adding them at all for some of the jobs that finish quickly. I decided which jobs to add it to by looking at the runtimes for each job in previous testing tickets. These checks should be pretty fast since redis is on the same server, and is generally very fast. But it is still a new cost.

In this particular case, there is a max of 3 calls for the job, unlike some of the others where it depends on the number of objects, so I'm not too worried about it in this case. I tried to place them in between long running steps, before it starts writing to the deposit model.

deposit/src/test/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJobTest.java
bbpennel: when multiple `thenReturn`s are chained like this, it means the first and second calls of `getState` will return `running`, and then all calls after that will return `paused`. So the number of runnings depended on the number of calls expected

deposit/src/test/java/edu/unc/lib/deposit/validate/VirusScanJobTest.java
bbpennel: this was so it would register as "paused" on the second file, since there is now an interrupt check at the beginning of every deposit job.

PR #1047, opened 2020-07-16
Return all export results with qualified ids

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2736

### DIFF
--- a/access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
+++ b/access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
@@ -119,9 +119,7 @@ public class SolrQueryLayerService extends SolrSearchService {
      * Retrieves the facet list for the search defined by searchState. The facet results optionally can ignore
      * hierarchical cutoffs.
      *
-     * @param searchState
-     * @param facetsToRetrieve
-     * @param applyCutoffs
+     * @param searchRequest
      * @return
      */
     public SearchResultResponse getFacetList(SearchRequest searchRequest) {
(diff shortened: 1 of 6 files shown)


access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -241,6 +239,10 @@ public boolean hasRole(AccessGroupSet accessGroups, UserRole userRole) {
     }
 
     public SearchResultResponse performSearch(SearchRequest searchRequest) {
+        return performSearch(searchRequest, false);
+    }
+
+    public SearchResultResponse performSearch(SearchRequest searchRequest, boolean allResults) {
```
bbpennel, line 245: Please add "allResults" as a field of SearchRequest since it is part of the request

bbpennel: You'll also need to remove the method parameter here since allResults is in the search request now

access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -249,7 +247,7 @@ public SearchResultResponse performSearch(SearchRequest searchRequest) {
         // Adjust the number of results to retrieve
         if (searchState.getRowsPerPage() == null || searchState.getRowsPerPage() < 0) {
             searchState.setRowsPerPage(searchSettings.defaultPerPage);
-        } else if (searchState.getRowsPerPage() > searchSettings.getMaxPerPage()) {
+        } else if (!searchState.getAllResults() && searchState.getRowsPerPage() > searchSettings.getMaxPerPage()) {
```
bbpennel, line 250: I feel a bit pedantic suggesting it, but something like `ignoreMaxRows` could be clearer about the flags purpose. To me, "getAllResults" reads as a method for retrieving a list of results, rather modifier to a query to override the row limit.

PR #1055, opened 2020-07-23
BXC-2753 - Prevent storage of FITS error mimetypes in transformations

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2753
* Prevent setting mimetype in deposit if it matches the pattern for FITS errors
* Update transformation tests to check for mimetype setting

### DIFF
--- a/migration-util/src/test/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformerTest.java
+++ b/migration-util/src/test/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformerTest.java
@@ -81,6 +81,7 @@ import org.mockito.Mock;
 import edu.unc.lib.dcr.migration.deposit.DepositDirectoryManager;
 import edu.unc.lib.dcr.migration.deposit.DepositModelManager;
 import edu.unc.lib.dcr.migration.fcrepo3.ContentModelHelper.Bxc3UserRole;
+import edu.unc.lib.dcr.migration.fcrepo3.ContentModelHelper.CDRProperty;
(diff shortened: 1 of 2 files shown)


migration-util/src/test/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformerTest.java
```diff
@@ -840,6 +846,60 @@ public void transformFolderWithMissingChild() throws Exception {
...
+
+        // Verify file properties
+        List<RDFNode> bagChildren = workBag.iterator().toList();
+        assertEquals(1, bagChildren.size());
+        assertTrue(bagChildren.contains(child1Resc));
+        assertTrue(child1Resc.hasProperty(RDF.type, Cdr.FileObject));
+        assertTrue(child1Resc.hasLiteral(CdrDeposit.md5sum, DATA_FILE_MD5));
+        assertTrue(child1Resc.hasLiteral(CdrDeposit.createTime, DEFAULT_CREATED_DATE));
+        assertTrue(child1Resc.hasLiteral(CdrDeposit.lastModifiedTime, DEFAULT_CREATED_DATE));
+        assertFalse(child1Resc.hasProperty(CdrDeposit.mimetype));
```
lfarrell, line 897: Is this checking that mimetype property is empty?

bbpennel: Its checking that it doesn't exist with any value

PR #1058, opened 2020-07-29
BXC-2746 - Cleanup PersistingRdfStreamConsumer spam

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2746


lfarrell:
So, it was caused by the way the tests were setup? Fun.

PR #1061, opened 2020-07-31
BXC-2744 - Improve recovery of binary transfer jobs

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2744
* Binary transfer job handles three error cases when a transfer runs into an existing file at the destination
    1. The file has already been transferred, but the storage uri was lost due to an error or the jvm shutting down. In this case, no transfer occurs, and the storage uri is recorded
    2. The file was partially transferred due to interruption. In this case, the file will be overwritten
    3. The pid of the file overlaps with an object in the repository, and therefore they have the same file uri. In this case, the deposit fails
* Removed redis tracking of binary transfers as having the JVM terminate will result in losing storage uris recorded in jena, which would be skipped during resumption, ultimately causing IngestContentJob to fail

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
@@ -150,7 +150,8 @@ public class TransferBinariesToStorageJob extends AbstractDepositJob {
 
             Path stagingPath = getModsHistoryPath(objPid);
             if (Files.exists(stagingPath)) {
-                transferFile(modsPid, stagingPath.toUri(), transferSession, resc, CdrDeposit.descriptiveHistoryStorageUri);
+                transferFile(modsPid, stagingPath.toUri(), transferSession, resc,
+                        CdrDeposit.descriptiveHistoryStorageUri);
                 log.debug("Finished transferring MODS history file {}", modsPid.getQualifiedId());
             }
         }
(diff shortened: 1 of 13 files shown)


deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
```diff
@@ -187,13 +179,38 @@ private void transferDepositManifests(PID objPid, Resource resc, BinaryTransferS
...
 
             PID manifestPid = getDepositManifestPid(objPid, manifestFile.getName());
+            transferFile(manifestPid, manifestUri, transferSession, resc, CdrDeposit.storageUri);
+        }
+    }
+
+    private void transferFile(PID binPid, URI stagingUri, BinaryTransferSession transferSession,
+            Resource resc, Property storageProperty) {
 
-            if (!isObjectCompleted(manifestPid)) {
```
lfarrell, line 191: Do the isObjectCompleted and markObjectCompleted methods in AbstractDepositJob need to be removed?

bbpennel: No, they are still being used by other jobs. It is reasonable to use objectCompleted in cases where the jena model isn't being updated, since those can get lost in cases where the jvm shuts down or the jena transactions are interrupted at the wrong time.

PR #1066, opened 2020-08-04
Set destinations that produce errors  to invalid drop targets instead of cancelling the entire drop

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2752

### DIFF
--- a/static/js/admin/src/MoveDropLocation.js
+++ b/static/js/admin/src/MoveDropLocation.js
@@ -210,7 +210,7 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
 							}
 						} catch (e) { // Just set errored destinations to invalid drop targets instead of cancelling drop
 							selector.addClass("invalid_target");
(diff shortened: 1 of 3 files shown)


static/js/admin/src/MoveDropLocation.js
```diff
@@ -201,10 +201,16 @@ define('MoveDropLocation', [ 'jquery', 'jquery-ui', 'ConfirmationDialog'],
...
+						try {
+							if (self._validTarget(this.metadata, destInfo)) {
+								selector.addClass("moving");
+							} else {
+								selector.addClass("invalid_target");
+								return false;
+							}
+						} catch (e) { // Just set errored destinations to invalid drop targets instead of cancelling drop
 							selector.addClass("invalid_target");
+							console.log('Error checking drop destination for: ' + JSON.stringify(destInfo));
```
bbpennel, line 213: could you output the error in here? When debugging I was finishing it more helpful to show the selector too
`console.log('Error checking drop destination for: ', destInfo, e);`

PR #1078, opened 2020-08-17
Add bxc version to software agent

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2783


migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformer.java
```diff
@@ -363,7 +364,7 @@ private void populateFileObject(Resource bxc3Resc, Model depositModel) {
         // transform existing PREMIS as the events for the file object, rather than work
         transformPremis(originalPid, filePid);
         if (isNewWork) {
-            addMigrationEvent(getPremisLogger(newPid));
+            addMigrationEvent(getPremisLogger(newPid), depositBxc3.getFullname());
```
bbpennel, line 367: if its a new work, then it was generated in bxc5, so you should need to override here

migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformer.java
```diff
@@ -381,18 +382,20 @@ private void transformPremis(PID bxc3Pid, PID bxc5Pid) {
         premisTransformer.compute();
 
         // Add migration event
-        addMigrationEvent(premisLogger);
+        addMigrationEvent(premisLogger, depositBxc3.getFullname());
```
bbpennel, line 385: this event describes the migration of the object from boxc3 to 5, so it should have the migration-util 5.0 agent.

migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformer.java
```diff
@@ -381,18 +382,20 @@ private void transformPremis(PID bxc3Pid, PID bxc5Pid) {
...
     }
 
     private PremisLogger getPremisLogger(PID bxc5Pid) {
         Path transformedPremisPath = directoryManager.getPremisPath(bxc5Pid);
         return premisLoggerFactory.createPremisLogger(bxc5Pid, transformedPremisPath.toFile());
     }
 
-    private void addMigrationEvent(PremisLogger premisLogger) {
+    private void addMigrationEvent(PremisLogger premisLogger, String version) {
+        SoftwareAgentConstants.setCdrVersion(version);
```
bbpennel, line 394: I think you can remove this. It looks like `TransformContentCommand` uses `service-context.xml` now as well, so cdrVersion should be getting set that way already.

migration-util/src/main/java/edu/unc/lib/dcr/migration/premis/ContentPremisToRdfTransformer.java
```diff
@@ -307,7 +308,11 @@ private void addFixityEvent(Element eventEl, String eventOutcome) {
     }
 
     private void addEvent(Element eventEl, Resource eventType, String eventDetail, SoftwareAgent agent) {
-        addEvent(eventEl, eventType, eventDetail, true, agent.getFullname());
+        String agentFullname = agent.getFullname();
```
bbpennel, line 311: i'm guessing these changes won't be necessary after updating depositBxc3 to have the name of the service, so you should be able to roll them back

PR #1080, opened 2020-08-18
Don't throw JobInterrupted exception if state is "finished" and waiting on cleanup job to run

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2782

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/work/AbstractDepositJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/work/AbstractDepositJob.java
@@ -56,7 +56,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 
-import edu.unc.lib.deposit.CleanupDepositJob;
 import edu.unc.lib.dl.event.PremisLogger;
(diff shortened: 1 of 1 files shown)


deposit/src/main/java/edu/unc/lib/deposit/work/AbstractDepositJob.java
```diff
@@ -370,7 +370,10 @@ protected void interruptJobIfStopped() throws JobInterruptedException {
         }
 
         DepositState state = getDepositStatusFactory().getState(depositId);
-        if (!DepositState.running.equals(state)) {
+
+        // Only throw error if state is changed and it's not a finished deposit waiting on cleanup job
+        if (!DepositState.running.equals(state) &&
+                (!DepositState.finished.equals(state) && !jobName.contains("CleanupDepositJob"))) {
```
bbpennel, line 376: I'm not totally sure if the check of the job name is necessary, but interrupting the job if somehow it gets set to finished and isn't a cleanup job is probably fine since that would be an error at this time.
My only suggestion is that instead of doing a contains check on the job name, you can do `this instanceof CleanupDepositJob` instead, or maybe it needs to be `getClass()`, I'm not sure, but either would be good and a bit safer than a string check.

PR #1081, opened 2020-08-20
Fix issue with enhancements trying to run on non-file objects

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2776


services/src/test/java/edu/unc/lib/dl/cdr/services/rest/RunEnhancementsIT.java
```diff
@@ -142,12 +144,34 @@ public void runEnhancementsFileObject() throws Exception {
     }
 
     @Test
-    public void runEnhancementsNonFileObject() throws Exception {
+    public void runEnhancementsWorkObject() throws Exception {
+        WorkObject workObj = repositoryObjectFactory.createWorkObject(null);
+        FileObject workFile = workObj
+                .addDataFile(makeContentUri(BINARY_CONTENT), "file.png", "image/png", null, null);
+        PID filePid = workFile.getPid();
+        setResultMetadataObject(filePid, ResourceType.Work.name());
```
bbpennel, line 152: i think to be more accurate to the case being tested, you should pass the pid of the work in here and then call the api on the work pid. I suspect it doesn't make a difference at the moment, but having a file labeled as a work in the solr indexed may produce unexpected results in the future

PR #1082, opened 2020-08-21
Have DirectoryToBagJob produce works instead of folders

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2731

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/normalize/DirectoryToBagJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/normalize/DirectoryToBagJob.java
@@ -83,8 +83,6 @@ public class DirectoryToBagJob extends AbstractFileServerToBagJob {
             Path filePath = sourcePath.getParent().relativize(file.toPath());
             String filePathString = filePath.toString();
             String filename = filePath.getFileName().toString();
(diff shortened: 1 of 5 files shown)


deposit/src/main/java/edu/unc/lib/deposit/normalize/DirectoryToBagJob.java
```diff
@@ -80,21 +81,32 @@ public void runJob() {
...
+
+            if (!file.isDirectory()) {
+                // Create work object
+                PID workPid = pidMinter.mintContentPid();
+                Bag workBag = model.createBag(workPid.getURI());
+                model.add(folderBag, RDF.type, Cdr.Work);
+                model.add(folderBag, CdrDeposit.label, filename);
 
-            if (!isDir) {
-                Resource fileResource = getFileResource(sourceBag, filePathString);
```
bbpennel, line 90: This is just occurring to me now, but it'd make sense to refactor `getFileResource` to generate a work which wraps the file. That way it continues to be useful, the Bagit job also gets to benefit from this change, and you wouldn't have to make much (or any) changes to the behavior here.
So you would basically want to move all this work construction and wrapping into `getFileResource`, which should already correctly be handling adding to the right parent folder.

PR #1088, opened 2020-09-01
BXC-2795 - Prevent deposit job blocking caused by TDB lock

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2795

* DepositModelManager is now used by both the deposit and migration-util, and was updated to support both use cases
* Deposit jobs which update a pre-populated deposit model now generally hold a read transaction, opening short lived write transactions per modification, or compiling all additions to be committed at once at the end
* Normalization jobs (that transform submissions into deposit models) now build an in memory model and then commit at the end to avoid long write transactions.
* Switches some deposit jobs that were unnecessarily using write models to using read models
* Now uses jena TDB2, which supersedes TDB1 and allows for reuse of models between transactions
* Tests mostly using in-memory TDB datasets, which significantly speeds them up


persistence/src/main/java/edu/unc/lib/dl/persist/services/deposit/DepositModelManager.java
```diff
@@ -0,0 +1,479 @@
...
+        }
+    }
+
+    /**
+     * Commit or abort changes in the dataset
+     * @param depositPid pid of the deposit
+     * @param dataset dataset to commit, or null, in which case the dataset will be retrieved
+     * @param abort if true, the commit will be aborted
+     */
+    public void commitOrAbort(PID depositPid, Dataset dataset, boolean abort) {
```
lfarrell, line 434: Should this variant of the method not worry about checking if the dataset is null and just assume it is? Then there's no need to pass it in "null".

persistence/src/main/java/edu/unc/lib/dl/persist/services/deposit/DepositModelManager.java
```diff
@@ -0,0 +1,479 @@
...
+     */
+    public DepositModelManager(String tdbBaseDir) {
+        this(Paths.get(tdbBaseDir));
+    }
+
+    /**
+     * Construct and initialize a deposit model manager
+     * @param depositsPtdbBasePathath
+     */
+    public DepositModelManager(Path tdbBasePath) {
```
lfarrell, line 85: JavaDoc and method signature don't match

bbpennel: Javadoc is butchered. Let me fix that

deposit/src/main/java/edu/unc/lib/deposit/normalize/BagIt2N3BagJob.java
```diff
@@ -32,11 +32,13 @@
 import java.util.Set;
 
 import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ModelFactory;
 import org.apache.jena.rdf.model.Property;
 import org.apache.jena.rdf.model.Resource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import edu.unc.lib.dl.util.RedisWorkerConstants.DepositField;
```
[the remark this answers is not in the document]
lfarrell: Answered my own question

PR #1090, opened 2020-09-03
Add progress counts for more job types

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2755


deposit/src/main/java/edu/unc/lib/deposit/validate/ExtractTechnicalMetadataJob.java
```diff
@@ -138,13 +138,16 @@ public void runJob() {
...
+                addClicks(1);
             } catch (JobFailedException | JobInterruptedException e) {
                 throw e;
             } catch (Exception e) {
                 failJob(e, "Failed to extract FITS details for {0} from document:\n{1}",
                         objPid, xmlOutputter.outputString(fitsDoc));
             }
         }
+
+        setTotalClicks(stagingList.size());
```
bbpennel, line 150: For some of the longer running jobs it'd be helpful to know in advance how much work there is total. 
There are some examples of precomputing the count, this is from the virus scan job:
```
List<Entry<PID, String>> hrefs = getPropertyPairList(model, CdrDeposit.stagingLocation);

        setTotalClicks(hrefs.size());
```

I think both the jobs you updated in this PR would be good candidates

static/js/admin/src/statusMonitor/DepositMonitor.js
```diff
@@ -105,18 +105,24 @@ define('DepositMonitor', [ 'jquery', 'jquery-ui', 'underscore', 'AbstractStatusM
...
 				var jobName = job.name.substring(job.name.lastIndexOf(".") + 1);
 				
 				job["shortName"] = jobName;
-				
-				if (jobName == "IngestContentObjectsJob" && "total" in job) {
-					var completion = results.ingestedObjects? results.ingestedObjects : 0;
-					completion += " / " + job.total;
-					job["completion"] = completion;
+
+				if (jobsWithCounts.includes(jobName) && "total" in job) {
```
bbpennel, line 124: Is there any reason to restrict which jobs we allow counts on? Seems like anything that has `num` set and greater than zero should be displayed. Also, I think its okay to display num even if total isn't set, just display without the second half

bbpennel (changes requested):
Can you also add clicks and total for TransferBinariesToStorageJob?

deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
```diff
@@ -94,6 +96,7 @@ public void runJob() {
         // All objects in deposit should have the same destination, so pull storage loc from deposit record
         try (BinaryTransferSession transferSession = getTransferSession(model)) {
             transferBinaries(depositBag, transferSession);
+            setTotalClicks(ingestedObjCount);
```
bbpennel, line 99: Hmm, with this approach the total count will shrink if you resume the job. It may be easiest to switch the count to be the number of objects being processed rather than the number of files. In that case you should be able to precompute the total (I believe counting the total number of subjects in the model will get you this), and the counts will be in line with most of the other jobs

static/js/admin/src/statusMonitor/DepositMonitor.js
```diff
@@ -105,18 +105,19 @@ define('DepositMonitor', [ 'jquery', 'jquery-ui', 'underscore', 'AbstractStatusM
...
 				var jobName = job.name.substring(job.name.lastIndexOf(".") + 1);
 				
 				job["shortName"] = jobName;
-				
-				if (jobName == "IngestContentObjectsJob" && "total" in job) {
-					var completion = results.ingestedObjects? results.ingestedObjects : 0;
-					completion += " / " + job.total;
-					job["completion"] = completion;
+
+				var hasCount = job.num > 0;
```
bbpennel, line 116: This section looks good.

While testing it, I was finding with so many additional progress numbers showing it was getting a bit hard to differentiate between all the numbers next to each other. Adding some text to split up the progress from the time spent helps a little. Changing the following line:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/static/templates/admin/statusMonitor/depositMonitorJobDetails.html#L30
to put "in" between progress and time:
`<li><span title="<%- job.shortName %>"><%= job.shortName %></span> <%= job.status %> <%= job.completion %> <% if (job.time) { %> in <%= job.time %> <% } %></li>`

bbpennel (changes requested):

deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
```diff
@@ -91,6 +92,15 @@ public void runJob() {
         Model model = getReadOnlyModel();
         Bag depositBag = model.getBag(getDepositPID().getRepositoryPath());
 
+        // Count how many objects are being deposited
+        int i = 0;
+        ResIterator subjectIterator = model.listSubjects();
```
bbpennel, line 97: I don't think the click counts will line up with the total anymore, since clicks are still counting number of files while the total is the number of resources. I believe you'll need to move the `addClicks` call into `transferBinaries` since that iterates over each resource in the model.

deposit/src/test/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJobTest.java
```diff
@@ -366,7 +366,7 @@ public void depositResumeFailed() throws Exception {
         assertFitsFileTransferred(postFileResc2);
 
         verify(jobStatusFactory, times(2)).setTotalCompletion(eq(jobUUID), eq(5));
-        verify(jobStatusFactory, times(4)).incrCompletion(eq(jobUUID), eq(1));
```
bbpennel: I think you might have to reset the click count between runs, otherwise the count is going to be higher than the total after a resume. Not sure if trying to skip `addClick` during a resume would be practical, but maybe its possible?

deposit/src/test/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJobTest.java
```diff
@@ -268,6 +268,8 @@ public void depositWithMultipleFilesResumed() throws Exception {
         // First, incomplete run
         job.run();
 
+        reset(repoObjFactory);
```
bbpennel, line 271: Do you need this reset?

bbpennel (approved):

PR #1091, opened 2020-09-11
Cleanup derivatives after object destroyed

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2506
Depends on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/82

### DIFF
--- a/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
+++ b/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
@@ -423,6 +423,9 @@ public class DestroyObjectsJobIT {
             Map<String, String> cleanupFile = filesToCleanup.get(uri);
             assertEquals(info.getChildTextTrim("pidId", CDR_MESSAGE_NS), cleanupFile.get("pid"));
 
+            String uuid = cleanupFile.get("uuid");
+            assertEquals(info.getChildTextTrim("uuid", CDR_MESSAGE_NS), uuid);
+
             String objType = cleanupFile.get("objType");
             assertEquals(info.getChildTextTrim("objType", CDR_MESSAGE_NS), objType);
 
@@ -446,7 +449,9 @@ public class DestroyObjectsJobIT {
     private Map<String, String> nonBinaryToCleanup(RepositoryObject obj) {
         Map<String, String> contentMetadata = new HashMap<>();
         contentMetadata.put("objType", ResourceType.getResourceTypeForUris(obj.getTypes()).getUri());
-        contentMetadata.put("pid", obj.getPid().getQualifiedId());
+        PID pid = obj.getPid();
+        contentMetadata.put("pid", pid.getQualifiedId());
+        contentMetadata.put("uuid", pid.getUUID());
         return contentMetadata;
     }
 
@@ -454,9 +459,11 @@ public class DestroyObjectsJobIT {
         HashMap<URI, Map<String, String>> cleanupBinaryUris = new HashMap<>();
 
         BinaryObject binary = fileObj.getOriginalFile();
+        PID binaryPid = binary.getPid();
         Map<String, String> contentMetadata = new HashMap<>();
         contentMetadata.put("objType", Cdr.FileObject.getURI());
-        contentMetadata.put("pid", binary.getPid().getQualifiedId());
+        contentMetadata.put("pid", binaryPid.getQualifiedId());
+        contentMetadata.put("uuid", binaryPid.getUUID());
         contentMetadata.put("mimeType", binary.getMimetype());
         cleanupBinaryUris.put(binary.getContentUri(), contentMetadata);
 
--- a/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
+++ b/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
@@ -222,7 +222,7 @@ public class DestroyObjectsJobIT {
         PID folderObjPid = objsToDestroy.get(0);
 
         FileObject fileObj = repoObjLoader.getFileObject(fileObjPid);
-        ArrayList<Map<String, String>> binaryDerivs = derivativesToCleanup(fileObj.getBinaryObjects());
+        List<Map<String, String>> binaryDerivs = derivativesToCleanup(fileObj.getBinaryObjects());
         URI contentUri = fileObj.getOriginalFile().getContentUri();
 
         job.run();
@@ -403,8 +403,8 @@ public class DestroyObjectsJobIT {
         }
     }
 
-    private ArrayList<Map<String, String>> derivativesToCleanup(List<BinaryObject> binaries) {
-        ArrayList<Map<String, String>> cleanupBinaryUris = new ArrayList<>();
+    private List<Map<String, String>> derivativesToCleanup(List<BinaryObject> binaries) {
+        List<Map<String, String>> cleanupBinaryUris = new ArrayList<>();
 
         for (BinaryObject binary : binaries) {
             Map<String, String> contentMetadata = new HashMap<>();
@@ -417,7 +417,7 @@ public class DestroyObjectsJobIT {
         return cleanupBinaryUris;
     }
 
-    private void assertDerivativeDestroyMsgSent(ArrayList<Map<String, String>> binary) {
+    private void assertDerivativeDestroyMsgSent(List<Map<String, String>> binary) {
         binary.forEach(b -> verify(binaryDestroyDerivativesMessageSender)
                 .sendRemoveDerivativesOperation(agent.getUsername(), PIDs.get(b.get("pid")),
                         b.get("mimeType")));
--- a/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
+++ b/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
@@ -319,8 +319,11 @@ public class DestroyObjectsJobIT {
         verify(indexingMessageSender).sendIndexingOperation(anyString(), eq(folderObjPid), eq(DELETE_SOLR_TREE));
 
         verify(binaryDestroyedMessageSender, times(2)).sendMessage(docCaptor.capture());
-        Document returnedMsgDoc = docCaptor.getValue();
-        assertMessagesEqual(returnedMsgDoc, binaryDerivs);
+
+        List<Document> returnedMsgDocs = docCaptor.getAllValues();
+        for (Document returnedMsgDoc : returnedMsgDocs) {
+            assertMessagesEqual(returnedMsgDoc, binaryDerivs);
+        }
     }
 
     @Test
@@ -403,7 +406,6 @@ public class DestroyObjectsJobIT {
 
     private void assertMessagesEqual(Document returnedDoc,
                                  Map<URI, Map<String, String>> binaryDerivs) {
-
         binaryDerivs.forEach((contentUri, metadata) -> {
             Document destroyMsg = makeDestroyOperationBody(agent.getUsername(), contentUri, metadata);
             assertEquals(returnedDoc.getRootElement().toString(), destroyMsg.getRootElement().toString());
--- a/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
+++ b/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
@@ -177,6 +177,7 @@ public class DestroyObjectsJobIT {
         initializeJob(asList(fileObjPid));
 
         FileObject fileObj = repoObjLoader.getFileObject(fileObjPid);
+        List<Map<String, String>> binaries = derivativesToCleanup(fileObj.getBinaryObjects());
 
         URI contentUri = fileObj.getOriginalFile().getContentUri();
         assertTrue(Files.exists(Paths.get(contentUri)));
@@ -203,11 +204,10 @@ public class DestroyObjectsJobIT {
         verify(indexingMessageSender).sendIndexingOperation(anyString(), eq(fileObjPid), eq(DELETE_SOLR_TREE));
         verify(binaryDestroyedMessageSender).sendMessage(contentUri.toString());
 
-        List<BinaryObject> binaries = fileObj.getBinaryObjects();
-        for (BinaryObject binary: binaries) {
+        for (Map<String, String> binary: binaries) {
             verify(binaryDestroyDerivativesMessageSender)
-                    .sendRemoveDerivativesOperation(agent.getUsername(), binary.getPid(),
-                    binary.getMimetype());
+                    .sendRemoveDerivativesOperation(agent.getUsername(), PIDs.get(binary.get("pid")),
+                    binary.get("mimeType"));
         }
     }
 
--- a/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
+++ b/persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
@@ -17,14 +17,17 @@ package edu.unc.lib.dl.persist.services.destroy;
 
 import static edu.unc.lib.dl.fcrepo4.RepositoryPaths.getContentRootPid;
 import static edu.unc.lib.dl.rdf.CdrAcl.markedForDeletion;
+import static edu.unc.lib.dl.services.DestroyObjectsMessageHelpers.makeDestroyOperationBody;
 import static edu.unc.lib.dl.sparql.SparqlUpdateHelper.createSparqlReplace;
 import static edu.unc.lib.dl.util.IndexingActionType.DELETE_SOLR_TREE;
 import static java.util.Arrays.asList;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.initMocks;
@@ -36,6 +39,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -46,11 +50,14 @@ import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.Resource;
 import org.apache.jena.vocabulary.RDF;
 import org.fcrepo.client.FcrepoClient;
+import org.jdom2.Document;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
 import org.mockito.Mock;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ContextConfiguration;
@@ -67,7 +74,6 @@ import edu.unc.lib.dl.fcrepo4.CollectionObject;
 import edu.unc.lib.dl.fcrepo4.ContentRootObject;
 import edu.unc.lib.dl.fcrepo4.FileObject;
 import edu.unc.lib.dl.fcrepo4.FolderObject;
-import edu.unc.lib.dl.fcrepo4.PIDs;
 import edu.unc.lib.dl.fcrepo4.RepositoryInitializer;
 import edu.unc.lib.dl.fcrepo4.RepositoryObject;
 import edu.unc.lib.dl.fcrepo4.RepositoryObjectFactory;
@@ -86,7 +92,6 @@ import edu.unc.lib.dl.search.solr.model.ObjectPath;
 import edu.unc.lib.dl.search.solr.service.ObjectPathFactory;
 import edu.unc.lib.dl.services.IndexingMessageSender;
 import edu.unc.lib.dl.services.MessageSender;
-import edu.unc.lib.dl.services.OperationsMessageSender;
 import edu.unc.lib.dl.sparql.SparqlUpdateService;
 import edu.unc.lib.dl.test.AclModelBuilder;
 import edu.unc.lib.dl.test.RepositoryObjectTreeIndexer;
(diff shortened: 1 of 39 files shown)


services-camel/src/main/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+                }
+                log.info("Derivative and parent directories destroyed for {}", binaryId);
+            }
+        } catch (IOException e) {
+            log.warn("Unable to destroy derivative and parent directories for {}: {}", binaryId, e.getMessage());
+        }
+    }
+
+    private boolean shouldRemoveFile(Path path) throws IOException {
+        List<String> rootDirs = Arrays.asList(
```
bbpennel, line 82: There's a bit of a disconnect between having the full list of derivatives hard coded in here while initializing a separate processor per type. I'd suggest either making the DestroyDerivativeProcessor able to figure out all the derivatives that need to be destroyed for a particular object (basically internalizing the stuff from the router), or reuse `derivativeBasePath` to determine if a file should be removed. It seems like anything that starts with `derivativeBasePath` but is not exactly it can be deleted if its empty.

services-camel/src/main/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouter.java
```diff
@@ -0,0 +1,79 @@
...
+    @BeanInject(value = "destroyLargeThumbnailProcessor")
+    private DestroyDerivativesProcessor destroyLargeThumbnailProcessor;
+
+    @BeanInject(value = "destroyAccessCopyProcessor")
+    private DestroyDerivativesProcessor destroyAccessCopyProcessor;
+
+    @BeanInject(value = "destroyFulltextProcessor")
+    private DestroyDerivativesProcessor destroyFulltextProcessor;
+
+    private static final String IMAGE_MIMETYPE_PATTERN = "^(image.*$|application.*?(photoshop|psd)$)";
```
bbpennel, line 46: This regex already exists in `ImageDerivativeProcessor`. I would suggest changing `ImageDerivativeProcessor.allowedImageType` to static and then using that in the conditional below rather than having a new regex here. The method actually runs two regexes as well.

services-camel/src/main/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouter.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final String TEXT_MIMETYPE_PATTERN = "^(text/|application/pdf|application/msword"
```
bbpennel, line 47: I'd suggest making FulltextRouter and this router use the same constant for this regex. You could centralize it in the FulltextProcessor so it follows a similar pattern to the image enhancements.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessorTest.java
```diff
@@ -0,0 +1,158 @@
...
+        derivativeDirBase = derivativeDir.getRoot().getAbsolutePath();
+    }
+
+    @Test
+    public void deleteFulltextTest() throws Exception {
+        derivativeTypeDir = FULLTEXT_EXTRACTION.getId();
+        derivativeFinalDir = derivativeDir.newFolder(derivativeTypeDir, "de", "75", "d8", "11");
+        file = new File(derivativeFinalDir, pathId + ".txt");
+
+        try (BufferedWriter writeFile = new BufferedWriter(new FileWriter(file))) {
```
bbpennel, line 87: You can use `FileUtils.writeStringToFile(file, "my text", StandardCharsets.UTF_8);`

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessorTest.java
```diff
@@ -0,0 +1,158 @@
...
+            writeFile.write("my text");
+        }
+
+        derivativeTypeBaseDir = new File(derivativeDirBase + "/" + derivativeTypeDir);
```
bbpennel, line 91:
```suggestion
        derivativeTypeBaseDir = new File(derivativeDirBase, derivativeTypeDir);
```

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessorTest.java
```diff
@@ -0,0 +1,158 @@
...
+        processor = new DestroyDerivativesProcessor(".txt", derivativeDirBase + "/" + derivativeTypeDir);
```
bbpennel, line 92:
```suggestion
        processor = new DestroyDerivativesProcessor(".txt", derivativeTypeBaseDir.getAbsolutePath());
```

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessorTest.java
```diff
@@ -0,0 +1,158 @@
...
+        // Deleted file
+        assertFalse(file.exists());
+        // Deleted parent dirs
+        assertFalse(new File(derivativeTypeBaseDir, "de").exists());
+        // Didn't delete root derivative type dir
+        assertTrue(derivativeTypeBaseDir.exists());
+    }
+
+    @Test
+    public void deleteThumbnailTest() throws Exception {
```
bbpennel, line 108: See the same suggestions as `deleteFulltextTest`

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessorTest.java
```diff
@@ -0,0 +1,158 @@
...
+        // Deleted file
+        assertFalse(file.exists());
+        // Deleted parent dirs
+        assertFalse(new File(derivativeTypeBaseDir, "de").exists());
+        // Didn't delete root derivative type dir
+        assertTrue(derivativeTypeBaseDir.exists());
+    }
+
+    @Test
+    public void deleteJp2Test() throws Exception {
```
bbpennel, line 134: Same suggestions as `deleteFulltextTest`

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessorTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.camel.Message;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.mockito.Mock;
+
+import edu.unc.lib.dl.fcrepo4.PIDs;
+
+public class DestroyDerivativesProcessorTest {
```
bbpennel, line 44: Could you add a test or two to verify that occupied directories don't get deleted? You could have one where there are two directories in "de", and/or one where there is another file in `derivativeTypeDir, "de", "75", "d8", "11"`.

Also, it'd be good to verify that nothing happens or fails if a file that is neither an image or text is destroyed.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    protected AbstractApplicationContext createApplicationContext() {
+        return new ClassPathXmlApplicationContext("/service-context.xml", "/destroy-derivatives-context.xml");
+    }
+
+    @Test
+    public void routeRequestText() throws Exception {
+        createContext(DESTROY_DERIVATIVES_ROUTE);
+        template.sendBodyAndHeaders("", createEvent("text/plain"));
+        assertMockEndpointsSatisfied();
```
bbpennel, line 65: Its not actually making any assertions about the endpoints. Normally it would specify what to expect with something like `getMockEndpoint("mock:direct:solr-update").expectedMessageCount(1);` if you have specify expectations about messages being sent somewhere.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterTest.java
```diff
@@ -0,0 +1,109 @@
...
+        });
+
+        context.start();
+    }
+
+    private static Map<String, Object> createEvent(String mimetype) {
+        final Map<String, Object> headers = new HashMap<>();
+        headers.put(CdrBinaryMimeType, mimetype);
+        headers.put(CdrBinaryPidId, "dee2614c-8a4b-4ac2-baf2-4b4afc11af87");
+       // http://localhost:8181/fcrepo/rest/content/de/e2/61/4c/dee2614c-8a4b-4ac2-baf2-4b4afc11af87/datafs/original_file
```
bbpennel, line 105: Can remove this line

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    public void routeRequestImage() throws Exception {
+        createContext(DESTROY_DERIVATIVES_ROUTE);
+        template.sendBodyAndHeaders("", createEvent("image/png"));
+        assertMockEndpointsSatisfied();
+    }
+
+    @Test
+    public void destroyTextDerivative() throws Exception {
```
bbpennel, line 76: Seems like in most of these tests you'd want to make sure the processors are called?

bbpennel (changes requested):
It would be helpful to have an Integration test for destroying objects, largely to make sure the plumbing between the destroy routes and all the stuff that comes after it fits together, particularly in terms of inputs. If that seems a bit too large for this ticket we would make a separate one, up to you.

bbpennel (changes requested):

bbpennel (changes requested):

persistence/src/test/java/edu/unc/lib/dl/persist/services/destroy/DestroyObjectsJobIT.java
```diff
@@ -296,7 +317,10 @@ public void destroyFolderTest() {
         assertTrue(folderResc.hasProperty(RDF.type, Cdr.Tombstone));
 
         verify(indexingMessageSender).sendIndexingOperation(anyString(), eq(folderObjPid), eq(DELETE_SOLR_TREE));
-        verify(binaryDestroyedMessageSender).sendMessage(contentUri.toString());
+
+        verify(binaryDestroyedMessageSender, times(2)).sendMessage(docCaptor.capture());
```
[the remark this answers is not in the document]
lfarrell: Hmm, I think this is because it's not checking for members so it only gets the folder and file. Seems bad.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterIT.java
```diff
@@ -0,0 +1,345 @@
...
+        initMarkForDeletionJob(fileObj.getPid());
+        deletionJob.run();
+        initializeDestroyJob(Collections.singletonList(fileObj.getPid()));
+        destroyJob.run();
+
+        verify(destroySmallThumbnailProcessor).process(any(Exchange.class));
+        verify(destroyLargeThumbnailProcessor).process(any(Exchange.class));
+        verify(destroyCollectionSrcImgProcessor, never()).process(any(Exchange.class));
+        verify(destroyAccessCopyProcessor).process(any(Exchange.class));
+        verify(destroyCollectionSrcImgProcessor, never()).process(any(Exchange.class));
```
bbpennel, line 218: you can remove this duplicate check (same check is at line 216)

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterIT.java
```diff
@@ -0,0 +1,345 @@
...
+        CollectionObject collectionWithImg = repoObjectFactory.createCollectionObject(null);
+        adminUnit.addMember(collectionWithImg);
+
+        treeIndexer.indexAll(baseAddress);
+
+        // Create collection thumbnail
+        PID collPid = collectionWithImg.getPid();
+        String uuid = collPid.getUUID();
+        String binarySubPath = idToPath(uuid, HASHED_PATH_DEPTH, HASHED_PATH_SIZE);
+        File existingFile = new File("target/" + binarySubPath + "/" + uuid);
```
bbpennel, line 233: please use:
```
@Rule
    public final TemporaryFolder tmpFolder = new TemporaryFolder();
```
to create a temp directory and put the file into there. The target directory isn't too bad, but won't get automatically cleaned up (in case the test fails) without a `mvn clean`

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterIT.java
```diff
@@ -0,0 +1,345 @@
...
+
+        verify(destroySmallThumbnailProcessor, never()).process(any(Exchange.class));
+        verify(destroyLargeThumbnailProcessor, never()).process(any(Exchange.class));
+        verify(destroyCollectionSrcImgProcessor, never()).process(any(Exchange.class));
+        verify(destroyAccessCopyProcessor, never()).process(any(Exchange.class));
+        verify(destroyFulltextProcessor, never()).process(any(Exchange.class));
+    }
+
+    private FileObject addFileToWork(WorkObject work, String mimetype) throws Exception {
+        FolderObject folder = repoObjectFactory.createFolderObject(null);
```
bbpennel, line 309: Its a bit surprising that `addFileToWork` creates a folder, is that necessary? Works can be the children of collections

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterIT.java
```diff
@@ -0,0 +1,345 @@
...
+        File contentFile = Files.createTempFile(storagePath, "file", ".txt").toFile();
+        String sha1 = "4f9be057f0ea5d2ba72fd2c810e8d7b9aa98b469";
+        String filename = contentFile.getName();
+        FileUtils.writeStringToFile(contentFile, bodyString, "UTF-8");
+
+        return work.addDataFile(contentFile.toURI(), filename, mimetype, sha1, null);
+    }
+
+    private void initMarkForDeletionJob(PID pid) {
+        deletionJob =  new MarkForDeletionJob(pid, "", agent, repoObjLoader,
```
bbpennel, line 325: It'd probably be fine to save a step by changing this method to `markObjectForDeletion` and have it immediately run the job. Then you could also remove the `deletionJob` instance variable

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterIT.java
```diff
@@ -0,0 +1,345 @@
...
+                sparqlUpdateService, aclService);
+    }
+
+    private void initializeDestroyJob(List<PID> objsToDestroy) {
```
bbpennel, line 329: Totally optional, but if you don't want to have to initialize lists in order to call this method, you can do `initializeDestroyJob(PID... objsToDestroy)` in which case you'll get an array of PIDs, which you can stream with `Arrays.stream(array)`. To call it you do `initializeDestroyJob(pid1, pid2, etc);`

You could probably also make the method into `destroyObject` and immediately execute the job.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+    @Test
+    public void destroyImageSrcCopyCollection() throws Exception {
+        createContext(DESTROY_SRC_COPY);
+
+        Map<String, Object> headers = createEvent("image/*", Cdr.Collection.getURI());
+        headers.put("CollectionThumb", true);
+        template.sendBodyAndHeaders("", headers);
+
+        verify(destroySmallThumbnailProcessor, never()).process(any(Exchange.class));
```
bbpennel, line 155: since you're directly testing the `CdrDestroyCollectionUpload` route here, it would be fine to only verify `destroyCollectionSrcImgProcessor` is called. The other processors aren't related to this route, and it confused me initially that the thumbnail processors weren't called for cleanup of a collection thumbnail.

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterTest.java
```diff
@@ -0,0 +1,220 @@
...
+        assertMockEndpointsSatisfied();
+    }
+
+    @Test
+    public void destroyImageAccessDerivative() throws Exception {
+        createContext(DESTROY_ACCESS_COPY_ROUTE);
+
+        template.sendBodyAndHeaders("", createEvent("image/png", Cdr.FileObject.getURI()));
+
+        verify(destroySmallThumbnailProcessor, never()).process(any(Exchange.class));
```
bbpennel, line 181: Similar to the collection thumbnail case, I'd say generally for sub-route tests like this you don't need to verify processors that aren't associated with that route

services-camel/src/test/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesRouterTest.java
```diff
@@ -0,0 +1,220 @@
...
+        verify(destroyLargeThumbnailProcessor, never()).process(any(Exchange.class));
+        verify(destroyAccessCopyProcessor).process(any(Exchange.class));
+        verify(destroyFulltextProcessor, never()).process(any(Exchange.class));
+    }
+
+    // See if any message is routed from direct:image.access.destroy
+    @Test
+    public void destroyImageAccessDerivativeCollection() throws Exception {
```
bbpennel, line 189: Could this test be merged with `destroyImageThumbnailDerivativeCollection`?

services-camel/src/main/java/edu/unc/lib/dl/services/camel/longleaf/LongleafRouter.java
```diff
@@ -28,6 +28,8 @@
  * @author bbpennel
  */
 public class LongleafRouter extends RouteBuilder {
+    @BeanInject(value = "getUrisProcessor")
+    private GetUrisProcessor getUrisProcessor;
```
bbpennel, line 32: Up to you if you want to change it, but `GetUrisProcessor` doesn't have any dependencies and is never mocked, so it'd be fine to initialize it directly in this class, either in the constructor or in the configure method.

services-camel/src/test/resources/destroy-derivatives-router-it-context.xml
```diff
@@ -0,0 +1,95 @@
...
+    </bean>
+
+    <bean id="contentPathFactory" class="edu.unc.lib.dl.fedora.ContentPathFactory"
+          init-method="init">
+        <property name="cacheMaxSize" value="100" />
+        <property name="cacheTimeToLive" value="5000" />
+        <property name="fcrepoClient" ref="fcrepoClient" />
+    </bean>
+
+    <bean id="objectAclFactory" class="edu.unc.lib.dl.acl.fcrepo4.ObjectAclFactory"
```
bbpennel, line 65: I think you can remove `objectAclFactory` and `inheritedPermissionEvaluator` since they are already being initialized in `acl-service-context.xml`

bbpennel (changes requested):

services-camel/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -219,6 +219,35 @@
     <bean id="fulltextProcessor" class="edu.unc.lib.dl.services.camel.fulltext.FulltextProcessor">
         <constructor-arg value="${cdr.enhancement.path.fulltext}" />
     </bean>
+
+    <bean id="binaryInfoProcessor" class="edu.unc.lib.dl.services.camel.destroyDerivatives.DestroyedMsgProcessor">
```
bbpennel, line 223: update name of the bean to match, the application wouldn't start for me until i did so 
```suggestion
    <bean id="destroyedMsgProcessor" class="edu.unc.lib.dl.services.camel.destroyDerivatives.DestroyedMsgProcessor">
```

bbpennel (changes requested):
After I fix the name of the bean and the name of the route in the config PR, I get a null pointer exception while destroying a text file. I'm guessing its getting null as the mimetype, which causes the regex to choke. So you'll probably need to test on your VM to make sure the mimetype is coming through. It may be that `allowedImageType` (and the one for images) should also handle null mimetypes without dying
```
java.lang.NullPointerException: null
        at java.util.regex.Matcher.getTextLength(Matcher.java:1283)
        at java.util.regex.Matcher.reset(Matcher.java:309)
        at java.util.regex.Matcher.<init>(Matcher.java:229)
        at java.util.regex.Pattern.matcher(Pattern.java:1093)
        at edu.unc.lib.dl.services.camel.images.ImageDerivativeProcessor.allowedImageType(ImageDerivativeProcessor.java:55)
```

services-camel/src/main/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessor.java
```diff
@@ -52,8 +53,10 @@ public DestroyDerivativesProcessor(String fileExtension, String derivativeBasePa
     public void process(Exchange exchange) {
         Message in = exchange.getIn();
         String binaryId = (String) in.getHeader(CdrBinaryPidId);
-        String binarySubPath = idToPath(binaryId, HASHED_PATH_DEPTH, HASHED_PATH_SIZE);
-        Path derivativePath = Paths.get(derivativeBasePath, binarySubPath, binaryId + fileExtension);
+        String binaryUUID = (String) in.getHeader(CdrBinaryUUID);
```
bbpennel, line 56: I'd really prefer not to add an extra UUID field/header. We've been gradually phasing out `pid.getUuid()` in favor of `pid.getId()`, since the former just calls the latter, and neither necessarily returns a UUID. You can also get the ID/UUID portion back from the PID field, which it looks like you were doing before in DestroyedMsgProcessor. Was there a reason you needed to switch? I'd suggest:
```
String binaryPidId = (String) in.getHeader(CdrBinaryPidId);
PID binaryPid = PIDs.get(binaryPidId);
String binaryId = binaryPid.getId();
```

bbpennel (changes requested):

services-camel/src/main/java/edu/unc/lib/dl/services/camel/destroyDerivatives/DestroyDerivativesProcessor.java
lfarrell: IdToPath was trying to generate the derivative base path from /content/uuuid. So you'd end up with /co/nt/en etc, which I think is why it wasn't actually deleting things.

bbpennel (approved):

PR #1094, opened 2020-09-30
BXC-2803: Switch ingest file transfers from move to copy

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2803

### DIFF
--- a/persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClient.java
+++ b/persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClient.java
@@ -108,16 +108,6 @@ public class FSToFSTransferClient implements BinaryTransferClient {
                 // Ignore. New file is already in place
                 log.warn("Unable to delete {}. Reason {}", oldFilePath, e.getMessage());
             }
-
-            // Remove source file after deposited
-            if (!source.isReadOnly()) {
-                Path sourcePath = sourceFile.toPath();
-                try {
-                    Files.deleteIfExists(sourcePath);
-                } catch (IOException e) {
-                    log.warn("Unable to delete source file {}. Reason {}", sourcePath, e.getMessage());
-                }
-            }
         } catch (IOException e) {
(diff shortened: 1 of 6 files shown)


persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/StreamToFSTransferClient.java
```diff
@@ -98,7 +98,7 @@ protected URI writeStream(PID binPid, InputStream sourceStream, boolean allowOve
             }
 
             // Move temp file into final location
-            Files.move(newFilePath, destPath, REPLACE_EXISTING);
```
bbpennel, line 101: we still want to use move for renaming the temp/old files since it is within the same storage location, its just for source files (which doesn't exist for this client) that we want to switch to copy

persistence/src/main/java/edu/unc/lib/dl/persist/services/transfer/FSToFSTransferClient.java
```diff
@@ -112,6 +108,16 @@ public URI transfer(PID binPid, URI sourceFileUri, boolean allowOverwrite) {
                 // Ignore. New file is already in place
                 log.warn("Unable to delete {}. Reason {}", oldFilePath, e.getMessage());
             }
+
+            // Remove source file after deposited
+            if (!source.isReadOnly()) {
```
bbpennel, line 113: We don't want to perform cleanup in here, the code calling the transfer client should now be responsible for cleanup of the source file. For instance, with deposits the cleanup deposit job should handle this (you should verify that it does so after these changes. To test on your vm I suggest setting the cleanup time property for the deposit application to something very low, like 20second). The intention of switching to copy is so that we don't lose the source data in the case that the larger operation fails.

You should also check if there are other uses of this `transfer` method from non-deposit related activities, since they should now be doing their own source file cleanup.

bbpennel (changes requested):

bbpennel (approved):

PR #1095, opened 2020-10-02
Only allow file server ingests at the collection level

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2805


static/js/admin/src/AddMenu.js
```diff
@@ -34,7 +33,7 @@ define('AddMenu', [ 'jquery', 'jquery-ui', 'underscore', 'CreateContainerForm',
 				items["addFile"] = { name : "Add File" };
 			}
 			
-			if (!workOrFile) {
+			if (this.container.type === "Collection") {
```
bbpennel, line 36: the option for be available for Folders and Collections, so you can probably just use `folderOrCollection`

bbpennel (changes requested):
See comment

bbpennel (approved):

PR #1099, opened 2020-10-13
Add delay before retrying ingests due to checksum failure

lfarrell:


deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -396,6 +396,15 @@ private FileObject addFileToWork(WorkObject work, Resource childResc)
         // Retry if there are checksum failures
         for (int retryCnt = 1; retryCnt <= CHECKSUM_RETRIES; retryCnt++) {
             try {
+                if (retryCnt > 1) {
```
bbpennel, line 399: might be simpler to perform the delay in the `ChecksumMismatchException` catch, it already has a block to check if retries are remaining and the automatic result would be there wouldn't be a delay until after the first try.

deposit/src/main/java/edu/unc/lib/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -396,6 +396,15 @@ private FileObject addFileToWork(WorkObject work, Resource childResc)
...
+                    try {
+                        Thread.sleep(retryCnt * 1000);
+                    } catch (InterruptedException e) {
+                        log.debug("Retry ingest pause for {} due to a checksum mismatch, {} retries remaining: {}",
```
bbpennel, line 403: rather than eating the interrupt, could you rethrow it as a `JobInterruptedException`? Shouldn't need an extra log statement in that case.

bbpennel (changes requested):

bbpennel (approved):

PR #1103, opened 2020-10-14
Check patron roles instead of princ roles when adding patron-settings tag

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2815

### DIFF
--- a/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
+++ b/solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
@@ -35,11 +35,11 @@ import edu.unc.lib.dl.acl.util.RoleAssignment;
 import edu.unc.lib.dl.acl.util.UserRole;
 import edu.unc.lib.dl.data.ingest.solr.exception.IndexingException;
 import edu.unc.lib.dl.data.ingest.solr.indexing.DocumentIndexingPackage;
+import edu.unc.lib.dl.fcrepo4.AdminUnit;
 import edu.unc.lib.dl.fcrepo4.ContentObject;
+import edu.unc.lib.dl.fcrepo4.CollectionObject;
 import edu.unc.lib.dl.fedora.PID;
-import edu.unc.lib.dl.rdf.Cdr;
 import edu.unc.lib.dl.search.solr.util.FacetConstants;
-import edu.unc.lib.dl.util.ResourceType;
 
 /**
  * Sets access-related status tags
@@ -108,43 +108,35 @@ public class SetAccessStatusFilter implements IndexDocumentFilter {
             status.add(FacetConstants.PUBLIC_ACCESS);
         }
 
-        if (addPatronSettings(dip.getContentObject())) {
+        if (hasPatronSettings(dip.getContentObject())) {
             status.add(FacetConstants.PATRON_SETTINGS);
         }
 
         return status;
     }
 
-    private boolean addPatronSettings(ContentObject contentObj) {
-        List<String> objTypes = contentObj.getTypes();
-        ResourceType objType = ResourceType.getResourceTypeForUris(objTypes);
-        if (objType == null || objType.equals(Cdr.AdminUnit.getURI())) {
+    private boolean hasPatronSettings(ContentObject contentObj) {
+        if (contentObj instanceof AdminUnit) {
             return false;
         }
 
-        return hasPatronSettings(contentObj.getPid(), objType);
-    }
-
-    private boolean hasPatronSettings(PID pid, ResourceType rescType) {
-        List<RoleAssignment> objPatronRoles = objAclFactory.getPatronRoleAssignments(pid);
+        List<RoleAssignment> objPatronRoles = objAclFactory.getPatronRoleAssignments(contentObj.getPid());
         int numRoles = objPatronRoles.size();
-        boolean isCollection = rescType.equals(Cdr.Collection.getLocalName());
+        boolean isCollection = contentObj instanceof CollectionObject;
 
-        if ((!isCollection && numRoles == 0)) {
+        if (!isCollection && numRoles == 0) {
             return false;
         } else if (isCollection && numRoles < 2) {
             return true;
         }
 
-        String canViewOrig = canViewOriginals.getURI().toString();
-        UserRole roleOne = objPatronRoles.get(0).getRole();
-
-        if (numRoles > 1) {
-            UserRole roleTwo = objPatronRoles.get(1).getRole();
-            return (!roleOne.equals(canViewOrig) || !roleTwo.equals(canViewOrig));
+        for (RoleAssignment role : objPatronRoles) {
+            if (!role.getRole().equals(canViewOriginals.getURI().toString())) {
+                return true;
+            }
         }
 
-        return !roleOne.equals(canViewOrig);
+        return false;
     }
 
     private boolean allPatronsRevoked(Map<String, Set<String>> objPrincRoles) {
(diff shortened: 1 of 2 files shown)


solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -105,7 +106,7 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
             status.add(FacetConstants.PUBLIC_ACCESS);
         }
 
-        if (objPrincRoles.size() > 0) {
+        if (objPatronRoles.size() > 0) {
```
bbpennel, line 109: After checking in with Julie about this, the tag should only appear if the patron permissions are set to something other than the default. 

For folders/works/files, the default would be either `objPatronRoles.size() == 0` OR the `authenticated` and `everyone` groups both set to `canViewOriginals`.

For collections, it has to work a little different, we would show `patron-settings` if `objPatronRoles.size() < 2`, or either`authenticated` or `everyone` groups had any value other than `canViewOriginals`.

And we can skip this tag altogether for Units.

solr-ingest/src/test/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilterTest.java
```diff
@@ -17,6 +17,7 @@
 
 import static edu.unc.lib.dl.acl.util.AccessPrincipalConstants.AUTHENTICATED_PRINC;
 import static edu.unc.lib.dl.acl.util.AccessPrincipalConstants.PUBLIC_PRINC;
+import static java.util.Arrays.asList;
```
bbpennel, line 20: Will probably want to add a test or two to verify that collections get the right patron-settings tag

bbpennel (changes requested):

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -105,13 +108,45 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
...
-        if (objPrincRoles.size() > 0) {
+        if (addPatronSettings(dip.getContentObject())) {
             status.add(FacetConstants.PATRON_SETTINGS);
         }
 
         return status;
     }
 
+    private boolean addPatronSettings(ContentObject contentObj) {
+        List<String> objTypes = contentObj.getTypes();
```
bbpennel, line 119: you should do `contentObj instanceof AdminUnit` instead of retrieving the types and comparing them

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -105,13 +108,45 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
...
+            return false;
+        }
+
+        return hasPatronSettings(contentObj.getPid(), objType);
+    }
+
+    private boolean hasPatronSettings(PID pid, ResourceType rescType) {
+        List<RoleAssignment> objPatronRoles = objAclFactory.getPatronRoleAssignments(pid);
+        int numRoles = objPatronRoles.size();
+        boolean isCollection = rescType.equals(Cdr.Collection.getLocalName());
```
bbpennel, line 131: you can pass the contentObj in instead of rescType, then do `contentObj instanceof CollectionObject`. Might make sense at some point to add a method to ContentObject to getResourceType or something, which would return the ResourceType. But we probably don't need that here.

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -105,13 +108,45 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
...
 
-        if (objPrincRoles.size() > 0) {
+        if (addPatronSettings(dip.getContentObject())) {
             status.add(FacetConstants.PATRON_SETTINGS);
         }
 
         return status;
     }
 
+    private boolean addPatronSettings(ContentObject contentObj) {
```
bbpennel, line 118: I'd suggest giving the method a name that indicates that it is evaluating a condition rather than actually performing an addition. Maybe `needsPatronSettingsStatus`, or merge it with `hasPatronSettings`

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -105,13 +108,45 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
...
+
+        return hasPatronSettings(contentObj.getPid(), objType);
+    }
+
+    private boolean hasPatronSettings(PID pid, ResourceType rescType) {
+        List<RoleAssignment> objPatronRoles = objAclFactory.getPatronRoleAssignments(pid);
+        int numRoles = objPatronRoles.size();
+        boolean isCollection = rescType.equals(Cdr.Collection.getLocalName());
+
+        if ((!isCollection && numRoles == 0)) {
```
bbpennel, line 133: you can remove the inner parens

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -105,13 +108,45 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
...
+            return false;
+        } else if (isCollection && numRoles < 2) {
+            return true;
+        }
+
+        String canViewOrig = canViewOriginals.getURI().toString();
```
bbpennel, line 139: `getURI()` already returns a string

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -105,13 +108,45 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
...
+        UserRole roleOne = objPatronRoles.get(0).getRole();
```
bbpennel, line 140: I'd suggest iterating through the list of roles and returning false if any of them aren't canViewOriginal. There will probably be more than 2 groups in the future, and should have the same affect.

bbpennel (changes requested):

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
lfarrell: Seems to return a URI, if the toString isn't added

solr-ingest/src/main/java/edu/unc/lib/dl/data/ingest/solr/filter/SetAccessStatusFilter.java
```diff
@@ -108,43 +108,35 @@ public void setObjectAclFactory(ObjectAclFactory oaf) {
...
         }
 
-        String canViewOrig = canViewOriginals.getURI().toString();
-        UserRole roleOne = objPatronRoles.get(0).getRole();
-
-        if (numRoles > 1) {
-            UserRole roleTwo = objPatronRoles.get(1).getRole();
-            return (!roleOne.equals(canViewOrig) || !roleTwo.equals(canViewOrig));
+        for (RoleAssignment role : objPatronRoles) {
+            if (!role.getRole().equals(canViewOriginals.getURI().toString())) {
```
bbpennel, line 134: `canViewOriginals.getURI()` should be sufficient since it returns a string

bbpennel (changes requested):

bbpennel (approved):

PR #1105, opened 2020-10-16
Only mark a file completed if it passes virus checking

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2812

### DIFF
--- a/deposit/src/test/java/edu/unc/lib/deposit/validate/VirusScanJobTest.java
+++ b/deposit/src/test/java/edu/unc/lib/deposit/validate/VirusScanJobTest.java
@@ -226,17 +226,17 @@ public class VirusScanJobTest extends AbstractDepositJobTest {
             fail();
         } catch(JobFailedException e) {
             verify(jobStatusFactory).setTotalCompletion(eq(jobUUID), eq(1));
-            verify(jobStatusFactory, times(0)).incrCompletion(eq(jobUUID), eq(0));
-        } finally {
-            job.run();
-
-            verify(jobStatusFactory, times(2)).setTotalCompletion(eq(jobUUID), eq(1));
-            verify(jobStatusFactory).incrCompletion(eq(jobUUID), eq(1));
-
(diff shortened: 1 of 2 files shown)


deposit/src/test/java/edu/unc/lib/deposit/validate/VirusScanJobTest.java
```diff
@@ -205,6 +207,38 @@ public void failOneScanTest() throws Exception {
...
+
+        job.closeModel();
+
+        try {
+            job.run();
+            fail();
+        } catch(JobFailedException e) {
+            verify(jobStatusFactory).setTotalCompletion(eq(jobUUID), eq(1));
+            verify(jobStatusFactory, times(0)).incrCompletion(eq(jobUUID), eq(0));
+        } finally {
```
bbpennel, line 230: I don't think you want this in a finally block, it means that it will still run even if the `fail()` up above was called

deposit/src/test/java/edu/unc/lib/deposit/validate/VirusScanJobTest.java
```diff
@@ -205,6 +207,38 @@ public void failOneScanTest() throws Exception {
...
+        PID filePid = addFileObject(depBag, textFile);
+
+        job.closeModel();
+
+        try {
+            job.run();
+            fail();
+        } catch(JobFailedException e) {
+            verify(jobStatusFactory).setTotalCompletion(eq(jobUUID), eq(1));
+            verify(jobStatusFactory, times(0)).incrCompletion(eq(jobUUID), eq(0));
```
bbpennel, line 229: very minor, but you can use `never()` instead of `times(0)`

bbpennel (changes requested):
This looks good aside from two minor things on the test. This change is a good thing, but it does mean we'll have to actually address virus scan failures since there's no way to proceed with the rest of the deposit. That might come up during the migration.

bbpennel (approved):

PR #1106, opened 2020-10-21
new collections should default to canViewOriginals permissions

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2819


bbpennel (approved):
Looks good, lets try it on the test server then probably good to go

PR #1108, opened 2020-10-23
WIP:Read only mode for dcr admin

lfarrell:
* Comment out links to put/post actions
* Turn refresh timers off
* Add read only note


lfarrell:
Only want this for bxc3
See: https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/1109

PR #1109, opened 2020-10-23
BXC3 - Read only mode for dcr admin (for freeze)

lfarrell:
* Comment out links to put/post actions
* Turn refresh timers off
* Add read only note


bbpennel (changes requested):
Looks good, there are a two extra things that we need to block:
* You can get to the MODS editor from the access interface by clicking the "Edit" button on the full record page. I'd suggest just disabling the controller that lets you POST the MODS back, so we don't have to modify the public interface too
* If you click and drag stuff you can still move it, so we'll need to disable that.

Also, I think it'd be a good idea to leave the reindex option available for admins, since we periodically run into situations were reindexing is needed and we don't know how long the freeze will be.

bbpennel:
It'd also be good to keep "Edit Access" around, but disallow editing. It's the main way to see the current and inherited permissions on an object. Could disable/remove the save button

bbpennel (approved):

PR #1114, opened 2020-10-30
Transform collections in collections into folders and set parentType destination

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2824
https://jira.lib.unc.edu/browse/BXC-2845

### DIFF
--- a/migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformer.java
+++ b/migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformer.java
@@ -457,10 +457,12 @@ public class ContentObjectTransformer extends RecursiveAction {
             return Cdr.Work;
         }
         if (contentModels.contains(COLLECTION.getResource())) {
-            if (Cdr.AdminUnit.equals(parentType) || !options.isTopLevelAsUnit()) {
-                return Cdr.Collection;
-            } else if (Cdr.Collection.equals(parentType)) {
+            if (Cdr.Collection.equals(parentType)) {
                 return Cdr.Folder;
+            } else if (Cdr.AdminUnit.equals(parentType) || !options.isTopLevelAsUnit()) {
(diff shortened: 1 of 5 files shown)


migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformer.java
```diff
@@ -459,6 +459,8 @@ private Resource getResourceType(Resource bxc3Resc) {
         if (contentModels.contains(COLLECTION.getResource())) {
             if (Cdr.AdminUnit.equals(parentType) || !options.isTopLevelAsUnit()) {
                 return Cdr.Collection;
+            } else if (Cdr.Collection.equals(parentType)) {
```
bbpennel, line 462: I don't think this is quite going to work as expected. When we migrate the RLA the `isTopLevelAsUnit` flag is going to be `false` in practice, so I think the first `if` is always going to win for Collections at the moment. Swapping their order would probably work. 

Additionally, the migration util doesn't seed the parentType parameter with the type of the bxc5 destination. This might be a problem if we end up migrating the RLA or other collections like it in multiple deposits, since it won't be aware the subcollections are going into a collection. It would be prudent to start doing this now, i'll create another ticket for that.

bbpennel (changes requested):

migration-util/src/main/java/edu/unc/lib/dcr/migration/content/ContentObjectTransformer.java
bbpennel: Also, could you add a warn level log statement with the pid of the object being changed from a collection to a folder?

bbpennel: https://jira.lib.unc.edu/browse/BXC-2845

bbpennel (approved):

PR #1122, opened 2020-11-09
BXC-2854 - Deposit datastream resources

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2854

* Refactors datastreams in the deposit model to be their own resources so that they can have more properties recorded for them
* Records digests for all datastreams which are transferred during deposits, including in some conflict scenarios
* Some additional properties recorded for more datastreams, like mimetype
* Removes IngestDepositRecordJobTest as it is the exact same test as IngestDepositRecordJobIT
* Removes unused methods and properties


deposit/src/test/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJobTest.java
```diff
@@ -181,6 +192,47 @@ public void depositWithWorkContainingFileAndFits() throws Exception {
...
+
+        job.closeModel();
+
+        job.run();
+
+        Model model = job.getReadOnlyModel();
+        Resource postFileResc = model.getResource(fileResc.getURI());
+
+        assertOriginalFileTransferred(postFileResc, FILE_CONTENT1);
+        assertFitsFileTransferred(postFileResc);
```
bbpennel, line 216: Are we verifying that fits history is being transferred?

metadata/src/main/java/edu/unc/lib/dl/util/DepositStatusFactory.java
```diff
@@ -65,13 +65,6 @@ public DepositStatusFactory() {
...
     }
 
-    public void addManifest(String depositUUID, String value) {
-        try (Jedis jedis = getJedisPool().getResource()) {
-            jedis.rpush(DEPOSIT_MANIFEST_PREFIX + depositUUID, value);
-        }
-
-    }
-
     public List<String> getManifestURIs(String depositUUID) {
```
bbpennel, line 68: delete me

PR #1125, opened 2020-11-11
Prevent text extraction for empty files

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2868

### DIFF
--- a/services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
+++ b/services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
@@ -105,21 +105,19 @@ public class FulltextProcessor implements Processor {
             return;
         }
 
-        if (text.length() > 0) {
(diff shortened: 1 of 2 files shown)


services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
```diff
@@ -105,19 +105,21 @@ public void process(Exchange exchange) throws Exception {
...
-        File derivative = derivativePath.toFile();
-        File parentDir = derivative.getParentFile();
-
-        // Create missing parent directories if necessary
-        if (parentDir != null) {
-            try {
-                Files.createDirectories(parentDir.toPath());
-            } catch (IOException e) {
-                throw new IOException("Failed to create parent directories for " + derivativePath + ".", e);
+        if (text.length() > 0) {
```
bbpennel, line 108: I think we still want to write out the extract file even if its empty, since I think this is how we determine if a service should be re-run with the "Run Enhancements" action if the force flag isn't set to true. So I'd suggest removing this `if`

bbpennel (changes requested):

PR #1126, opened 2020-11-12
BXC-2856 - Fix user based permissions issues

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2856
* Make sure that the user principal is provided for most ACL checks by using getAgentPrincipal instead of getGroups.
* Store agent principals in the group thread store rather than regenerating on each method call.
* Fix NPE when accessing structure browse in admin interface with no permissions to the content root

### DIFF
--- a/access-common/src/main/java/edu/unc/lib/dl/ui/access/StoreAccessLevelFilter.java
+++ b/access-common/src/main/java/edu/unc/lib/dl/ui/access/StoreAccessLevelFilter.java
@@ -73,9 +73,9 @@ public class StoreAccessLevelFilter extends OncePerRequestFilter implements Serv
                 session.removeAttribute("accessLevel");
                 accessLevel = new AccessLevel(username);
                 session.setAttribute("accessLevel", accessLevel);
-                AccessGroupSet groups = GroupsThreadStore.getGroups();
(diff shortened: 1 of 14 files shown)


access-common/src/main/java/edu/unc/lib/dl/ui/access/StoreAccessLevelFilter.java
```diff
@@ -73,17 +73,17 @@ public void doFilterInternal(HttpServletRequest request, HttpServletResponse res
                 session.removeAttribute("accessLevel");
                 accessLevel = new AccessLevel(username);
                 session.setAttribute("accessLevel", accessLevel);
-                AccessGroupSet groups = GroupsThreadStore.getGroups();
-                if (globalPermissionEvaluator.hasGlobalPrincipal(groups)) {
-                    Set<UserRole> roles = globalPermissionEvaluator.getGlobalUserRoles(groups);
+                AccessGroupSet principals = GroupsThreadStore.getAgentPrincipals().getPrincipals();
```
lfarrell, line 76: Not a big issue, but would a wrapper method be useful for all of these `GroupsThreadStore.getAgentPrincipals().getPrincipals()` calls?

PR #1127, opened 2020-11-13
Allow ingestors to create "staff only" folders

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2871


bbpennel (changes requested):
This seems to do, but could you add tests to `AddContainerIT`:
* verify that an ingester can create public folder
* verify that an ingester can create staff only folder
* verify that a user with canAccess role cannot create a folder. There is an authorization test in there (testAuthorizationFailure), but it seems to be testing for a user with no permissions.

bbpennel (approved):

PR #1151, opened 2020-11-25
BXC-2930 - Set tika limit to 1million characters

bbpennel:
Follow up for https://jira.lib.unc.edu/browse/BXC-2930

### DIFF
--- a/services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
+++ b/services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
@@ -118,4 +118,19 @@ public class FulltextProcessorTest {
         processor.process(exchange);
         assertFalse(finalDerivativeFile.exists());
     }
+
+    @Test
+    public void extractFulltextExceedsCharacterLimit() throws Exception {
+        processor.setCharacterLimit(10);
+
+        originalFile = tmpDir.newFile(originalFileName);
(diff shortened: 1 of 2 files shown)


services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
```diff
@@ -118,4 +118,19 @@ public void extractFromInvalidPdfTest() throws Exception {
         processor.process(exchange);
         assertFalse(finalDerivativeFile.exists());
     }
+
+    @Test
+    public void extractFulltextExceedsCharacterLimit() throws Exception {
```
cazzerson, line 123: Test the error message as well? Seems like a possible regression in the future.

cazzerson (approved):

services-camel/src/test/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessorTest.java
bbpennel: Do you mean check the log message warning? Not sure we're doing that anywhere at the moment, but I think its possible
https://stackoverflow.com/questions/29076981/how-to-intercept-slf4j-with-logback-logging-via-a-junit-test
It is checking that the extracted text is truncated to the character limit though, so it seems like if some other error occurs it probably won't pass that case

cazzerson:
Ah, yes, that makes sense. This looks fine as is.

PR #1155, opened 2020-11-30
BXC-2950 - Filter fedora queue

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2950

* Adds filters to reduce number of extraneous messages being sent from Fedora queue to enhancement queue
* Filter some unused fedora objects out from triples indexing
* Consolidates fedora jms constants
* Adds a test to confirm that pids for intermediate container ids parse correctly


services-camel/src/test/java/edu/unc/lib/dl/services/camel/routing/MetaServicesRouterTest.java
```diff
@@ -78,6 +84,110 @@ public void testRouteStartContainer() throws Exception {
...
+        template.sendBodyAndHeaders("", createEvent(CONTAINER_ID + "/datafs/original_file",
+                Fcrepo4Repository.Binary.getURI()));
+
+        assertMockEndpointsSatisfied();
+    }
+
+    @Test
+    public void testRouteStartPremisBinary() throws Exception {
+        getMockEndpoint("mock:direct-vm:index.start").expectedMessageCount(1);
+        getMockEndpoint("mock:direct:process.enhancement").expectedMessageCount(1);
```
lfarrell, line 181: Should event logs have enhancements run on them?

bbpennel: The issue at the moment is that longleaf registration happens in the EnhancementRouter, while the new filtering is in MetaServicesRouter, so all binaries have to reach MetaServicesRouter at the moment. But we don't need enhancements on any binaries except originals currently. I could move the call to longleaf up into the metaservicesrouter... that might be better in some ways, as running enhancements doesn't affect the original binary, so it shouldn't need to register in longleaf again. On the other hand, we might have to add a way to trigger longleaf registration for cases where it fails. I think that's fine though. I'll try to split longleaf out of the enhancement router.

PR #1156, opened 2020-11-30
Add separate queue for migration deposits

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2940
Depends on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/104


deposit/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -110,6 +110,21 @@
...
+                <constructor-arg name="config" ref="jesqueConfig" />
+                <constructor-arg name="queues">
+                    <util:list>
+                        <value>PREPARE_MIGRATION</value>
+                    </util:list>
+                </constructor-arg>
+                <constructor-arg ref="jobFactory" />
+            </bean>
+        </constructor-arg>
+        <constructor-arg name="numWorkers" value="${workers.high.num:4}" />
```
bbpennel, line 126: Could you give this a separate property, probably `workers.migration.num` and set its default to 2?

bbpennel (changes requested):
This looks good, added one comment. One other thought, another option would be to queue the migration deposits as a different priority. Maybe `Priority.low` since we aren't using that at the moment and we don't want them to compete with the user submitted deposits. Do you have any preference between that approach and the current approach of checking the packaging type?

lfarrell:
Either way's fine with me, depositType or priority

deposit/src/main/java/edu/unc/lib/deposit/work/DepositSupervisor.java
```diff
@@ -187,7 +187,7 @@ public DepositSupervisor() {
     }
 
     private static enum Queue {
-        PREPARE, DELAYED_PREPARE, CDRMETSCONVERT, PREPARE_HIGH_PRIORITY;
+        PREPARE, DELAYED_PREPARE, CDRMETSCONVERT, PREPARE_HIGH_PRIORITY, PREPARE_MIGRATION;
```
bbpennel, line 190: Sorry to be a pain, but since you've gone with the `low` priority thing, I'd suggest naming the queue `PREPARE_LOW_PRIORITY` as well

deposit/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -110,6 +110,21 @@
...
+                <constructor-arg name="config" ref="jesqueConfig" />
+                <constructor-arg name="queues">
+                    <util:list>
+                        <value>PREPARE_MIGRATION</value>
+                    </util:list>
+                </constructor-arg>
+                <constructor-arg ref="jobFactory" />
+            </bean>
+        </constructor-arg>
+        <constructor-arg name="numWorkers" value="${workers.migration.num:2}" />
```
bbpennel, line 126: similarly, i would name the property `workers.low.num` now

bbpennel (changes requested):

bbpennel (approved):

PR #1162, opened 2020-12-09
Skip JSON files when extracting full text

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2963

### DIFF
--- a/services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
+++ b/services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
@@ -97,12 +97,13 @@ public class FulltextProcessor implements Processor {
 
         String fedoraUri = (String) in.getHeader(FCREPO_URI);
         String binaryPath = (String) in.getHeader(CdrBinaryPath);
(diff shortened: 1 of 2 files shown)


services-camel/src/main/java/edu/unc/lib/dl/services/camel/fulltext/FulltextProcessor.java
```diff
@@ -97,12 +97,13 @@ public void process(Exchange exchange) throws Exception {
 
         String fedoraUri = (String) in.getHeader(FCREPO_URI);
         String binaryPath = (String) in.getHeader(CdrBinaryPath);
+        String mimetype = (String) in.getHeader(CdrFcrepoHeaders.CdrBinaryMimeType);
```
bbpennel, line 100: What about having `FulltextProcessor.allowedTextType` filter out json files instead of doing so at extraction time?

lfarrell: Hmm, looks like they should already be filtered out, as long as they get the right mimetype from fits. If they get the text/plain mimetype they'll pass the test in this pr and still have their text extracted. Seems this PR can just be closed.

lfarrell:
This shouldn't be needed.

PR #1164, opened 2020-12-10
BXC-2959 - Refactor fixity check job to run checks in parallel

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2959
* Refactors FixityCheckJob to run fixity check jobs in parallel
    * Each fixity check job runs in a separate worker thread using a pool thread
    * Updates to the deposit model and PREMIS logs are flushed ever ~2 seconds to reduce the number of transactions against jena TDB
    * Fixity check jobs are gradually submitted to the queue for execution in order to prevent one deposit from taking over all the workers for a long time, which would delay the other deposits.
* Has its own thread pool for the moment since VirusScanJob is not currently setup to play nice with other jobs in terms of number of tasks queued at once.


cazzerson (approved):
This is great--thank you!!

cazzerson (approved):
Thanks for these changes--merge whenever you're ready!

PR #1170, opened 2020-12-14
Set Shibboleth redirect to always use https

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2975

### DIFF
--- a/access/src/main/webapp/WEB-INF/jsp/index.jsp
+++ b/access/src/main/webapp/WEB-INF/jsp/index.jsp
@@ -33,7 +33,7 @@
     <c:set var="currentUrl" value="${currentUrl}?${pageContext.request.queryString}"/>
 </c:if>
 <c:set var="currentRelativeUrl" scope="request" value="${currentUrl}"/>
-<c:set var="currentAbsoluteUrl" scope="request" value="${pageContext.request.scheme}://${pageContext.request.serverName}${currentRelativeUrl}"/>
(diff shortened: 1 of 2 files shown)


access/src/main/webapp/WEB-INF/jsp/index.jsp
```diff
@@ -33,7 +33,7 @@
     <c:set var="currentUrl" value="${currentUrl}?${pageContext.request.queryString}"/>
 </c:if>
 <c:set var="currentRelativeUrl" scope="request" value="${currentUrl}"/>
-<c:set var="currentAbsoluteUrl" scope="request" value="${pageContext.request.scheme}://${pageContext.request.serverName}${currentRelativeUrl}"/>
+<c:set var="currentAbsoluteUrl" scope="request" value="${accessBaseUrl}${currentRelativeUrl}"/>
```
bbpennel, line 36: This mostly works, the only issue is that it produces a URL with two /'s at the end. That seems to be okay on the front page, but if you log on from a search result page, you get sent to:
https://dcr-test.lib.unc.edu//search?anywhere=
which causes an error. There is `edu.unc.lib.dl.util.URIUtil.join(String, String...)` which I believe can be used to join together strings in a way that will prevent the double slashes. Its being used here:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/access/src/main/webapp/WEB-INF/jsp/common/headerSmall.jsp#L65-L66

bbpennel (changes requested):

bbpennel (approved):

PR #1171, opened 2020-12-14
Allow user to pass in a display label.

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2955


migration-util/src/main/java/edu/unc/lib/dcr/migration/MigrationCLI.java
```diff
@@ -85,6 +85,11 @@
             description = "Port for redis. Default 6379.")
     protected int redisPort;
 
+    @Option(names = {"--display-label"},
```
bbpennel, line 88: since this option only applies to SubmitDepositCommand and TransformContentCommand, could you add the option directly to those to keep down the clutter in the --help option?

PR #1174, opened 2020-12-16
BXC-2961 - Parallel Binary Transfers

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2961

* Refactors binary transfers job to perform transfers in parallel to each, following a similar setup to https://github.com/UNC-Libraries/Carolina-Digital-Repository/pull/1164
    * All transfers per object happen in the same runnable executed in a thread pool
    * Updates to the deposit model every few seconds to reduce the number of write transactions against jena TDB
    * jobs are gradually submitted to the queue for execution in order to prevent one deposit from taking over all the workers for a long time, which would delay the other deposits.
* Due to needing to read from the jena tdb model in multiple threads, a read transaction must be opened in each transfer worker.
* Refactored job to process objects needing transfers based off a single query of the deposit model rather than following the deposit hierarchy, in order to prevent deadlock issues when spawning jobs.

### DIFF
--- a/deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
+++ b/deposit/src/main/java/edu/unc/lib/deposit/transfer/TransferBinariesToStorageJob.java
@@ -15,7 +15,6 @@
  */
 package edu.unc.lib.deposit.transfer;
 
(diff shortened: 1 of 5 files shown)


cazzerson (approved):
Thanks for reviewing this with me--it looks good!

PR #1179, opened 2021-01-05
Add more metadata view options

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2994

### DIFF
--- a/static/js/admin/src/ResultObjectActionMenu.js
+++ b/static/js/admin/src/ResultObjectActionMenu.js
@@ -16,7 +16,7 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 	
 	ResultObjectActionMenu.prototype.create = function() {
 		var self = this;
-		
+
 		if (this.options.multipleSelectionEnabled) {
 			this.batchActions = {};
 			var actionClasses = [];
@@ -106,6 +106,9 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 		
 		var items = {};
 		var isContainerFlag = false;
+		var datastreams = metadata.datastream;
+		var viewHidden = $.inArray('viewHidden', metadata.permissions) !== -1;
+
 		if (resultObject.isContainer) {
 			isContainerFlag = true;
 			items["openContainer"] = {name : "Open"};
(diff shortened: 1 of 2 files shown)


static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -115,11 +118,30 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 			var originalFile = resultObject.getDatastream("original_file");
 			items["viewFile"] = {name : "View File"
 				+ " ("+ StringUtilities.readableFileSize(originalFile['fileSize']) + ")"};
-
-			items["viewFits"] = {name: "View FITS"}
 		}
-		if ($.inArray('viewHidden', metadata.permissions) != -1) {
-			items["viewEventLog"] = {name : "View Event Log"};
+
+		if (viewHidden || isFile || /md_descriptive|tech_md/ig.test(datastreams)) {
```
bbpennel, line 123: I think you can remove `tech_md` from the regex here, it should only/always be present if the object is a file (the work may list the tech_md of its primary object, but I don't think we need to link to it from the work).

static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -115,11 +118,30 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
...
+			items["metadata"] = {name: "View metadata", items: {}}
+
+			if (/md_descriptive/ig.test(datastreams)) {
+				items['metadata']['items']["viewMods"] = {name: "View MODS"};
+			}
+
+			if (/md_descriptive_history/ig.test(datastreams)) {
```
bbpennel, line 130: this datastream requires `viewHidden` permission, at the moment i can see the option for items that I can't retrieve the datastream for (gives a blank screen). I tested this by spoofing as an observer and then checked the available options for Units I didn't have admin permission to.

static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -115,11 +118,30 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
...
+				items['metadata']['items']["viewModsHistory"] = {name: "View MODS History"};
+			}
+
+			if (isFile) {
```
bbpennel, line 134: Both FITS and its history require viewHidden. Although I actually don't think there's any way to get to a file object in the admin ui that you don't have a staff role for. See https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/fcrepo-clients/src/main/java/edu/unc/lib/dl/model/DatastreamType.java for which permissions datastreams require.

static/js/admin/src/ResultObjectActionMenu.js
bbpennel: Also, I think it'd be fine to require `viewHidden` to access the `metadata` submenu at all. The only case where you can view an object in the admin UI that you don't have a staff role for is for a AdminUnit that you don't have permissions to. Seems reasonable to not offer admin access to datastreams for objects you don't have permission to.

If you opt to add this restriction, then you can ignore my other two comments to add `viewHidden` restrictions in. You'll also want to remove the check from the event log.

bbpennel (changes requested):
This looks good, pending feedback from users. I added a few suggestions regarding viewHidden permissions

static/js/admin/src/ResultObjectActionMenu.js
lfarrell: That might explain Anna's error. :)

bbpennel (approved):

PR #1186, opened 2021-01-11
Restrict CSV export to 50,000 objects

lfarrell:
* Restrict CSV export to 50,000 objects
* Make sure parent gets a child count

https://jira.lib.unc.edu/browse/BXC-2999

### DIFF
--- a/static/js/admin/src/ResultView.js
+++ b/static/js/admin/src/ResultView.js
@@ -213,7 +213,7 @@ define('ResultView', [ 'jquery', 'jquery-ui', 'ResultObjectList', 'URLUtilities'
 			}
 		
 			if (container) {
-				new ParentResultObject({metadata : this.setChildCounts(container),
+				new ParentResultObject({metadata : container,
 						element : $(".container_entry")});
 		
 				if (this.addMenu) {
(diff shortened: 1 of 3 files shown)


static/js/admin/src/ResultView.js
```diff
@@ -242,6 +242,21 @@ define('ResultView', [ 'jquery', 'jquery-ui', 'ResultObjectList', 'URLUtilities'
 			this.resizeResults();
 			this.$window.resize($.proxy(this.resizeResults, this));
 		},
+
+		// Set count of descendants
+		setChildCounts: function(container) {
```
bbpennel, line 247: this is clever, but seems like a lot of work. I think you could change the response from the controller to add a count in for the parent container.
In here:
https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/fcrepo4/admin/src/main/java/edu/unc/lib/dl/admin/controller/SearchController.java#L173

Pass `resultResponse.getSelectedContainer()` into `childrenCountService.addChildrenCounts` similar to https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/9cad1419fa96a243348f2a9e8307aa0b4be44079/admin/src/main/java/edu/unc/lib/dl/admin/controller/AbstractSearchController.java#L63

You would probably have to wrap it in a temporary list

bbpennel (changes requested):

bbpennel (approved):

PR #1198, opened 2021-02-02
Skip empty segments when creating uris

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3025


bbpennel:
Good work, and thanks for adding a test

PR #1200, opened 2021-02-10
Move loading of setCollectionSupplementalInformationFilter so it loads after setPathFilter

lfarrell:
Relies on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/114


lfarrell:
Relies on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/114

PR #1209, opened 2021-02-22
Setup Universal Viewer

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3039


access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -141,11 +156,38 @@ public void streamJP2(String simplepid, String region, String size, String rotat
...
+    public String getManifest(String id, String datastream, HttpServletRequest request) throws JsonProcessingException {
+        String[] url = request.getRequestURL().toString().split("\\/");
+        String urlBase = "https://" + url[2];
+        String manifestBase = urlBase + "/" + id;
+        ObjectMapper iiifMapper = new IiifObjectMapper();
+
+        Manifest manifest = new Manifest(manifestBase + "/manifest");
+
+        Canvas canvas = new Canvas(manifestBase);
+        String path = urlBase + "/jp2Proxy/" +
```
bbpennel, line 168: It looks like this is reconstructing `request.getRequestURL()` with `/manifest` removed from the end. It seems like you could probably skip passing `datastream` into this method (and possibly `id` as well), and glue back together the pieces in `String[] url`. 
One other suggestion, we have `URIUtil.join` which can put together path segments in a URI, and will handle making sure the /'s are in the right places.

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -141,11 +156,38 @@ public void streamJP2(String simplepid, String region, String size, String rotat
...
+                id + "/" + datastream;
+        canvas.addIIIFImage(path, ImageApiProfile.LEVEL_TWO);
+
+        Sequence seq = new Sequence(manifestBase + "/sequence/normal");
+        seq.addCanvas(canvas);
+
+        return iiifMapper.writerWithDefaultPrettyPrinter().writeValueAsString(manifest.addSequence(seq));
```
bbpennel, line 175: probably don't need to pretty print this long term

access-common/pom.xml
```diff
@@ -123,5 +130,11 @@
             <groupId>edu.unc.lib.cdr</groupId>
             <artifactId>fcrepo-clients</artifactId>
         </dependency>
-    </dependencies>
+      <dependency>
```
bbpennel, line 133: You should be able to remove this dependency after making some of the changes suggested down in the loris service. I'd rather not get OSGI dependencies in here

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -78,6 +89,10 @@ public void getMetadata(String simplepid, String datastream, OutputStream outStr
                     response.setHeader("Content-Type", "application/json");
                     response.setHeader("content-disposition", "inline");
 
+                    String updatedRespBody = EntityUtils.toString(httpResp.getEntity())
+                            .replaceAll("http://localhost.*?jp2", "/jp2Proxy/" + simplepid + "/jp2");
```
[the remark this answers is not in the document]
bbpennel: If you want to go the root of doing regex replacement of the url, you can get the content body as a string without any new dependencies by doing
```
IOUtils.toString(httpResp.getEntity().getContent(), StandardCharsets.UTF_8);
```

access-common/pom.xml
```diff
@@ -81,6 +81,13 @@
             <groupId>org.jdom</groupId>
             <artifactId>jdom2</artifactId>
         </dependency>
+
+        <!-- IIIF -->
+        <dependency>
+            <groupId>de.digitalcollections.iiif</groupId>
+            <artifactId>iiif-apis</artifactId>
+            <version>0.3.8</version>
```
bbpennel, line 89: Could you add this dependency to the root pom.xml and make a variable for the version number like with the other dependencies?

bbpennel (changes requested):
Do we need to check all the UV files and dependencies into our repo? Its a lot of files/lines. I'd much rather exclude them if possible and rely on the build process to download them when needed.
It seems like we must be missing some CSS stuff for the popup menus in the viewer, they are styled rather differently than in hyc.

bbpennel (approved):
Much cleaner now, this PR looks good to me! The jquery interference could be because we are using a rather old version of jquery (would be good to update that), or maybe just the fact that jquery was being imported twice

PR #1213, opened 2021-02-24
Catch restricted access exceptions

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3047

### DIFF
--- a/access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
+++ b/access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
@@ -122,8 +122,13 @@ public class FullRecordController extends AbstractSolrSearchController {
 
         AccessGroupSet principals = getAgentPrincipals().getPrincipals();
 
-        aclService.assertHasAccess("Insufficient permissions to access full record",
-                pid, principals, Permission.viewMetadata);
+        try {
+            aclService.assertHasAccess("Insufficient permissions to access full record",
+                    pid, principals, Permission.viewMetadata);
+        } catch (AccessRestrictionException e) {
+            LOG.info("{}", e.getMessage());
+            throw new InvalidRecordRequestException();
+        }
 
         SimpleIdRequest idRequest = new SimpleIdRequest(pidString, principals);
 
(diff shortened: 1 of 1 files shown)


access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
```diff
@@ -162,8 +163,13 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
 
         AccessGroupSet principals = getAgentPrincipals().getPrincipals();
 
-        aclService.assertHasAccess("Insufficient permissions to access full record",
-                pid, principals, Permission.viewMetadata);
+        String errorMsg = "Insufficient permissions to access full record";
```
bbpennel, line 166: Put this directly into the assertHasAccess call and remove it from the log statement. At the moment, my log is getting it 3 times:
```
e.u.l.d.u.c.FullRecordController - Insufficient permissions to access full record. Insufficient permissions to access full
 record
edu.unc.lib.dl.acl.exception.AccessRestrictionException: Insufficient permissions to access full record
```

access/src/main/java/edu/unc/lib/dl/ui/controller/FullRecordController.java
```diff
@@ -162,8 +163,13 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
...
+        try {
+            aclService.assertHasAccess(errorMsg,
+                    pid, principals, Permission.viewMetadata);
+        } catch (AccessRestrictionException e) {
+            LOG.info("{}. {}", errorMsg, e.getMessage(), e);
+        }
```
bbpennel, line 172: This is eating the assertHasAccess call, so not a whole lot of purpose to it now. Also, this is still logging a full stacktrace for me on permission failures. I'd suggest a few changes, including throwing an exception, not passing the exception into the log statement
```suggestion
        try {
            aclService.assertHasAccess("Insufficient permissions to access full record",
                    pid, principals, Permission.viewMetadata);
        } catch (AccessRestrictionException e) {
            LOG.info("{}", e.getMessage());
            throw new InvalidRecordRequestException();
        }
```

bbpennel: Also, can you make these same changes up in the `/{pid}/metadataView` end point in this same class? It has the same problem if you access it directly

bbpennel (changes requested):

bbpennel (approved):

PR #1215, opened 2021-02-24
Bump up number of returned collections

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3049


access-common/src/main/java/edu/unc/lib/dl/ui/service/SolrQueryLayerService.java
```diff
@@ -70,12 +70,14 @@
      * @return
      */
     public SearchResultResponse getCollectionList(AccessGroupSet accessGroups) {
+        int MAX_COLLECTIONS_TO_RETRIEVE = 350;
```
bbpennel, line 73: Move this variable out of the method and make it into a constant, its already got the capitalization for being one. Also, I'd suggest going to a higher number, we're probably going to be adding a lot of new collections in the not too distant future. 1000 seems like a reasonable number

PR #1216, opened 2021-03-01
Add multiple images with metadata

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3040


bbpennel (changes requested):
Here's some preliminary feedback, I know there will likely be other adjustments based on metadata changes and maybe ordering.

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -234,25 +236,24 @@ public String getManifest(HttpServletRequest request, List<BriefObjectMetadata>
...
         if (datastream != null) {
-            for (Datastream stream : datastream) {
-                String streamValues = stream.toString();
-                if (streamValues.trim().startsWith("jp2")) {
-                    String[] uuid_parts = streamValues.split("\\|");
-                    if (uuid_parts.length == 7) {
-                        return uuid_parts[6];
-                    }
-                    return uuid_parts[2].split("\\.")[0];
+            String streamValues = datastream.toString();
```
bbpennel, line 241: the Datastream class already explodes the string representation of the solr field, so you can just ask it for whatever fields you're trying to get. `owner` is the 6th field, etc. See `edu.unc.lib.dl.search.solr.model.Datastream.Datastream(String)`

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -202,16 +204,16 @@ public String getManifest(HttpServletRequest request, List<BriefObjectMetadata>
...
-        setMedataField(manifest, "Languages", language);
+        setMetadataField(manifest, "Creators", creators);
+        setMetadataField(manifest, "Keywords", keywords);
+        setMetadataField(manifest, "Subjects", subjects);
+        setMetadataField(manifest, "Languages", language);
 
         Sequence seq = new Sequence(URIUtil.join(manifestBase, "sequence", "normal"));
 
-        List<String> uuidList = new ArrayList<String>();
+        List<String> uuidList = new ArrayList<>();
```
bbpennel, line 214: Doesn't matter much for data at this scale, but a Set/HashSet is often a good choice for a data structure for quickly checking to see if a value has already occurred.

bbpennel (changes requested):

bbpennel (changes requested):
Some more cleanup and validation, but otherwise we can probably move ahead with this soon

bbpennel (changes requested):

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -167,23 +176,140 @@ public void streamJP2(String simplepid, String region, String size, String rotat
...
+    }
 
-        Canvas canvas = new Canvas(manifestBase);
-        String path = URIUtil.join(basePath, "jp2Proxy", uuid, datastream);
-        canvas.addIIIFImage(path, ImageApiProfile.LEVEL_TWO);
+    private String jp2Pid(BriefObjectMetadata briefObj) {
+        Datastream datastream = briefObj.getDatastreamObject(DatastreamType.JP2_ACCESS_COPY.getId());
+        if (datastream != null) {
+            String id = datastream.getOwner();
+            if (id.equals("")) {
```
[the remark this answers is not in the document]
lfarrell: If you return "" you miss anything that's not the primary obj. The file name, in every example I've seen, is the uuid + file extension.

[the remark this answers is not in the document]
bbpennel: Ah okay, I understand, I was looking at the original_file datastream's filename field. Anyways, since all you need in this case is the ID of the object itself, you could return `briefObj.getId()` instead of parsing it out of the filename

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -167,23 +176,140 @@ public void streamJP2(String simplepid, String region, String size, String rotat
...
+
+            uuidList.add(datastreamUuid);
+        }
+
+        return seq;
+    }
+
+    private Canvas createCanvas(String path, BriefObjectMetadata briefObj) {
+        String title = getTitle(briefObj);
+        String uuid = jp2Pid(briefObj);
```
bbpennel, line 258: if there's no JP2 for the object, I think this method is going to produce an invalid canvas path (uuid will either be blank or null). This only impacts `getCanvas`, since the usage in `createSequence` is already verifying that there is a jp2. For `getCanvas`, i'm not sure what the best approach would be.  Returning the canvas without adding any of the jp2/thumbnail stuff? Or throwing an exception?

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -167,23 +176,140 @@ public void streamJP2(String simplepid, String region, String size, String rotat
...
+
+        canvas.addIIIFImage(canvasPath, ImageApiProfile.LEVEL_TWO);
+        ImageContent thumb = new ImageContent(URIUtil.join(basePath,
+                "services", "api", "thumb", uuid, "large"));
+        canvas.addImage(thumb);
+
+        String label = briefObj.getLabel();
+        if (label != null) {
+            canvas.addLabel(label);
+        }
```
bbpennel, line 278: I think this can be removed. `new Canvas(path, title)` adds the title as a label, and it looks like the label field isn't being populated in bxc5.

access-common/src/main/java/edu/unc/lib/dl/ui/service/LorisContentService.java
```diff
@@ -167,23 +176,140 @@ public void streamJP2(String simplepid, String region, String size, String rotat
...
+        List<String> subjects = rootObj.getSubject();
+        List<String> language = rootObj.getLanguage();
+
+        if (abstractText != null) {
+            manifest.addDescription(abstractText);
+        }
+
+        if (label != null) {
+            manifest.addLabel(label);
+        }
```
bbpennel, line 200: I believe this and L189 can be removed too, since label isn't populated

bbpennel (changes requested):

PR #1219, opened 2021-03-03
Update admin result entries after changes

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3054


admin/src/main/java/edu/unc/lib/dl/admin/controller/ResultEntryController.java
```diff
@@ -61,6 +62,7 @@ String getResultEntry(@PathVariable("pid") String pid, Model model, HttpServletR
             throw new ResourceNotFoundException("The requested record either does not exist or is not accessible");
         }
 
+        childrenCountService.addChildrenCounts(Collections.singletonList(entryBean), accessGroups);
```
bbpennel, line 65: Can probably skip this for FileObjects, something like `!entryBean
.getResourceType().equals(searchSettings.resourceTypeFile)`

bbpennel (commented):
This looks really good, just added one minor suggestion

PR #1221, opened 2021-03-04
BXC-3066 Update record view for description subelement of name

agoslen:


access/src/main/resources/recordTransformations/modsToFullRecord.xsl
```diff
@@ -69,10 +69,15 @@
 
 		<xsl:variable name="description" select="*[local-name() = 'description']"/>
 		<xsl:if test="boolean($description)">
-			<xsl:text>Description:  </xsl:text><xsl:value-of select="$description"/><br/><xsl:value-of select="$newline"/>
+			<span>
+			    <xsl:text>Description:  </xsl:text>
+				<xsl:for-each select="*[local-name() = 'description']">
+					<xsl:value-of select="text()"/><br/><xsl:value-of select="$newline"/>
+				</xsl:for-each>
+			</span>
```
bbpennel, line 77: This is the same selector that the `description` variable is set to, so I believe you can reuse it:
```suggestion
				<xsl:text>Description:  </xsl:text>
				<xsl:for-each select="$description">
					<xsl:value-of select="text()"/><br/><xsl:value-of select="$newline"/>
				</xsl:for-each>
			</span>
```
And while i'm being pedantic, the `<xsl:text>` tag is indented with spaces instead of tabs like the rest of the file. We should probably switch this file over to spaces at some point.

bbpennel (changes requested):
This works as expected for me, just left a comment about some minor cleanup

access/src/main/resources/recordTransformations/modsToFullRecord.xsl
agoslen: Foiled by spaces vs tabs once again

bbpennel (approved):

PR #1223, opened 2021-03-09
Remove unused accession number and media id

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3057

### DIFF
--- a/static/js/xmleditor/templates/archives.xml
+++ b/static/js/xmleditor/templates/archives.xml
@@ -3,9 +3,13 @@
   <mods:titleInfo>
     <mods:title />
   </mods:titleInfo>
+  <mods:identifier displayLabel="Accession Number" type="local" />
   <mods:identifier displayLabel="Digital Folder Number" type="local" />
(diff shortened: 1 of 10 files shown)


bbpennel (changes requested):
Please also remove the DepositField enum values related to these fields. The code that populates these fields during a deposit are still being set too. See `edu.unc.lib.deposit.normalize.AbstractFileServerToBagJob.addDescription`

While you're looking at that method, the ticket states that the "label" field should be recorded as the MODS title now.

static/js/xmleditor/templates/archives.xml
```diff
@@ -3,13 +3,9 @@
   <mods:titleInfo>
     <mods:title />
   </mods:titleInfo>
-  <mods:identifier displayLabel="Accession Number" type="local" />
```
bbpennel, line 6: Looks like we want to keep these fields in the MODS template here

bbpennel (approved):

PR #1235, opened 2021-03-30
BXC-2459 - IP Address Patron Groups

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2459

* Adds custom patron groups, which will be matched to users based on IP address ranges
    * these groups are added to the user's access groups at request time via a new PatronPrincipalProvider class
    * groups are defined in a config file
    * custom groups must begin with "unc:patron:" and IP groups should begin with "unc:patron:ipp:"
* Updates patron access control editor to allow for custom patron groups to be added
    * Custom patron groups must be selected from the list of allowable configured patron groups
    * Refactoring to separate some interdependencies, data nesting, depend more on computed methods, adjust previews for custom groups, rendering of all patron groups together.
* Updates access control calculation classes to allow for custom patron groups
    * they operate like the existing patron groups, except if the two standard groups are set to `none`, then the custom group will also be assumed to be `none` unless explicitly set.
* Updates solr indexing of access status field so that the "staff only" flag is not shown when a custom group is active, and a few other related adjustments.

### DIFF
--- a/static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
+++ b/static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
@@ -49,6 +49,30 @@
                             </select>
                         </div>
                     </li>
+                    <li>
+                        <div v-for="(other_patron_princ, index) in otherAssignedPrincipals" :key="index" class="other-patron-assigned">
+                            <p>{{ otherPrincipalName(other_patron_princ) }}</p>
+                            <select @change="updateRole('other')" v-model="other_patron_princ.role" :disabled="shouldDisable">
+                                <template v-for="(role, index) in possibleRoles">
+                                    <option v-if="index > 0" :value="role.role">{{ role.text }}</option>
+                                </template>
+                            </select>
+                        </div>
+                        <div id="add-new-patron-principal" :class="{'hidden': shouldShowAddOtherPrincipals}">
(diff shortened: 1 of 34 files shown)


static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
```diff
@@ -118,25 +135,23 @@
...
+                selected_patron_assignments: [],
+                embargo: null,
+                deleted: false,
+                new_assignment_role: VIEW_ORIGINAL_ROLE,
+                new_assignment_principal: '',
                 response_message: '',
-                unsaved_changes: false,
-                user_type: ''
+                user_type: null,
+                shouldShowAddPrincipal: false,
```
bbpennel, line 153: snake case

static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
```diff
@@ -299,159 +392,36 @@
...
+                            ((current.type === 'inherited' && firstRole.type === 'inherited') ||
+                            (current.type === 'assigned' && firstRole.type === 'assigned')))) {
+                        allSame = false;
+                        break;
                     }
                 }
 
-                return [everyoneRole, authenticatedRole];
+                if (allSame) {
+                    firstRole.principal = firstRole.role === STAFF_ONLY_ROLE ? ACCESS_TYPE_STAFF_ONLY : ACCESS_TYPE_DIRECT;
```
bbpennel, line 420: switch to the principal constants

PR #1242, opened 2021-04-05
BXC-3058 - Make creation of folder for bag deposits optional

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3058

* Creation of a parent folder for BagIt Bag deposits is now optional, depending on the createParentFolder flag
* StaffOnlyPermissionJob works for multiple top level deposit objects
* Updates the Add From File Server confirmation/metadata page to support setting this flag for BagIt bags
    * Folder name field not required for Bagit deposits unless creating parent folder
    * Adjustments to styling and field order, more information displayed via tooltips
* Updates the qtip tooltip library to its most recent (final) version


fcrepo-clients/src/main/java/edu/unc/lib/dl/persist/api/ingest/DepositData.java
```diff
@@ -275,4 +276,16 @@ public boolean getOverrideTimestamps() {
     public void setOverrideTimestamps(boolean overrideTimestamps) {
         this.overrideTimestamps = overrideTimestamps;
     }
+
+    /**
+     * @return if true, then if this deposit would create a parent container to wrap
```
lfarrell, line 281: Is this note write? It seems like if it's true it should create a folder.

bbpennel: You're right, I flipped the direction of the flag halfway through working on it. I'll update

PR #1246, opened 2021-04-12
Patron access editor Save/Cancel button updates

lfarrell:
* Fix submit/cancel buttons to always display
* Fix issue with "warning" not popping up when there are unsaved changes

https://jira.lib.unc.edu/browse/BXC-3092

### DIFF
--- a/static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
+++ b/static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
@@ -662,16 +662,45 @@
 
         .submit-btn-options {
             position: absolute;
-            bottom: 1vh;
-            padding: 10px 30px;
+            top: 90.5vh;
             width: 100%;
             max-width: 620px;
             pointer-events: none;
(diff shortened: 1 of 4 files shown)


static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
```diff
@@ -512,6 +516,7 @@
              * See mixins/displayModal.js
              */
             showModal() {
+                this.unsaved_changes = this.hasUnsavedChanges;
```
bbpennel, line 519: I don't see any usages of `this.unsaved_changes`, maybe it can be removed?

bbpennel (changes requested):

static/js/admin/vue-permissions-editor/src/components/patronRoles.vue
lfarrell: It's used by the modal mixin to determine how it'll close

bbpennel (approved):

PR #1247, opened 2021-04-20
BXC-3091 - Indexing delays

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3091

- Solr updates performed using a concurrent update client with additional logging to record on partial batch failures. Switch to config options for update client
- Object models are no longer always refreshed, instead a refresh must be requested or a new object should be retrieved.
    - Add cache invalidation step to camel processing so that longer cache TTLs can be used to decrease number of times objects are retrieved
    - Small solr updates no longer batched
- Refactor UpdateDescriptionService to take request objects
- Add flag for indexing messages and some operation messages to indicate what indexing priority to use. Send low priority indexing messages to a separate queue
- Trigger commit of jms session to prevent bulk xml operation from acculmulating all updates in a single session and sending all at once when the job finishs
- Skip unnecessary retrieval of patron roles for adminunits and root objects
- Fix incorrect test expectation to resolve flapping test, add short delays to ensure tests with 0 expected messages have time for the message to reach all routes
- Use repositoryObjectLoader to return copy of object from object factory for more consistent caching and management of objects, resolving some consistency issues when instances of objects come from different sources

### DIFF
--- a/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
+++ b/fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
@@ -282,9 +282,9 @@ public abstract class RepositoryObject {
     }
 
     /**
-     * Refresh the state of this object from data sources
+     * Indicate that the state of this object should be refreshed
      */
-    public void refresh() {
+    public void shouldRefresh() {
         model = null;
         premisLog = null;
         types = null;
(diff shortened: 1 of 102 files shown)


fcrepo-clients/src/main/java/edu/unc/lib/dl/fcrepo4/RepositoryObject.java
```diff
@@ -272,6 +281,16 @@ public boolean isUnmodified() {
         return remoteEtag.equals(getEtag());
     }
 
+    /**
+     * Refresh the state of this object from data sources
+     */
+    public void refresh() {
```
bbpennel, line 287: maybe `shouldRefresh`

services/src/main/webapp/WEB-INF/access-fedora-context.xml
```diff
@@ -42,11 +42,17 @@
         <property name="objectAclFactory" ref="objectAclFactory" />
     </bean>
     
+    <bean id="repositoryObjectLoaderNoCache" class="edu.unc.lib.dl.fcrepo4.RepositoryObjectLoader" init-method="init">
```
bbpennel, line 45: Might be able to refactor this out, and explicitly ask for refreshed copies of objects

PR #1257, opened 2021-05-20
BXC-3085 - Bulk Patron Access Updates

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3085

* Users may now update the patron access control settings of multiple items in the admin UI at once, by selecting them then picking the Update Patron Access action
* Bulk mode adds additional "No Change" option, as well as "Clear embargo" for embargoes. Custom embargo gets to be a radio button too.
* ACL API endpoints behind filter for storing access level, which is used to determine access for retrieving configured patron principals
* Most patron access updates now take place in camel via a persistent queue
* Request objects used for passing patron access updates around


services/src/main/java/edu/unc/lib/dl/cdr/services/rest/modify/UpdatePatronAccessController.java
```diff
@@ -65,19 +76,104 @@
...
+            result.put("error", "Request must includes ids and access details");
+            return new ResponseEntity<>(result, HttpStatus.BAD_REQUEST);
         }
 
+        AgentPrincipals agent = AgentPrincipals.createFromThread();
+        List<PID> pids = bulkAccessDetails.getIds().stream().map(PIDs::get).collect(Collectors.toList());
+        int count = 0;
+        try {
+            for (PID pid : pids) {
+                processUpdate(agent, pid, bulkAccessDetails.getAccessDetails(), bulkAccessDetails.isSkipEmbargo(),
```
bbpennel, line 121: Switch to validating up front and then sending.

static/js/admin/vue-permissions-editor/tests/unit/patronRoles.spec.js
```diff
@@ -1162,6 +1159,159 @@ describe('patronRoles.vue', () => {
...
+                expect(global.confirm).toHaveBeenCalled();
+                expectSaveButtonDisabled();
+                let btn = wrapper.find('#is-canceling');
+                expect(btn.text()).toBe('Close');
+                done();
+            });
+        });
+    });
+
+    it("Enables 'save' by default in bulk mode", () => {
```
bbpennel, line 1275: Should not allow users to submit if both no changes are selected

PR #1261, opened 2021-06-07
BXC-2639 - Multi-select facets

bbpennel:
https://jira.lib.unc.edu/browse/BXC-2639
https://jira.lib.unc.edu/browse/BXC-3183
Initial server side implementation, still need to update the public ui front end.

* In search requests to select multiple values of the same facet, the values should be separated by `||`. For example: `subject=subject1||subject2`
* Facet retrieval split into a separate service with tests added
* Facet filters in search states are always lists of SearchFacets now, helper methods added to support this.
* Removes dead code and unused parameters related to old implementations of this
* Parent Collection facet should now use the parentCollection field rather than ancestor path


static/js/vue-cdr-access/src/components/facets.vue
```diff
@@ -71,6 +72,17 @@
                     });
                 });
                 return display_list;
+            },
+
+            sortedFacetsList() {
+                return this.facetList.map((facet) => {
+                    if (facet.name === 'PARENT_COLLECTION') {
```
bbpennel, line 79: I'd suggest changing this so it only sorts for the `CONTENT_TYPE` facet instead, since it is the only one with nested child values at present, and we wouldn't want to change the sort of some of the other potential facets.

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/SearchState.java
```diff
@@ -135,11 +132,70 @@ public void setSearchFields(Map<String, String> searchFields) {
...
+    public void setFacet(SearchFacet value) {
+        setFacet(SearchFieldKeys.valueOf(value.getFieldName()), value);
+    }
+
+    /**
+     * Set the value of the specified facet to the provided value
+     * @param key key of the facet to add
+     * @param value value of the facet to set.
+     */
+    @SuppressWarnings("unchecked")
```
lfarrell, line 179: Why is this unchecked?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/MultiSelectFacetListService.java
```diff
@@ -0,0 +1,140 @@
...
+                List<SearchFacet> merged = getMergedHierarchivalFacetValues(facetName, facetEntry.getValue(),
+                        selectedResponse, searchRequest);
+                resultFacets.add(new FacetFieldObject(facetName, merged));
+            } else {
+                resultFacets.add(selectedResponse.getFacetFields().get(0));
+            }
+        }
+
+        resultFacets.sort(searchRequest.getSearchState().getFacetsToRetrieve());
+        return resultResponse;
```
lfarrell, line 106: How are the resultFacets being added to the resultReponse? Are they being updated in place from the FacetFieldList from line 77?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/model/SearchState.java
bbpennel: Its because of `vals = (List<SearchFacet>) value;`. I could probably split this method into two methods, one for `List<SearchFacet>` value and another for everything else

solr-search/src/main/java/edu/unc/lib/dl/search/solr/util/FacetFieldUtil.java
```diff
@@ -41,69 +46,128 @@
...
-    }
-
-    private void addCaseInsensitiveFacetValue(CaseInsensitiveFacet facet, SolrQuery solrQuery) {
-        solrQuery.addFilterQuery(solrSettings.getFieldName(facet.getSearchName()) + ":\""
-                + SolrSettings.sanitize(facet.getSearchValue()) + "\"");
-    }
+        return filterQuery.toString();
+    };
+
+    private Function<SearchFacet, String> genericFacetToFq = (facet) -> {
```
lfarrell, line 161: What's a genericFacet?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/service/MultiSelectFacetListService.java
bbpennel: `searchService.getSearchResults(facetRequest)` at line 75 performs a search and constructs a new result response, which contains a facetFields object populated with the facet values from that first search. I can add a comment that describes the steps here a little more. Effectively, it first issues a query for the search with all the facet filters applied, returning the facet fields which do not have filters applied to them (so if you were filtering by CONTENT_TYPE, the first query would just return PARENT_COLLECTION). Then it issues queries to get the facet values for each facet field which is being filtered, with all the OTHER filters applied but not its own filter (otherwise the only value it'd return would be the one being filtered on).

solr-search/src/test/java/edu/unc/lib/dl/search/solr/service/MultiSelectFacetListServiceIT.java
```diff
@@ -0,0 +1,537 @@
...
+        searchState.setFacetsToRetrieve(FACETS_TO_RETRIEVE);
+
+        SearchRequest request = new SearchRequest(searchState, accessGroups);
+        request.setRootPid(testCorpus.privateFolderPid);
+        request.setApplyCutoffs(false);
+        service.getFacetListResult(request);
+    }
+
+    @Test
+    public void filterByFacetNotBeingRetrievedTest() throws Exception {
```
lfarrell, line 476: What's the difference between this and singleCollectionFacetTest on line 137?

solr-search/src/main/java/edu/unc/lib/dl/search/solr/util/FacetFieldUtil.java
bbpennel: `edu.unc.lib.dl.search.solr.model.GenericFacet`
Any facet field not mapped to a specific class is a generic facet. For example, PARENT_COLLECTION is a generic facet, while CONTENT_TYPE is a MultivaluedHierarchicalFacet. You can see the classes used per facet in search.properties https://github.com/UNC-Libraries/Carolina-Digital-Repository/blob/main/access/src/main/resources/search.properties#L100

solr-search/src/test/java/edu/unc/lib/dl/search/solr/service/MultiSelectFacetListServiceIT.java
bbpennel: This test is verifying that you can filter by a facet which is not in the list of facets being retrieved. It is filtering on PARENT_COLLECTION, which is not in the facetsToRetrieve list, whereas it is in that list in `singleCollectionFacetTest`. I believe this situation didn't quite behave correctly initially.

lfarrell (changes requested):
Not changes per se, but a few questions

static/js/vue-cdr-access/src/components/facets.vue
```diff
@@ -240,17 +245,17 @@
...
-                });
+                const facet_type = this.facetType(value);
+                const current_facet_value = this.selected_facets.filter(f => f.startsWith(facet_type));
 
                 if (current_facet_value.length === 1) {
-                    const joined_facets = `${current_facet_value[0]}||${value.limitToValue}`;
-                    const deduped_facets = Array.from(new Set(joined_facets.split('||')));
-                    return `${deduped_facets.join('||')}`
+                    const selected_facet_parts = current_facet_value[0]
+                        .replace(new RegExp(facet_type), '')
```
bbpennel, line 253: I don't think the regexp is needed here, replacing `facet_type` by itself should work?

lfarrell (approved):

PR #1267, opened 2021-06-23
BXC-3187 - Create common-utils module

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3187

* Splits a new common-utils module out from the metadata module
* A few updates to remove unnecessary dependencies


common-utils/src/test/java/edu/unc/lib/boxc/common/util/DateTimeUtilTest.java
```diff
@@ -78,15 +47,13 @@ public void utcToDateTimeTest() throws Exception {
...
         assertEquals(1, date.getDayOfMonth());
-        assertEquals(17, date.getHourOfDay());
         assertEquals(13, date.getMinuteOfHour());
         assertEquals(14, date.getSecondOfMinute());
 
         date = DateTimeUtil.parseUTCToDateTime("2004-02-04T12:13:14.005");
         assertEquals(2004, date.getYear());
         assertEquals(2, date.getMonthOfYear());
         assertEquals(4, date.getDayOfMonth());
-        assertEquals(17, date.getHourOfDay());
```
lfarrell, line 89: We don't want the hour anymore?

lfarrell (changes requested):

common-utils/src/test/java/edu/unc/lib/boxc/common/util/DateTimeUtilTest.java
bbpennel: The test was failing because of timezone issues in gh actions, although I'm realizing now its because I removed the jodatime setting of the timezone. I've added that and the hour checks back in. I'm kind of confused why the test passed both during and outside of daylight saving since I'd think the hour would be impacted, but I guess it's passed all this time.

I spent a while trying to remove the jodatime dependencies in this class since its the only place in the application that uses it, but java.time doesn't have an extensive ISO8601 parsing utility, which we are using to interpret dates from MODS records.

lfarrell: Ah, the joy of dates

lfarrell (approved):

PR #1272, opened 2021-07-06
Display search parameters as tags

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3198

### DIFF
--- a/static/js/vue-cdr-access/src/components/filterTags.vue
+++ b/static/js/vue-cdr-access/src/components/filterTags.vue
@@ -82,32 +82,32 @@ export default {
             return params;
         },
 
-        formatSearchValue(field, fieldValue) {
-            if (field === undefined || TYPES[fieldValue] === undefined) {
+        formatSearchValue(fieldValue, type) {
+            if (fieldValue === undefined || TYPES[type] === undefined) {
                 return '';
             }
 
-            let display_text = decodeURIComponent(field);
+            let display_text = decodeURIComponent(fieldValue);
 
             // Format time based tags
-            if (fieldValue === 'added' || fieldValue === 'created') {
-                display_text = this._formatTime(field);
+            if (type === 'added' || type === 'created') {
+                display_text = this._formatTime(fieldValue);
             }
 
             let tag_info = {
-                type: fieldValue,
-                type_text: TYPES[fieldValue],
-                original_value: field,
+                type: type,
+                type_text: TYPES[type],
+                original_value: fieldValue,
                 value_text: display_text
             };
 
             // Return non multi-value tags
-            if (fieldValue !== 'format' && fieldValue !== 'collection') {
+            if (type !== 'format' && type !== 'collection') {
                 return tag_info;
             }
 
             // Multi-value tags
-            return this._updateMultiValueTags(display_text, fieldValue, tag_info);
+            return this._updateMultiValueTags(display_text, type, tag_info);
         },
 
         _formatTime(field) {
--- a/static/js/vue-cdr-access/src/components/filterTags.vue
+++ b/static/js/vue-cdr-access/src/components/filterTags.vue
@@ -29,7 +29,7 @@ const TYPES = {
 }
 
 export default {
-    name: "facetTags",
+    name: "filterTags",
 
     mixins: [routeUtils],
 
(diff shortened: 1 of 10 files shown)


static/js/vue-cdr-access/src/components/searchWrapper.vue
```diff
@@ -89,14 +89,8 @@
         methods: {
             retrieveData() {
                 let param_string = `${this.formatParamsString(this.$route.query)}&getFacets=true`;
-                let search_path = 'searchJson';
-
-                this.collection = UUID_REGEX.test(this.$route.path) ? this.$route.path.split('/')[2] : '';
```
bbpennel, line 94: I wonder if we should keep this logic? It allows scoping a search result to containers other than collections, which I personally find kind of helpful, and we might want to expose again later or in conjunction with adding more faceting/search options to the result pages.

static/js/vue-cdr-access/tests/unit/facetTags.spec.js
```diff
@@ -0,0 +1,156 @@
...
+                    }
+                ]
+            }
+        });
+
+        facet_tags = wrapper.findAll('ul li');
+    });
+
+    it("displays selected tags", async () => {
+        wrapper.vm.$router.push('/search/?collection=d77fd8c9-744b-42ab-8e20-5ad9bdf8194e');
```
bbpennel, line 77: for what its worth, the urls i'm seeing in the live interface are in the form `/search?...` rather than `search/?...`

static/js/vue-cdr-access/tests/unit/facetTags.spec.js
```diff
@@ -0,0 +1,156 @@
...
+        await wrapper.vm.$nextTick();
+        expect(wrapper.find('.search-text').text()).toMatch(/testCollection/);
+    });
+
+    it("displays multiple selected tags", async () => {
+        wrapper.vm.$router.push('/search/?collection=d77fd8c9-744b-42ab-8e20-5ad9bdf8194e&format=image');
+        await wrapper.vm.$nextTick();
+        const selected_tags = wrapper.findAll('.search-text');
+        expect(selected_tags.at(0).text()).toMatch(/testCollection/);
```
bbpennel, line 86: Might be good to verify that it is displaying `Collection > testCollection` and `Format > image` to make sure the type labels are present and on the right terms

static/js/vue-cdr-access/tests/unit/facetTags.spec.js
```diff
@@ -0,0 +1,156 @@
...
+        await wrapper.vm.$nextTick();
+        expect(wrapper.find('.search-text').text()).toMatch(/stone/);
+    });
+
+    it("does not display search query if empty", async () => {
+        wrapper.vm.$router.push('/search/?anywhere=');
+        await wrapper.vm.$nextTick();
+        expect(wrapper.find('.search-text').exists()).toBe(false);
+    });
+});
```
bbpennel, line 156: Add tests for one of the date term fields to make sure the formatting is working.

bbpennel: Also add test for searching the title index, and searching both the title and the keyword index at the same time.

bbpennel (changes requested):
This works well. Left a few cleanup comments, plus some extra tests and few minor bugs

static/js/vue-cdr-access/src/components/filterTags.vue
```diff
@@ -82,52 +82,55 @@ export default {
             return params;
         },
 
-        formatSearchValue(field, type) {
-            if (field === undefined || TYPES[type] === undefined) {
+        formatSearchValue(field, fieldValue) {
```
bbpennel, line 85: I think you renamed the wrong variable? It looks like `field` is the value of the field, while `type` (now `fieldValue`) is the name of the parameter from the query string.

static/js/vue-cdr-access/tests/unit/filterTags.spec.js
```diff
@@ -74,83 +74,115 @@ describe('facetTags.vue', () => {
...
     });
 
     it("displays multiple selected tags", async () => {
-        wrapper.vm.$router.push('/search/?collection=d77fd8c9-744b-42ab-8e20-5ad9bdf8194e&format=image');
+        wrapper.vm.$router.push('/search?collection=d77fd8c9-744b-42ab-8e20-5ad9bdf8194e&format=image');
         await wrapper.vm.$nextTick();
         const selected_tags = wrapper.findAll('.search-text');
-        expect(selected_tags.at(0).text()).toMatch(/testCollection/);
-        expect(selected_tags.at(1).text()).toMatch(/image/);
+        expect(selected_tags.at(0).text()).toMatch(/Collection.*?testCollection/s);
```
bbpennel, line 86: Just curious, what does the `/s` do here? I was having a hard time looking it up

bbpennel (changes requested):

static/js/vue-cdr-access/tests/unit/filterTags.spec.js
lfarrell: It allows the . in a regex to match the  \n character. Don't know why it doesn't by default  `¯\_(ツ)_/¯`

bbpennel (approved):

PR #1273, opened 2021-07-06
BXC-3206 - Bag Deposit too many file handles

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3206

* Add executor service to limit the number of files the BagVerifier checks for at once, to prevent it from opening thousands of file handles simultaneously


lfarrell (approved):

lfarrell:
Should there be a pr in boxc-ansible for `job.bagitValidation.workers`?

bbpennel:
@lfarrell it has a default value, so it works without it, but we can override the value via boxc ansible in the future if we need to

PR #1278, opened 2021-07-26
BXC-3190 - Reorganize persistence module

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3190

* Added operations, operations-jms, deposit-utils modules
    * Moved appropriate classes into these modules
    * Purpose of persistence module refocused on persistence of binaries. Added persistence-api module to prevent dependency issues
* Cleaned up dependencies, unused code, unused resources. Removed some usages of classes from activemq that should have been from base java.


services/src/test/resources/spring-test/cdr-client-container.xml
```diff
@@ -186,8 +184,4 @@
         class="org.springframework.web.multipart.commons.CommonsMultipartResolver">
         <property name="maxUploadSize" value="2000000000"/>
```
lfarrell, line 185: Do we want this to be hard coded?

lfarrell (changes requested):
One question, unrelated to the pr.

services/src/test/resources/spring-test/cdr-client-container.xml
bbpennel: Good point, I've created a ticket https://jira.lib.unc.edu/browse/BXC-3222

lfarrell: See https://jira.lib.unc.edu/browse/BXC-3222

lfarrell (approved):

PR #1281, opened 2021-08-05
BXC-3178 - Support migration SIP ingest

bbpennel:
Part of https://jira.lib.unc.edu/browse/BXC-3178

* Adds deposit job to import preconstructed deposit directories
* Adds agent and deposit method constants for migration utility.


lfarrell (approved):
Looks good

PR #1296, opened 2021-09-22
BXC-3261 Revise templates

agoslen:


bbpennel (approved):

static/js/admin/editDescription.js
```diff
@@ -102,7 +102,7 @@ define("editDescription", ["module", "jquery", "jquery-ui", "ace", "xmleditor",
 					},
 					{
 						filename: 'generic.xml',
-						title: "Generic Object",
+						title: "Generic Work",
```
bbpennel, line 105: This form could be used for any kind of object rather than just works, but I assume in practice it should just be used for works? In which case including the word "work" in the title of the form seems fine

PR #1298, opened 2021-10-01
BXC-3203 - Access surrogates support

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3203

* Access surrogates provided to deposit pipeline will be copied to derivatives directory
* BinaryMetadataProcessor will use the path and mimetype of the access surrogate if one is detected
    * Since there is nowhere to store it, the mimetype of the access surrogate is determined using tika
* Renames access_copy to  access_surrogate in most places for clarity/consistency
* access_surrogate datastream is implicitly indexed to solr, along with all other external datastreams.


lfarrell (approved):
Is this using the available access copy, if present, instead of generating new one?

bbpennel:
@lfarrell If an access surrogate is present, this PR makes it so that surrogate will be used to produce the JP2 and thumbnails instead of the original_file datastream. Its very similar to how we produce thumbnails for collections. The access surrogate is never generated by box-c, it is supplied by users (currently only during cdm migrations). I renamed the datastream to ACCESS_SURROGATE to try to differentiate it from other access copies, since we've often referred to the JP2 as the access copy in the past.

lfarrell:
Ah okay. Makes sense

PR #1301, opened 2021-11-01
BXC-3256 - Additional Metadata Types in XML Export

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3256

* Allows for FITS, FITS History, MODS History, and PREMIS Event Logs to be included in the XML export
* Renames the `object/update` tag to `object/datastream`, with an optional `operation="update"` attribute
* Refactored out constants for xml import and export operations
* Updates to DatastreamPids helper methods
* Renamed "Export MODS" to "Export Metadata". Still named "Import MODS" since that is the only importable datastream at present.


bbpennel:
@lfarrell there are currently 5 datastreams that can be exported using this tool at the moment. I didn't attempt to export all of them, I was mainly attempting to verify that one and more than one datastream type could be exported. I could add a test to see if the full inventory of datastreams is exportable rather than various combinations

PR #1309, opened 2021-12-01
BXC-3287 - Filter MODS fields from public display

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3287

* Filters MODS fields from the public display based on a list of configured xpath expressions

### DIFF
--- a/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
+++ b/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
@@ -15,36 +15,6 @@
  */
 package edu.unc.lib.boxc.web.access.controllers;
 
-import static edu.unc.lib.boxc.auth.fcrepo.services.GroupsThreadStore.getAgentPrincipals;
-import static edu.unc.lib.boxc.common.xml.SecureXMLFactory.createSAXBuilder;
-import static edu.unc.lib.boxc.search.api.FacetConstants.MARKED_FOR_DELETION;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.Date;
-import java.util.List;
-
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
-import edu.unc.lib.boxc.web.common.services.XmlDocumentFilteringService;
-import org.jdom2.Document;
-import org.jdom2.JDOMException;
-import org.jdom2.input.SAXBuilder;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.http.HttpStatus;
-import org.springframework.stereotype.Controller;
-import org.springframework.ui.Model;
-import org.springframework.web.bind.annotation.ExceptionHandler;
-import org.springframework.web.bind.annotation.PathVariable;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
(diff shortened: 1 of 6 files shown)


web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -27,6 +27,7 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import edu.unc.lib.boxc.web.common.services.XmlDocumentFilteringService;
```
lfarrell, line 30: Import looks like it's in the wrong place

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/XmlDocumentFilteringService.java
```diff
@@ -0,0 +1,86 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.boxc.web.common.services;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import edu.unc.lib.boxc.model.api.xml.JDOMNamespaceUtil;
```
lfarrell, line 19: Is this import in the right place?

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/XmlDocumentFilteringService.java
```diff
@@ -0,0 +1,86 @@
...
+     * Modifies the provided document to exclude the configured xml fields
+     * @param doc XML document to filter
+     */
+    public void filterExclusions(Document doc) {
+        notNull(doc, "Must provide XML document");
+        for (String xPathString: xPathStrings) {
+            // Instantiating new xpath expressions since they are not thread safe, so cannot be safely reused
+            XPathExpression<Element> xPath = xPathFactory.compile(xPathString, Filters.element(),
+                    null, JDOMNamespaceUtil.MODS_V3_NS);
+            for (Element el: xPath.evaluate(doc)) {
```
lfarrell, line 63: Should el just be renamed to detachEl since it's renamed in the loop?

lfarrell (changes requested):

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/XmlDocumentFilteringService.java
bbpennel: `el` is the element directly matched by the xpath expression, `detachEl` is the element that gets detached from the document. They often aren't the same, since the loop below seeks the topmost parent element that would be left empty by the removal of `el`

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
bbpennel: I think it is a disagreement between eclipse and intellij about the sort order of the imports, but I've got ahead and had intellij organize the imports

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/XmlDocumentFilteringService.java
bbpennel: I think this one is in the right place according to my IDE

lfarrell (approved):

PR #1313, opened 2022-01-04
Always show images in viewer, even if not primary object

lfarrell:
* Display image viewer if any images present.
* Display PDF viewer if no images and pdf file is primary object or work has no primary object.

https://jira.lib.unc.edu/browse/BXC-3385


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -97,6 +101,49 @@ public boolean hasViewableFiles(ContentObjectRecord briefObj, AccessGroupSet pri
         return resp.getResults().getNumFound() > 0;
     }
 
+    /**
+     * Returns true if the file is a pdf and is the primary object or the object has no primary object
+     * @param workObjs
+     * @return
+     */
+    public boolean pdfViewerNeeded(List<ContentObjectRecord> workObjs) {
```
bbpennel, line 109: I think this implementation might run into issues if the work has more than one page worth of files and the primary object isn't on the first page. But I think there are a few performance improvements we can make at the same time.

You should only need to pass the workObject itself into this method, rather than querying for all its children beforehand. We don't need to query for children if the work uses the UV (this method doesn't need to be called in that case), and we don't need to query for them if the Work itself passes `hasPdf` (which implies that the primaryObject is a pdf). 

So that leaves us with works without primary objects. For that scenario, I believe you can query for the first child (so rows=1). Then you can check `searchResult.getResultCount()` to get the total number of results for the query, and if it is > 1 then its a multi-file work and you would want to return false. If its == 1, then you would return true if hasPdf passes for that child.

I really wish this class had a unit test... It might be worth while to add one to be honest. Let me know if I can help with doing so, i just suspect it will make writing this functionality easier

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -178,9 +182,16 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
             throw new InvalidRecordRequestException();
         }
 
-        // Retrieve the objects record from Solr
-        SimpleIdRequest idRequest = new SimpleIdRequest(pid, principals);
-        ContentObjectRecord briefObject = queryLayer.getObjectById(idRequest);
+        // Retrieve the object's record from Solr
+        SearchRequest searchRequest = new SearchRequest();
```
bbpennel, line 186: I don't think I understand why this change was made for how the object's record itself is retrieved?

lfarrell: I think I just borrowed the syntax from another class. I can switch it back.

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -219,7 +221,40 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
...
+                // Check if primary object is a pdf
+                pdfViewerNeeded = accessCopiesService.hasViewablePdf(briefObject, principals);
+
+                // Check if first child is a pdf
+                if (!pdfViewerNeeded) {
+                    // Get child objects
+                    SearchRequest searchRequest = new SearchRequest();
+                    SearchState searchState = searchStateFactory.createSearchState();
+                    searchState.setFacetsToRetrieve(null);
+                    searchState.setIgnoreMaxRows(true);
```
bbpennel, line 239: Can you replace this setting with setting the number of rows to 1? There's no need for any rows past the first one.

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -219,7 +221,40 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
...
+                    searchRequest.setRootPid(pid);
+                    searchRequest.setApplyCutoffs(true);
+                    SearchResultResponse resultResponse = queryLayer.performSearch(searchRequest);
+
+                    if (resultResponse.getResultCount() == 1) {
+                        ContentObjectRecord childObject = resultResponse.getResultList().get(0);
+                        pdfViewerNeeded = accessCopiesService.hasViewablePdf(childObject, principals);
+
+                        if (pdfViewerNeeded) {
+                            model.addAttribute("filePid",childObject.getId());
```
bbpennel, line 251: I'd prefer if we scoped the name of this attribute, `filePid` could collide with other attributes in the future. Renaming it something like `pdfViewerPid` would work.

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -0,0 +1,100 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+/**
+ * @author lfarrell
+ */
+public class AccessCopiesServiceTest {
+    @Mock
```
bbpennel, line 41: Here you're mocking PermissionsHelper, but in the `init` method you are instantiating a new instance, so the mocked instance is lost. So I think you can remove this annotation

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -97,6 +99,11 @@ public boolean hasViewableFiles(ContentObjectRecord briefObj, AccessGroupSet pri
         return resp.getResults().getNumFound() > 0;
     }
 
+    public boolean hasViewablePdf(ContentObjectRecord workObj, AccessGroupSet principals) {
```
bbpennel, line 102: This method accepts work and file objects, so i'd suggest renaming `workObj` to `contentObj`. Also, could you add javadoc to this method to describe what it does?

You've moved most of the logic out of this method up into the controller, which unfortunately means its not being tested. Up to you if you want to change it or not, but this method could be renamed `getViewablePdfFilePid`, and the method could perform lines 236-255 from the controller, returning the ID of the file object to use in the viewer or null.

bbpennel (changes requested):

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
lfarrell: Yeah, I think a lot of it may need to be moved back over to the service for checking which file to download.

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -97,6 +99,65 @@ public boolean hasViewableFiles(ContentObjectRecord briefObj, AccessGroupSet pri
...
+     * @param principals
+     * @return boolean
+     */
+    public boolean hasViewablePdf(ContentObjectRecord contentObj, AccessGroupSet principals) {
+        return permissionsHelper.hasOriginalAccess(principals, contentObj) &&
+                DatastreamUtil.originalFileMimetypeMatches(contentObj, "application/(x-)?pdf");
+    }
+
+    /**
+     * Retrieves Retrieves the first ContentObjectRecord of a work and
```
bbpennel, line 114:
```suggestion
     * Retrieves the first ContentObjectRecord of a work and
```

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -0,0 +1,180 @@
...
+        accessCopiesService.setFacetFieldUtil(facetFieldUtil);
+        accessCopiesService.setSolrClient(solrClient);
+
+        when(solrClient.query(any(SolrQuery.class))).thenReturn(queryResponse);
+        when(queryResponse.getResults()).thenReturn(solrDocumentList);
+        when(solrDocumentList.getNumFound()).thenReturn(1L);
+    }
+
+    @Test
+    public void testHasViewAblePdf() {
```
bbpennel, line 122: "Viewable" is one word, so "Able" shouldn't be capitalized in any of these tests

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -0,0 +1,180 @@
...
+        assertFalse("Work has viewable PDF content",
+                accessCopiesService.hasViewablePdf(mdObject, principals));
+    }
+
+    @Test
+    public void testHasViewAblePdfPid() {
+        hasPermissions(mdObject, true);
+
+        List<ContentObjectSolrRecord> mdObjects = new ArrayList<>();
+        mdObjects.add(mdObject);
```
bbpennel, line 147: can use singletonList or Arrays.asList

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Test
+    public void testHasViewAblePdf() {
+        hasPermissions(mdObject, true);
+        assertTrue("Work does not have PDF viewable content",
+                accessCopiesService.hasViewablePdf(mdObject, principals));
+    }
+
+    @Test
+    public void testDoesNotHasViewAblePdf() {
```
bbpennel, line 129: `testDoesNotHaveViewablePdf` or `testNonPdfHasViewablePdf`

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -0,0 +1,180 @@
...
+        assertNotNull(filePid);
+        assertEquals(filePid, mdObject.getId());
+    }
+
+    @Test
+    public void testDoesNotHaveViewAblePdfPidOneContentObject() {
+        hasPermissions(mdObjectImg, true);
+
+        List<ContentObjectSolrRecord> mdObjects = new ArrayList<>();
+        mdObjects.add(mdObjectImg);
```
bbpennel, line 160: an use singletonList or Arrays.asList

bbpennel (changes requested):
This looks good, just some pedantic cleanup

PR #1316, opened 2022-01-10
BXC-3157/3158 - Handle invalid pid/object not found

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3157
https://jira.lib.unc.edu/browse/BXC-3158

* Broadens `ExceptionHandler` mappings for common access exceptions to apply to most controller methods
* Old record URLs redirect to current record URL format, stripping out "uuid:" prefix and whitespace or switching from query params to path (primary cause of BXC-3158, mostly from finding aid links)
* Trim input to `PIDs.get` method to avoid whitespace issues when parsing ids


web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -98,17 +102,33 @@
     @Autowired
     private RepositoryObjectLoader repositoryObjectLoader;
 
-    @RequestMapping(value = "/{pid}", method = RequestMethod.GET)
-    public String handleRequest(@PathVariable("pid") String pid, Model model, HttpServletRequest request) {
-        return getFullRecord(pid, model, request);
+    @GetMapping("/{pid}")
+    public ModelAndView handleRequest(@PathVariable("pid") String pid, Model model, HttpServletRequest request) {
```
sharonluong, line 106: how does this go from being a string to a ModelAndView? What is ModelAndView and how does it work?

bbpennel: Its a wrapper for passing along the view routing and model information for a request. The Spring MVC framework is executing these controller methods and it knows how to receive quite a few object types as the return type from controller methods, including `String`s and `ModelAndView`. I switched return types here because it allows me to set a different status code some of the time. There are some other ways to do that, but this seemed preferable.

sharonluong (approved):

PR #1317, opened 2022-01-12
Update when download button displays and move audio file check to service

lfarrell:
* Update when download button displays
* Move audio file checks to service

https://jira.lib.unc.edu/browse/BXC-3384

### DIFF
--- a/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
+++ b/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
@@ -252,6 +252,18 @@ public class FullRecordController extends AbstractErrorHandlingSearchController 
             }
 
             model.addAttribute("pdfViewerNeeded", pdfViewerNeeded);
+
+            // Check for audio file
+            boolean audioPlayerNeeded = false;
+            if (!imageViewerNeeded && !pdfViewerNeeded) {
+                audioPlayerNeeded = accessCopiesService.hasPlayableAudioFile(briefObject, principals);
+            }
+            model.addAttribute("audioPlayerNeeded", audioPlayerNeeded);
+
+
+            // Get the file to download
+            String dataFileUrl = accessCopiesService.getDownloadPath(briefObject, principals);
+            model.addAttribute("dataFileUrl", dataFileUrl);
         }
 
         List<String> objectStatus = briefObject.getStatus();
--- a/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
+++ b/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
@@ -238,12 +238,14 @@ public class FullRecordController extends AbstractErrorHandlingSearchController 
             boolean pdfViewerNeeded = false;
 
             if (!imageViewerNeeded) {
+                String pdfMimeTypeRegx = "application/(x-)?pdf";
                 // Check if primary object is a pdf
-                pdfViewerNeeded = accessCopiesService.hasViewablePdf(briefObject, principals);
+                pdfViewerNeeded = accessCopiesService.hasDatastreamContent(briefObject, principals, pdfMimeTypeRegx);
 
                 // Check if first child is a pdf
                 if (!pdfViewerNeeded) {
-                    String childObjectPid = accessCopiesService.getViewablePdfFilePid(briefObject, principals);
+                    String childObjectPid = accessCopiesService.getDatastreamPid(briefObject, principals,
+                            pdfMimeTypeRegx);
                     if (childObjectPid != null) {
                         pdfViewerNeeded = true;
                         model.addAttribute("pdfViewerPid", childObjectPid);
@@ -256,10 +258,9 @@ public class FullRecordController extends AbstractErrorHandlingSearchController 
             // Check for audio file
             boolean audioPlayerNeeded = false;
             if (!imageViewerNeeded && !pdfViewerNeeded) {
-                audioPlayerNeeded = accessCopiesService.hasPlayableAudioFile(briefObject, principals);
+                audioPlayerNeeded = accessCopiesService.hasPlayableAudio(briefObject, principals);
             }
             model.addAttribute("audioPlayerNeeded", audioPlayerNeeded);
-
 
             // Get the file to download
             String dataFileUrl = accessCopiesService.getDownloadUrl(briefObject, principals);
--- a/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
+++ b/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
@@ -70,6 +70,8 @@ import java.util.List;
 import static edu.unc.lib.boxc.auth.fcrepo.services.GroupsThreadStore.getAgentPrincipals;
 import static edu.unc.lib.boxc.common.xml.SecureXMLFactory.createSAXBuilder;
 import static edu.unc.lib.boxc.search.api.FacetConstants.MARKED_FOR_DELETION;
+import static edu.unc.lib.boxc.web.common.services.AccessCopiesService.AUDIO_MIMETYPE_REGEX ;
+import static edu.unc.lib.boxc.web.common.services.AccessCopiesService.PDF_MIMETYPE_REGEX;
 
 /**
  * Controller which retrieves data necessary for populating the full record page, retrieving supplemental information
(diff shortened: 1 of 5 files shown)


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -125,6 +136,41 @@ public String getViewablePdfFilePid(ContentObjectRecord briefObj, AccessGroupSet
...
+
+    /**
+     * Get the path of the file that can be downloaded
+     * @param contentObjectRecord
+     * @param principals
+     * @return
+     */
+    public String getDownloadPath(ContentObjectRecord contentObjectRecord, AccessGroupSet principals) {
+        ContentObjectRecord contentObj = getContentObject(contentObjectRecord, principals);
+        if (contentObj != null && permissionsHelper.hasOriginalAccess(principals, contentObj)) {
```
bbpennel, line 167: I think hasOriginalAccess is already being executed on contentObj at line 152?

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -125,6 +136,41 @@ public String getViewablePdfFilePid(ContentObjectRecord briefObj, AccessGroupSet
...
+     * @param briefObj
+     * @param principals
+     * @return
+     */
+    public ContentObjectRecord getContentObject(ContentObjectRecord briefObj, AccessGroupSet principals) {
+        if (permissionsHelper.hasOriginalAccess(principals, briefObj)) {
+            return briefObj;
+        }
+
+        ContentObjectRecord contentObj = getChildFileObject(briefObj, principals);
```
bbpennel, line 151: I'm realizing that getChildFileObject should return null immediately if the briefObj it is passed is anything other than a work, otherwise this call will attempt to find the children of file objects. FileObjects don't have queryable children according to the solr index, they just have datastreams.

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -125,6 +136,41 @@ public String getViewablePdfFilePid(ContentObjectRecord briefObj, AccessGroupSet
...
+        if (contentObj != null && permissionsHelper.hasOriginalAccess(principals, contentObj)) {
+            return contentObj;
+        }
+
+        return null;
+    }
+
+    /**
+     * Get the path of the file that can be downloaded
```
bbpennel, line 160:
```suggestion
     * Get the path of the original_file datastream within contentObjectRecord that can be downloaded, or null if no appropriate original_file is present
```

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -125,6 +136,41 @@ public String getViewablePdfFilePid(ContentObjectRecord briefObj, AccessGroupSet
...
+     * @param contentObjectRecord
+     * @param principals
+     * @return
+     */
+    public String getDownloadPath(ContentObjectRecord contentObjectRecord, AccessGroupSet principals) {
```
bbpennel, line 165: `getDownloadUrl` would be more consistent with the naming elsewhere and the content of the returned value.

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -171,6 +194,52 @@ public void testDoesNotHaveViewablePdfPidMultipleFileObjects() {
...
+
+        List<ContentObjectSolrRecord> resultList = Collections.singletonList(mdObject);
+        when(queryResponse.getBeans(ContentObjectSolrRecord.class)).thenReturn(resultList);
+
+        String downloadUrl = accessCopiesService.getDownloadPath(noOriginalFileObj, principals);
+        assertEquals("indexablecontent/" + mdObject.getId(), downloadUrl);
+    }
+
+    @Test
+    public void testDoesNotHasDownloadUrl() {
```
bbpennel, line 217:
```suggestion
    public void testDoesNotHaveDownloadUrl() {
```

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -94,6 +101,22 @@ public void init() throws IOException, SolrServerException {
...
 
+        mdObjectAudio = new ContentObjectSolrRecord();
+        mdObjectAudio.setId("45c8d1c5-14a1-4ed3-b0c0-6da67fa5f6d1");
+        List<String> audioDatastreams = Collections.singletonList(
+                ORIGINAL_FILE.getId() + "|audio/mpeg|file.mp3|mp3|766|urn:sha1:checksum|");
+        mdObjectAudio.setDatastream(audioDatastreams);
+
+        noOriginalFileObj = new ContentObjectSolrRecord();
+        noOriginalFileObj.setId("45c8d1c5-14a1-4ed3-b0c0-6da67fa5f6d1");
+        hasPermissions(noOriginalFileObj, true);
```
bbpennel, line 112: Did you mean to create permission to this object all the time? The others are granted within test methods

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -171,6 +194,52 @@ public void testDoesNotHaveViewablePdfPidMultipleFileObjects() {
...
+        hasPermissions(mdObjectXml, true);
+
+        List<ContentObjectSolrRecord> resultList = Collections.singletonList(mdObjectXml);
+        when(queryResponse.getBeans(ContentObjectSolrRecord.class)).thenReturn(resultList);
+
+        String downloadUrl = accessCopiesService.getDownloadPath(noOriginalFileObj, principals);
+
+        assertEquals("", downloadUrl);
+    }
+
```
bbpennel, line 228: Could you add a test to verify that a work that contains multiple files and no primary object doesn't get a download url? And perhaps a test to verify that a work with multiple files and a primary object does get the right one.

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -94,6 +101,22 @@ public void init() throws IOException, SolrServerException {
...
+        List<String> audioDatastreams = Collections.singletonList(
+                ORIGINAL_FILE.getId() + "|audio/mpeg|file.mp3|mp3|766|urn:sha1:checksum|");
+        mdObjectAudio.setDatastream(audioDatastreams);
+
+        noOriginalFileObj = new ContentObjectSolrRecord();
+        noOriginalFileObj.setId("45c8d1c5-14a1-4ed3-b0c0-6da67fa5f6d1");
+        hasPermissions(noOriginalFileObj, true);
+
+        mdObjectXml = new ContentObjectSolrRecord();
+        mdObjectXml.setId("45c8d1c5-14a1-4ed3-b0c0-6da67fa5f6d1");
```
bbpennel, line 115: You should switch to using different ids for each test object, otherwise you won't be able to tell if its generating download URLs to the wrong object. You could use `UUID.randomUUID().toString()` or hard code the values.

web-access-app/src/main/webapp/WEB-INF/jsp/fullRecord/aggregateRecord.jsp
```diff
@@ -145,7 +142,7 @@
                     <c:when test="${pdfViewerNeeded}">
                         <c:import url="fullRecord/pdfViewer.jsp" />
                     </c:when>
-                    <c:when test="${hasOriginalFileAccess && cdr:originalFileMimetypeMatches(briefObject, 'audio/(x-)?mpeg(-?3)?')}">
+                    <c:when test="${not empty dataFileUrl && audioPlayerNeeded}">
```
bbpennel, line 145: just curious why audioPlayerNeeded needs to check for dataFileUrl first when pdfViewerNeeded doesn't?

bbpennel (changes requested):
Looks good over all, some questions and comments

web-access-app/src/main/webapp/WEB-INF/jsp/fullRecord/aggregateRecord.jsp
lfarrell: Hmm, I don't think it actually does need it. It should already be checked in hasPlayableAudioFile()

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
lfarrell: Nope

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
lfarrell: Not sure how to add a test for a work with multiple files and a primary object, as it should never query for the files if the primary object is set.

web-access-app/src/main/webapp/WEB-INF/jsp/fullRecord/aggregateRecord.jsp
lfarrell: Apparently shows the player if removed, but displays a null message in the player if no primary object set.

bbpennel: For pdfs, it checks the primary object and then the first child if there is no primary object. You may need to do something similar for audio files. The two implementations would be pretty much identical aside from what mimetype to check for. Maybe they could use the same method internally and then hasPlayableAudioFile would pass the mimetype pattern in to the shared method? 

Alternatively, hasViewablePdf could be replaced by "getViewableDatastream" which would return the name of the datastream which is viewable. Internally it would do the first child solr query once and check against the mimetype patterns. Might be more scalable if we add more viewers, but on the other hand I think we're hoping that audio and video will become part of the UV display in the future.

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
bbpennel: Okay, probably fine to leave it off. We could sort of set it up (mostly by having the total result count set to 2), but might not make sense to go through the effort at the moment

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -238,12 +238,14 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
             boolean pdfViewerNeeded = false;
 
             if (!imageViewerNeeded) {
+                String pdfMimeTypeRegx = "application/(x-)?pdf";
                 // Check if primary object is a pdf
-                pdfViewerNeeded = accessCopiesService.hasViewablePdf(briefObject, principals);
+                pdfViewerNeeded = accessCopiesService.hasDatastreamContent(briefObject, principals, pdfMimeTypeRegx);
```
bbpennel, line 243: I'm finding the various slightly different helper methods kind of hard to follow and i'm not sure why audio behaves differently than pdfs? 

How about using `getDatastreamPid` for both pdfs and audio. `getDatastreamPid` will return the pid of the primary object if it has one, and if not then it will do the looking up children stuff it already does. Then for model attributes you could have "pdfViewerNeeded", "audioPlayerNeeded", and "viewerPid", where the latter is the pid of the file to display regardless of viewer. Or you could combine all the flag attributes into one, and have "viewerType" with values "uv", "pdf", "audio", and "viewerPid". We only show one viewer at a time so we don't really need to allow for multiple viewer attributes to be set.

So
```
String viewerType = null;
String viewerPid = null;
boolean imageViewerNeeded = accessCopiesService.hasViewableFiles(briefObject, principals);
if (imageViewerNeeded) {
    viewerType = "uv";
} else {
    // Check for PDF to display
    viewerPid = accessCopiesService.getDatastreamPid(briefObject, principals, pdfMimeTypeRegx);
    if (viewerPid != null) {
        viewerType = "pdf";
    } else {
        // Check for viewable audio file
        viewerPid = accessCopiesService.getDatastreamPid(briefObject, principals, audioMimeTypeRegx);
        if (viewerPid != null) {
            viewerType = "audio";
        }
    }
}
model.addAttribute("viewerType", viewerType);
model.addAttribute("viewerPid", viewerPid);
```

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -238,12 +238,14 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
             boolean pdfViewerNeeded = false;
 
             if (!imageViewerNeeded) {
+                String pdfMimeTypeRegx = "application/(x-)?pdf";
```
bbpennel, line 241: Might be good to have all the mimetype regex patterns as constants, possibly down in AccessCopiesService

bbpennel (approved):

PR #1321, opened 2022-01-18
Decide which thumbnail to show

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3386


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -137,6 +137,40 @@ public String getDatastreamPid(ContentObjectRecord briefObj, AccessGroupSet prin
...
 
+    /**
+     * Gets first object with an original file to use as thumbnail
+     * Returns null if briefObject has an original file, as we don't need to add the object twice.
+     * @param briefObj
+     * @param principals
+     * @return
+     */
+    public ContentObjectRecord getThumbnailObject(ContentObjectRecord briefObj, AccessGroupSet principals) {
+        if (permissionsHelper.hasOriginalAccess(principals, briefObj)) {
```
bbpennel, line 148: Thumbnail access is based off of the viewMetadata permission rather than original access, you can see it here:
https://github.com/UNC-Libraries/box-c/blob/main/auth-api/src/main/java/edu/unc/lib/boxc/auth/api/services/DatastreamPermissionUtil.java#L48
And there's actually a helper method for checking thumbnail permissions specifically `permissionHelper.hasThumbnailAccess(principals, briefObj);`

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -259,6 +259,12 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
             // Get the file to download
             String dataFileUrl = accessCopiesService.getDownloadUrl(briefObject, principals);
             model.addAttribute("dataFileUrl", dataFileUrl);
+
+            // Get thumbnail object
+            ContentObjectRecord thumbnailObject = accessCopiesService.getThumbnailObject(briefObject, principals);
```
bbpennel, line 264: Rather than make this determination in the FullRecordController, I think we should move this over to the `/services/api/thumb/${pid}` endpoints. The reason being that we will probably need to embed thumbnails into finding aids and other sites, which will be making requests directly against the services api, and for archival folders it will be pointing at the work for the top level thumbnail I believe. If you move the logic over to there, then I don't think any changes will be needed in the public UI directly.

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -137,6 +137,40 @@ public String getDatastreamPid(ContentObjectRecord briefObj, AccessGroupSet prin
...
+        }
+
+        // Check if there's only one file
+        ContentObjectRecord contentObj = getChildFileObject(briefObj, principals);
+        if (contentObj != null && permissionsHelper.hasOriginalAccess(principals, contentObj)) {
+            return contentObj;
+        }
+
+        // Check other children for thumbnails
+        QueryResponse childContentObjs = getChildFileObjects(briefObj, principals, MAX_FILES);
```
bbpennel, line 159: Instead of querying for all the children, I believe you can execute a solr query that limits to files which have one of the thumbnail datastreams and which the user has permission to view. For query based on the presence of a particular datastream, see:
https://github.com/UNC-Libraries/box-c/blob/main/web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java#L227

For the permission restriction, the readGroup field will actually tell you this since it reflects if the user has viewMetadata permissions, so simply doing a normal permission check (adding the principals to a search request) should be sufficient in the case of thumbnails. Might want to add a comment to the method to make that clear though.

I believe this query could be limited to 1 result, and it would also double for the case where there is only one child, so you might be able to remove that part of the method.

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -271,6 +272,33 @@ public void doesNotHavePlayableAudiofile() {
...
+    public void noPrimaryObjThumbnailMultipleFiles() {
+        hasPermissions(noOriginalFileObj, true);
+        hasPermissions(mdObjectXml, true);
+        hasPermissions(mdObjectImg, true);
+        List<ContentObjectSolrRecord> resultList = Arrays.asList(mdObjectXml, mdObjectImg);
+        when(queryResponse.getBeans(ContentObjectSolrRecord.class)).thenReturn(resultList);
+        when(queryResponse.getResults().size()).thenReturn(resultList.size());
+        assertEquals(mdObjectImg, accessCopiesService.getThumbnailObject(noOriginalFileObj, principals));
+    }
+
```
bbpennel, line 301: I feel like there should be a test for a work with files, but none with thumbnails. But if its all happening in a solr query then it might be hard to represent...

bbpennel (changes requested):

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
lfarrell: Would all these calls be to the file object? Moving the check over seems to break file level thumbnails and shows the wrong thumbnail for works without primary objects since the thumbnail brief object is whatever the original is
```
<c:set var="thumbnailObject" value="${thumbnailObject}" scope="request" />
<c:import url="common/thumbnail.jsp">
      <c:param name="target" value="file" />
      <c:param name="size" value="large" />
</c:import>
```

bbpennel: I think you'll need to add a check up front in `getThumbnailObject` to see if the object being requested is a WorkObject. If it is, then you would need to do the logic to check for a primary object and fall back to querying for children with thumbnails. For any other type (FileObjects, Collections with thumbnails, etc), you would only want to check the object itself.

This is a functionality change for the thumbnail endpoint, which previously didn't work for WorkObjects directly. So I guess we will need to make changes in the public UI to get it to request the Works thumbnail rather than its primary objects. We would want to switch to this behavior for search results too. I'm not quite sure what the impact of this change will be on how the placeholders get displayed. It might still work?

bbpennel:
Replaced by https://github.com/UNC-Libraries/box-c/pull/1333

PR #1330, opened 2022-02-07
BXC-3434 - Fix advanced search contributor/subject/title searches

bbpennel:
* Fix bug that caused contributor/subject/title searches to search all fields. 
* Refactor addSearchFields to perform term searches other than keyword/anywhere searches using fq parameters.


sharonluong (approved):
Had to look at the file to see how much cleaner it was, good job! Also manually joining instead of using `String.join` is 😱

bbpennel:
Yeah, `String.join` wasn't available until java 8, but there were third party libraries we include which had their own join methods, so I don't really remember why I decided to roll my own here.

PR #1331, opened 2022-02-08
Remove unused JS and JSP from public interface

bbpennel:
* Deletes unused files, and squashes a few files that were only a line or two and only used in one place.


sharonluong (approved):
nice work!

PR #1333, opened 2022-02-09
BXC-3386 - Thumbnail selection updates

bbpennel:


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -103,74 +109,99 @@ public boolean hasViewableFiles(ContentObjectRecord briefObj, AccessGroupSet pri
...
+    public String getThumbnailId(ContentObjectRecord contentObjectRecord, AccessGroupSet principals,
+                                 boolean checkChildren) {
+        // Find thumbnail datastream recorded directly on the object, if present
+        var thumbId = DatastreamUtil.getThumbnailOwnerId(contentObjectRecord);
+        if (thumbId != null) {
+            log.debug("Found thumbnail object directly assigned to object {}", thumbId);
+            return thumbId;
+        }
+        // Don't need to check any further if object isn't a work or doesn't contain files with thumbnails
+        if (!ResourceType.Work.name().equals(contentObjectRecord.getResourceType())
```
lfarrell, line 160: Should this be above line 154, so it skips the check for most object types?

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -103,74 +109,99 @@ public boolean hasViewableFiles(ContentObjectRecord briefObj, AccessGroupSet pri
...
+        if (record == null) {
+            return;
+        }
+        record.setThumbnailId(getThumbnailId(record, principals, checkChildren));
+    }
 
-        return "";
+    public void populateThumbnailIds(List<ContentObjectRecord> records, AccessGroupSet principals,
+                                     boolean checkChildren) {
+        for (var record: records) {
```
lfarrell, line 202: Should there be a space after record?

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -103,74 +109,99 @@ public boolean hasViewableFiles(ContentObjectRecord briefObj, AccessGroupSet pri
...
+        }
+        if (!checkChildren) {
+            log.debug("Not checking children for work {}, so using self as thumbnail id", contentObjectRecord.getId());
+            return contentObjectRecord.getId();
+        }
+
+        SolrQuery query = buildFirstChildQuery(contentObjectRecord, principals);
+        // Limit query to just children which have a thumbnail datastream
+        query.addFilterQuery(solrSettings.getFieldName(SearchFieldKey.DATASTREAM.name()) + ":"
+                + DatastreamType.THUMBNAIL_LARGE.getId() + "|*");
```
lfarrell, line 174: Is anything with a thumbnail always going to have a large thumbnail?

lfarrell (changes requested):
The newish "var" keyword makes me think there's JS mixed in.

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
bbpennel: I keep forgetting whether we do spaces there or not, i'll add one

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
bbpennel: I believe so, we don't have any workflows that produce just one size thumbnail. I went with the large size here since there were some existing parts where it was defaulting to large if the size wasn't specified, although it did mess up one test that was only populating small thumbnails in the datastream field

PR #1343, opened 2022-02-15
Fix bug preventing file list from populating

bbpennel:
Don't exclude file objects when performing a list query and rollup is turned off, otherwise can't list children in works. This was a side effect of https://github.com/UNC-Libraries/box-c/pull/1334


bbpennel:
We have tests of the vuejs component that renders the file list, but we won't have a full integration test that verifies that it works against a live instance of the java backend application. I have been wondering about whether we should take the step of starting to add those to boxc, or whether we should depend on the cypress tests.

lfarrell (approved):

PR #1345, opened 2022-02-16
Fix File resource type removal

bbpennel:
Prevent all collections from SearchSettings from being modifiable, and initialize resourceType lists in search states from new ArrayLists so they can be safely modified. Addresses bug where a blank advanced search would cause the File object type to be removed from default resource types list


sharonluong (approved):
looks good!

PR #1350, opened 2022-02-21
Update oral histories templates according to MAP changes

krblount:


agoslen (changes requested):
Update so that subject top-level element is repeated for each subject type

agoslen (approved):
looks good

PR #1352, opened 2022-02-22
Only index subject names and topics

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3345


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -274,8 +243,14 @@ private void extractSubjects(Element mods, IndexDocumentBean idb) {
...
                 for (Element subjectEl: subjectParts) {
-                    if (subjectEl.getChildren().size() == 0) {
-                        subjects.add(subjectEl.getValue());
+                    String subjectName = subjectEl.getName();
+
+                    if (subjectName.equals("name")) {
+                        subjects.add(formatName(subjectEl));
+                    }
+
+                    if (subjectEl.getChildren().size() == 0 && subjectName.equals("topic")) {
```
bbpennel, line 252: us `isEmpty()` instead of comparing to 0.

static/js/vue-cdr-access/src/components/filterTags.vue
```diff
@@ -24,7 +24,7 @@ const TYPES = {
     created: 'Date Created',
     contributorIndex: 'Contributor',
     format: 'Format',
-    subjectIndex: 'Subject',
```
bbpennel, line 27: I think we want to also allow subjectIndex as a tag (it could also display as "Subject"), since users will be able to facet or search subjects.

indexing-solr/src/test/resources/datastream/inventoryMods.xml
```diff
@@ -6,6 +6,13 @@
   <mods:genre>Dissertation</mods:genre>
   <mods:subject>
     <mods:topic>Test resource</mods:topic>
+    <mods:name>
```
bbpennel, line 9: Checked with anna about this, it sounds like if there were multiple subjects on a record they would be under separate `mods:subject` tags, but if this situation ever did occur the behavior in this PR of creating multiple solr subject entries is what we would want. So maybe keep this block as is, but also add a second `mods:subject/mods:topic` to verify that it will work with multiple top level subject tags?

bbpennel (changes requested):

indexing-solr/src/test/resources/datastream/inventoryMods.xml
```diff
@@ -6,6 +6,7 @@
   <mods:genre>Dissertation</mods:genre>
   <mods:subject>
     <mods:topic>Test resource</mods:topic>
+    <mods:topic>Test resource two</mods:topic>
```
bbpennel, line 9: sorry, i meant moving this to its own mods:subject tag, so you'd have
```
<mods:subject>
    <mods:topic>Test resource</mods:topic>
    <mods:name>
      <mods:namePart type="family">rules</mods:namePart>
      <mods:namePart type="given">boxy</mods:namePart>
    </mods:name>
</mods:subject>
<mods:subject>
    <mods:topic>Test resource two</mods:topic>
</mods:subject>

bbpennel (approved):

PR #1354, opened 2022-02-23
WIP: Add facet for date created

lfarrell:


search-solr/src/main/java/edu/unc/lib/boxc/search/solr/responses/SearchResultResponse.java
```diff
@@ -73,6 +74,14 @@ public void setResultCount(long resultCount) {
         this.resultCount = resultCount;
     }
 
+    public String getMinimumSearchYear() {
```
bbpennel, line 77: I feel like this should be minimumDateCreatedYear to make it clear what field it contains. Although i'm also sort of wondering if we'd want a more generic name and to make it into a map or something in case we need to present multiple date filters at some point. But that is probably overkill at this point, we could always rename is later

bbpennel (changes requested):
Looks good overall, left some suggestions

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -439,4 +444,10 @@ private String formatName(Element nameEl) {
         }
         return nameValue;
     }
+
+    private String setDateYear(Date date) {
```
bbpennel, line 448: Just due to java naming conventions I'd prefer not to use "set" here, could you name this something like "extractDateYear"

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MultiSelectFacetListService.java
```diff
@@ -119,6 +121,28 @@ public SearchResultResponse getFacetListResult(SearchRequest searchRequest) {
...
+    /**
+     * Finds and returns the minimum year created for the given query
+     * @param selectedState
+     * @param originalRequest
+     * @return String
+     */
+    public String getMinimumDateCreatedYear(SearchState selectedState, SearchRequest originalRequest) {
+        selectedState.setRowsPerPage(1);
+        selectedState.setResultFields(Collections.singletonList(SearchFieldKey.DATE_CREATED_YEAR.name()));
+        selectedState.setSortNormalOrder(false);
```
bbpennel, line 133: You still need to specify the sort type in addition to reversing the order, so add in
```
selectedState.setSortType("dateCreated");
```

search-api/src/main/java/edu/unc/lib/boxc/search/api/requests/SearchState.java
```diff
@@ -290,15 +290,18 @@ public RangePair(String leftHand, String rightHand) {
         }
 
         public RangePair(RangePair rangePair) {
-            this.leftHand = rangePair.getLeftHand();
-            this.rightHand = rangePair.getRightHand();
+            String leftHand = rangePair.getLeftHand();
```
bbpennel, line 293: Seems like we don't need to set to local variables first before setting the instance variables, so might be able to revert changes to this method.

search-api/src/main/java/edu/unc/lib/boxc/search/api/requests/SearchState.java
```diff
@@ -323,6 +327,16 @@ public String toString() {
             }
             return leftHand + "," + rightHand;
         }
+
+        private void checkValidRangePair(String start, String end) {
+            // As long as one side is set we're good to go.
+            if (start != null && end != null) {
+                if (Integer.parseInt(start) >= Integer.parseInt(end)) {
```
bbpennel, line 334: if either start or end is null then `Integer.parseInt` will throw a NumberFormatException. So I think we'll need to individually check each variable before parsing it. We may need to add some unit tests for RangePair so we can make sure its behaving as expected (breaking it out into its own file would be fine as well). 

We need to verify:
1. if either start or end is set to anything other than null or a number, that it will throw an exception (IllegalArgumentException is fine)
2. leftHand can be set to a valid value while rightHand is set to null, via the setters AND the string splitting constructor (plus the reverse)
3. if leftHand is greater than rightHander, that an IllegalArgumentException is thrown. Need to test this via setters and constructor.
4. that both the left and right hand can be set to valid integers

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
```diff
@@ -272,6 +272,12 @@ private void populateQueryableFields(SearchState searchState, Map<String,String[
             }
             if (searchSettings.searchableFields.contains(key)) {
                 searchFields.put(key, value);
+            } else if (SearchSettings.FIELDS_RANGE_SEARCHABLE.contains(key)) {
+                try {
+                    rangeFields.put(key, new SearchState.RangePair(value));
+                } catch (ArrayIndexOutOfBoundsException e) {
```
bbpennel, line 278:
```suggestion
                } catch (ArrayIndexOutOfBoundsException | IllegalArgumentException e) {
```
Unless it would be better for the exception to be communicated back to the user, but i think adding the catch here would be inline with the existing behavior of catching ArrayIndexOutOfBoundsException

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
```diff
@@ -397,13 +397,35 @@ private void populateSearchState(SearchState searchState, Map<String,String[]> r
...
             searchState.setFacetsToRetrieve(facetsToRetrieve);
+
+
+            //Add date added year range, which is a sort of facet
+            try {
+                SearchState.RangePair dateCreatedYear = new SearchState.RangePair();
+                parameter = getParameter(request, searchSettings.searchFieldParam(
+                        SearchFieldKey.DATE_CREATED_YEAR.name()));
+                if (parameter != null && parameter.length() > 0) {
+                    String[] dateRange = parameter.split(",");
```
bbpennel, line 408: I'd suggest using the `RangePair(String pairString)` constructor here since it already knows how to parse the range field syntax, and adding a call inside of the constructor to `checkValidRangePair` to make sure its valid.

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
```diff
@@ -397,13 +397,35 @@ private void populateSearchState(SearchState searchState, Map<String,String[]> r
...
+                    dateCreatedYear.setLeftHand(dateRange[0]);
+                    dateCreatedYear.setRightHand(dateRange[1]);
+                }
+
+                if (dateCreatedYear.getLeftHand() != null && dateCreatedYear.getRightHand() != null) {
```
bbpennel, line 413: I think the RangePair validation method would catch if both were null, and it should be fine if one is null, so can probably remove this check

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
```diff
@@ -397,13 +397,35 @@ private void populateSearchState(SearchState searchState, Map<String,String[]> r
...
+                    searchState.getRangeFields().put(SearchFieldKey.DATE_CREATED_YEAR.name(), dateCreatedYear);
+                }
+            } catch (IllegalArgumentException e) {
+                // An invalid range was specified, throw away the range pair
+                Map<String, SearchState.RangePair> currentRangeFields = searchState.getRangeFields();
```
bbpennel, line 418: Did it ever occur that the RangePair got left behind after an exception was thrown? It seems like it shouldn't, since it doesn't get added to searchState.rangeFields until after validation would have occurred. So might be able to remove this part (it's a good idea to keep a log statement about the error having occurred though)

static/js/vue-cdr-access/tests/unit/facets.spec.js
```diff
@@ -276,4 +342,35 @@ describe('facets.vue', () => {
...
+        expect(wrapper.vm.dates.invalid_date_range).toEqual(false);
+        expect(wrapper.find('.date_error').exists()).toBe(false);
+
+        let date_boxes = wrapper.findAll('form input[type=number');
+        date_boxes[0].setValue(2021);
+        date_boxes[1].setValue(2019);
+        await wrapper.find('form input[type=submit]').trigger('click');
+
+        expect(wrapper.vm.dates.invalid_date_range).toEqual(true);
+        expect(wrapper.vm.selected_facets).not.toContain('createdYear=2021,2019');
```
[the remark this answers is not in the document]
bbpennel: In that case, could we check that selected_facets doesn't contain anything at all? just to make sure it hasn't added some other variation on 'createdYear='

bbpennel (changes requested):

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
lfarrell: Yes, it seems to be there. This threw me off for a good while, as I was thinking the same thing. So, I ended up explicitly removing it.

search-api/src/main/java/edu/unc/lib/boxc/search/api/requests/SearchState.java
lfarrell: In this case shouldn't it avoid trying to parse null, since it only runs the check if both sides have values?

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
```diff
@@ -405,14 +405,9 @@ private void populateSearchState(SearchState searchState, Map<String,String[]> r
...
-                    String[] dateRange = parameter.split(",");
-                    dateCreatedYear.setLeftHand(dateRange[0]);
-                    dateCreatedYear.setRightHand(dateRange[1]);
-                }
-
-                if (dateCreatedYear.getLeftHand() != null && dateCreatedYear.getRightHand() != null) {
-                    searchState.getRangeFields().put(SearchFieldKey.DATE_CREATED_YEAR.name(), dateCreatedYear);
+                    new SearchState.RangePair(parameter);
                 }
+                searchState.getRangeFields().put(SearchFieldKey.DATE_CREATED_YEAR.name(), dateCreatedYear);
```
bbpennel, line 410: Should move this line up inside of the `if` starting at line 407 so that we aren't adding DATE_CREATED_YEAR when not actually requested.

search-api/src/main/java/edu/unc/lib/boxc/search/api/requests/SearchState.java
```diff
@@ -327,13 +327,35 @@ public String toString() {
...
-            // As long as one side is set we're good to go.
-            if (start != null && end != null) {
-                if (Integer.parseInt(start) >= Integer.parseInt(end)) {
-                    throw new IllegalArgumentException("Start value, '" + start + "', is greater than end value, " +
-                            "'" + end + "'");
-                }
+            boolean validStart = validRangeValue(start);
+            boolean validEnd = validRangeValue(end);
+
+            if ((validStart && isEmptyValue(end)) || (isEmptyValue(start) && validEnd)) {
```
bbpennel, line 347: Might be good to add a comment here describing what this case is for

search-api/src/main/java/edu/unc/lib/boxc/search/api/requests/SearchState.java
```diff
@@ -327,13 +327,35 @@ public String toString() {
...
+                return;
+            }
+
+            if (!validStart && !validEnd) {
+                throw new IllegalArgumentException("Invalid search range. Start Value: '" + start + "', End Value: " +
+                        "'" + end + "'");
+            }
+
+            if (Integer.parseInt(start) >= Integer.parseInt(end)) {
```
bbpennel, line 356: I think its fine for the start and end to be equal, its how you would specify a single year as the range

bbpennel (changes requested):

bbpennel:
Reading through the behavior in https://github.com/UNC-Libraries/box-c/blob/main/search-solr%2Fsrc%2Fmain%2Fjava%2Fedu%2Func%2Flib%2Fboxc%2Fsearch%2Fsolr%2Futils%2FDateFormatUtil.java#L100 which is used for constructing the search query, we are going to need to make we are selecting the correct range of dates in some live queries, because the way I'm reading it it might be adjusting the end of the range by a year because of not having worked with solr range fields before.

search-solr/src/test/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactoryTest.java
```diff
@@ -222,4 +247,24 @@ public void extractRangePairsNonNumeric() {
...
+
+        SearchState searchState = searchStateFactory.createSearchState(parameters);
+        assertFalse(searchState.getRangeFields().containsKey(SearchFieldKey.DATE_CREATED_YEAR.name()));
+    }
+
+    @Test
+    public void extractRangePairsNull() {
+        Map<String, String[]> parameters = new LinkedHashMap<>();
+        parameters.put("anywhere", new String[]{""});
+        parameters.put("createdYear", new String[]{"null"});
```
bbpennel, line 265: I was thinking more like `parameters.put("createdYear", null);`

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SolrSearchService.java
```diff
@@ -562,6 +562,9 @@ private String getRangeValue(String key, String value) {
         if (StringUtils.isBlank(value)) {
             return "*";
         }
+        if (SearchSettings.FIELDS_RANGE_SEARCHABLE.contains(key)) {
```
bbpennel, line 565: this set also contains `SearchFieldKey.DATE_CREATED.name(), SearchFieldKey.DATE_ADDED.name(), SearchFieldKey.DATE_UPDATED.name()`, have you tested to see if those fields are still searchable in the advanced search? It seems like they would need to go to the FIELDS_DATE_SEARCHABLE block. Both FIELDS_RANGE_SEARCHABLE and FIELDS_DATE_SEARCHABLE contain the same values at the moment

PR #1360, opened 2022-03-03
Set up internationalization of vue components

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3482


bbpennel (commented):
This looks pretty good to me. We'll need to do another pass of pulling text out into the translation file, but you did more of that then I was expecting in this PR, so good work. I would like for us to test this out on dcr-test but otherwise looks good

PR #1363, opened 2022-03-11
Update how creators/contributors are indexed/faceted

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3342

Depends on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/181


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -151,43 +153,49 @@ private void extractTitles(Element mods, IndexDocumentBean idb) {
...
-                                isCreator = true;
-                                break;
-                            }
-                            if (isCreator) {
-                                break;
+                        if (roleTerms.isEmpty()) {
+                            isContributor = true;
+                        }  else {
+                            for (Element roleTerm : roleTerms) {
+                                String roleType = roleTerm.getValue();
```
bbpennel, line 180:
```suggestion
                                String roleType = roleTerm.getTextTrim();
```
Sounds like `getValue()` does a lot more stuff than we need, plus this will trim the value.

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -151,43 +153,49 @@ private void extractTitles(Element mods, IndexDocumentBean idb) {
...
         List<Element> names = mods.getChildren("name", JDOMNamespaceUtil.MODS_V3_NS);
         List<String> creators = new ArrayList<>();
         List<String> contributors = new ArrayList<>();
+        List<String> creatorsContributors = new ArrayList<>();
 
         for (Element nameEl : names) {
             // First see if there is a display form
             String nameValue = formatName(nameEl);
 
             if (nameValue != null) {
```
bbpennel, line 166: since things are getting pretty busy in this block at this point, I'd suggest doing an early skip here rather than nesting
```suggestion
            if (nameValue == null) {
                continue;
            }
```

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -151,43 +153,49 @@ private void extractTitles(Element mods, IndexDocumentBean idb) {
...
-                contributors.add(nameValue);
-
                 List<Element> roles = nameEl.getChildren("role", JDOMNamespaceUtil.MODS_V3_NS);
-                // Person is automatically a creator if no role is provided.
-                boolean isCreator = roles.size() == 0;
-                if (!isCreator) {
+                // Person is automatically a contributor if no role is provided.
+                boolean isContributor = roles.size() == 0;
```
bbpennel, line 169:
```suggestion
                boolean isContributor = roles.isEmpty();
```

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -151,43 +153,49 @@ private void extractTitles(Element mods, IndexDocumentBean idb) {
         }
     }
 
-    private void extractNamesAndAffiliations(Element mods, IndexDocumentBean idb, boolean splitDepartments) {
+    private void extractNames(Element mods, IndexDocumentBean idb) {
         List<Element> names = mods.getChildren("name", JDOMNamespaceUtil.MODS_V3_NS);
         List<String> creators = new ArrayList<>();
         List<String> contributors = new ArrayList<>();
+        List<String> creatorsContributors = new ArrayList<>();
```
bbpennel, line 160: Do we actually need to track this? The creatorsContributors field seems to be getting populated via copyFields.

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
```diff
@@ -446,6 +446,11 @@ private void populateSearchStateAdvancedSearch(SearchState searchState, Map<Stri
             searchState.getSearchFields().put(SearchFieldKey.CONTRIBUTOR_INDEX.name(), parameter);
         }
 
+        parameter = getParameter(request, searchSettings.searchFieldParam(SearchFieldKey.CONTRIBUTOR_INDEX.name()));
```
bbpennel, line 449: This appears to be a duplicate of the block right before it. Was this supposed to be `SearchFieldKey.CREATOR_CONTRIBUTOR`?

static/js/vue-cdr-access/src/components/filterTags.vue
```diff
@@ -106,7 +108,8 @@ export default {
             };
 
             // Return non multi-value tags
-            if (type !== 'format' && type !== 'collection') {
+            const MULTI_VALUED_FIELDS = ['format', 'collection', 'creatorContributor'];
```
bbpennel, line 111: Would probably make sense to declare this as a module level constant rather than in this function, maybe up next to `TYPE`

web-access-app/src/main/resources/search.properties
```diff
@@ -1,11 +1,11 @@
 search.results.maxPerPage=2000
 #Fields which are searchable via keyword searches
-search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX,IDENTIFIER\
+search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CREATOR_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX,IDENTIFIER\
```
bbpennel, line 3: don't think CREATOR_INDEX exists
```suggestion
search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX,IDENTIFIER\
```

web-admin-app/src/main/resources/search.properties
```diff
@@ -1,10 +1,11 @@
 search.results.defaultPerPage=500
 search.results.maxPerPage=5000
 #Fields which are searchable via keyword searches
-search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX
+search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CREATOR_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX
```
bbpennel, line 4:
```suggestion
search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX
```

web-services-app/src/main/resources/search.properties
```diff
@@ -4,11 +4,12 @@ search.results.pagesToDisplay=10
...
-search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX,IDENTIFIER\
+search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CREATOR_INDEX,CONTRIBUTOR_INDEX,SUBJECT_INDEX,IDENTIFIER\
 ,RLA_SITE_CODE,RLA_CATALOG_NUMBER,RLA_CONTEXT_1
 #Set to one higher than you want to display as the last value indicates whether the UI should show a "show more" link
 search.facet.facetsPerGroup=6
 search.facet.expandedFacetsPerGroup=20
 search.facet.maxFacetsPerGroup=50
-search.facet.fields=ANCESTOR_PATH,CONTENT_TYPE,DEPARTMENT,DEPARTMENT_LC,SUBJECT,GENRE,LANGUAGE,LOCATION,PARENT_COLLECTION,CONTENT_MODEL,RESOURCE_TYPE,STATUS,CONTENT_STATUS,SCOPE,ROLE_GROUP
+search.facet.fields=ANCESTOR_PATH,CONTENT_TYPE,DEPARTMENT,DEPARTMENT_LC,SUBJECT,GENRE,LANGUAGE,LOCATION,\
+  PARENT_COLLECTION,CONTENT_MODEL,RESOURCE_TYPE,STATUS,CONTENT_STATUS,SCOPE,ROLE_GROUP,CREATOR,CONTRIBUTOR
```
bbpennel, line 14: I'd suggest also adding `CREATOR_CONTRIBUTOR` here, so the api can work with creator, contributor, and both of them combined (since the other apps can do that last one)

bbpennel (changes requested):

web-access-app/src/main/resources/search.properties
lfarrell: Whoops, it did exist at one point in my pr.

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactory.java
lfarrell: Actually, I think it's not needed at all in this case.

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -441,30 +456,36 @@ private String formatName(Element nameEl) {
...
-                    Element familyPart = JDOMQueryUtil.getElementByAttribute(nameParts, "type", "family");
-                    StringBuilder nameBuilder = new StringBuilder();
-                    if (familyPart != null) {
-                        nameBuilder.append(familyPart.getValue());
-                        if (givenPart != null) {
-                            nameBuilder.append(',').append(' ');
-                        }
-                    }
+                Element givenPart = JDOMQueryUtil.getElementByAttribute(nameParts, "type", null);
+                if (givenPart == null) {
```
bbpennel, line 460: I'd suggest changing these checks to using `StringUtils.isBlank` as well, I was able to get facet entries like ", Given" by doing the following:
```
  <mods:name>
    <mods:namePart type="given">Given</mods:namePart>
    <mods:role>
      <mods:roleTerm>Creator</mods:roleTerm>
    </mods:role>
    <mods:namePart type="family" />
  </mods:name>
```

bbpennel (changes requested):

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -151,44 +154,51 @@ private void extractTitles(Element mods, IndexDocumentBean idb) {
...
         List<Element> names = mods.getChildren("name", JDOMNamespaceUtil.MODS_V3_NS);
         List<String> creators = new ArrayList<>();
         List<String> contributors = new ArrayList<>();
+        List<String> creatorsContributors = new ArrayList<>();
 
         for (Element nameEl : names) {
             // First see if there is a display form
             String nameValue = formatName(nameEl);
 
-            if (nameValue != null) {
```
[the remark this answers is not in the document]
lfarrell: I could never get an empty value to show up as a facet value.

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -496,4 +496,8 @@ private String extractDateYear(Date date) {
         calendar.setTime(date);
         return Integer.toString(calendar.get(Calendar.YEAR));
     }
+
+    private boolean hasNodeValue(Element node) {
```
bbpennel, line 500: Another approach could be to have a private `retrieveNamePartText(Element nameParts, String fieldName)` method to replace all the `JDOMQueryUtil.getElementByAttribute` calls, which would retrieve the value when the element is present, and then you could switch all your conditionals back to just doing `if (datePart != null)`

Something like
```
Element namePart = JDOMQueryUtil.getElementByAttribute(nameParts, "type", fieldName);
if (namePart == null) {
    return null;
} else {
    String value = namePart.getTextTrim();
    return StringUtils.isBlank(value) ? null : value;
}
```

If you don't want to go with that approach, you should switch to using getTextTrim instead of getValue in this method and everywhere in `formatName`, and line 460 isn't using this helper method.

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
```diff
@@ -218,6 +219,9 @@ public void testNamePartConcatenation() throws Exception {
         assertTrue(idb.getCreator().contains("Repo2, Boxy"));
         assertTrue(idb.getCreator().contains("Gilmer, Jeremy Francis, 1918-2020"));
         assertTrue(idb.getCreator().contains("Boxy, Berry Jean, Jr., 1991-"));
+        assertTrue(idb.getCreator().contains("Given"));
```
bbpennel, line 222: Instead of checking for negative buggy cases, you could instead check the size of idb.getCreator() to make sure there are only exactly as many creator values as are expected

indexing-solr/src/test/resources/datastream/nameParts.xml
```diff
@@ -17,6 +17,20 @@
             <mods:roleTerm authority="marcrelator" type="text">Author</mods:roleTerm>
         </mods:role>
     </mods:name>
+    <mods:name type="personal">
```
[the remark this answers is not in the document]
lfarrell: Nothing that I'm aware of. I think I just copied one of the existing  mods:name elements and adjusted its values

bbpennel (approved):

PR #1365, opened 2022-03-16
Add publisher indexing/faceting

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3346
Depends on https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/182


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -315,6 +316,26 @@ private void extractLanguages(Element mods, IndexDocumentBean idb) {
...
 
+    private void extractPublisher(Element mods, IndexDocumentBean idb) {
+        List<Element> originInfoEls = mods.getChildren("originInfo", JDOMNamespaceUtil.MODS_V3_NS);
+        List<String> publishers = new ArrayList<>();
+        if (!originInfoEls.isEmpty()) {
+            for (Element originInfoEl : originInfoEls) {
+                List<Element> publisherEls = originInfoEl.getChildren("publisher", JDOMNamespaceUtil.MODS_V3_NS);
+                if (!publisherEls.isEmpty()) {
+                    for (Element publisher : publisherEls) {
+                        publishers.add(publisher.getValue());
```
bbpennel, line 327:
```suggestion
                        publishers.add(publisher.getTextTrim());
```

bbpennel (changes requested):
I think we'll want to add publisher as a field that gets serialized in `SerializationUtil`.

Might be worth checking with anna if she thinks it should get returned by the API in the "full" field set https://github.com/UNC-Libraries/box-c/blob/main/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/SearchRestController.java#L67

bbpennel (approved):
Looks good! I think the vue code will be impacted by the creator/contributor PR. Which would you like to merge first?

lfarrell:
Merging this one first is fine.

PR #1367, opened 2022-03-17
Remove unused solr fields

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3327
Associated PR: https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/183


model-api/src/main/java/edu/unc/lib/boxc/model/api/ResourceType.java
```diff
@@ -28,29 +28,23 @@
  *
  */
 public enum ResourceType {
-    AdminUnit(1, Cdr.AdminUnit),
```
bbpennel, line 31: These sort orders are still needed, they are used for sorting by object type and are associated with the RESOURCE_TYPE_SORT solr field

web-admin-app/src/main/java/edu/unc/lib/boxc/web/admin/controllers/AbstractSearchController.java
```diff
@@ -41,12 +41,11 @@
     protected static List<String> resultsFieldList = Arrays.asList(SearchFieldKey.ID.name(),
             SearchFieldKey.TITLE.name(), SearchFieldKey.CREATOR.name(), SearchFieldKey.DATASTREAM.name(),
             SearchFieldKey.DATE_ADDED.name(), SearchFieldKey.DATE_UPDATED.name(),
-            SearchFieldKey.RESOURCE_TYPE.name(), SearchFieldKey.CONTENT_MODEL.name(),
+            SearchFieldKey.RESOURCE_TYPE.name(),
             SearchFieldKey.STATUS.name(), SearchFieldKey.VERSION.name(),SearchFieldKey.ROLE_GROUP.name(),
-            SearchFieldKey.RELATIONS.name(), SearchFieldKey.CONTENT_TYPE.name(),
+            SearchFieldKey.CONTENT_TYPE.name(),
             SearchFieldKey.CONTENT_STATUS.name(), SearchFieldKey.LABEL.name(), SearchFieldKey.TIMESTAMP.name(),
```
bbpennel, line 47: remove label

web-admin-app/src/main/resources/search.properties
```diff
@@ -6,6 +6,6 @@ search.field.searchable=ID,DEFAULT_INDEX,TITLE_INDEX,CONTRIBUTOR_INDEX,SUBJECT_I
 search.facet.facetsPerGroup=15
 search.facet.expandedFacetsPerGroup=50
 search.facet.maxFacetsPerGroup=50
-search.facet.fields=ANCESTOR_PATH,CONTENT_TYPE,DEPARTMENT,DEPARTMENT_LC,SUBJECT,GENRE,LANGUAGE,LOCATION,PUBLISHER,\
-  PARENT_COLLECTION,CONTENT_MODEL,RESOURCE_TYPE,STATUS,CONTENT_STATUS,SCOPE,ROLE_GROUP,CREATOR_CONTRIBUTOR
+search.facet.fields=ANCESTOR_PATH,CONTENT_TYPE,SUBJECT,GENRE,LANGUAGE,LOCATION,PUBLISHER,\
+  PARENT_COLLECTION,RESOURCE_TYPE,STATUS,CONTENT_STATUS,SCOPE,ROLE_GROUP,CREATOR_CONTRIBUTOR
```
bbpennel, line 10: remove scope

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/ExportCsvService.java
```diff
@@ -73,8 +73,6 @@
     public static final String OBJ_TYPE_HEADER = "Object Type";
     public static final String PID_HEADER = "PID";
     public static final String TITLE_HEADER = "Title";
-    public static final String PATH_HEADER = "Path";
```
bbpennel, line 76: Already mentioned elsewhere, but we should keep path here

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ExportCsvIT.java
```diff
@@ -97,10 +97,6 @@
         @ContextConfiguration("/export-csv-it-servlet.xml")
 })
 public class ExportCsvIT extends AbstractAPIIT {
-    private static final String COLLECTION_PATH =
```
bbpennel, line 100: Might end up needing to revert this file

bbpennel (changes requested):
Looks good, mostly commented on a few extra things to remove, and one thing we want to add back (the ancestor names in the csv, but not the solr field)

Can also remove:
`deposit-app/src/test/resources/dept-ontology.xml`
`operations/src/test/resources/samples/dept-ontology.xml`
 and probably `departments.skos.url` from the deposit apps properties file.

At some point we're going to need to tidy up our csv since I see references in it to some of these fields, but probably out of scope for this PR

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetObjectTypeFilterTest.java
```diff
@@ -88,7 +88,6 @@ public void testMultipleRdfTypes() throws Exception {
         filter.filter(dip);
 
         verify(idb).setResourceType(eq(ResourceType.AdminUnit.name()));
-        verify(idb).setResourceTypeSort(eq(ResourceType.AdminUnit.getDisplayOrder()));
```
bbpennel, line 91: Can restore this line too

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/models/ContentObjectSolrRecord.java
```diff
@@ -192,27 +192,6 @@ public void setRoleGroup(List<String> roleGroup) {
...
-            if (values == null) {
-                values = new ArrayList<>();
-                this.relationsMap.put(rdfParts[0], values);
-            }
-            values.add(rdfParts[1]);
-        }
-    }
-
     @Override
     public List<String> getRelation(String relationName) {
```
bbpennel, line 196: I think this method can be removed, as well a `relationsMap`

bbpennel (changes requested):

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/models/ContentObjectSolrRecord.java
lfarrell: It looks like getRelation is used here to figure out if there's an embargo https://github.com/UNC-Libraries/box-c/blob/removed-fields/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/models/ContentObjectSolrRecord.java#L260

bbpennel: Hmm, good point. Nothing is populating the relationsMap anymore, so `getActiveEmbargo` is probably not functional. The only place it is used is here: https://github.com/UNC-Libraries/box-c/blob/main/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java#L209
Which is probably dead code at the moment. The service which expires embargoes does not use the solr field, it appears to rely on the triple store:
https://github.com/UNC-Libraries/box-c/blob/main/operations/src/main/java/edu/unc/lib/boxc/operations/impl/acl/ExpireEmbargoService.java#L131

Given that the code can't actually be working at the moment, my inclination is to delete getActiveEmbargo and its used in the FullRecordController. We can make a followup ticket to add the functionality back in using the correct clients later

PR #1372, opened 2022-03-24
BXC-3514 - Full record page faceting

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3514


static/js/vue-cdr-access/src/components/displayWrapper.vue
```diff
@@ -96,17 +106,23 @@ Top level component for full record pages with searching/browsing, including Adm
                 return this.showWidget || this.coerceWorksOnly(this.$route.query.works_only);
             },
 
-
+            minimumCreatedYear() {
```
lfarrell, line 109: I wonder if we want to move this to mixin that can be used across components?

static/js/vue-cdr-access/tests/unit/browseSearch.spec.js
```diff
@@ -1,4 +1,4 @@
-import { shallowMount, flushPromises } from '@vue/test-utils';
+import { shallowMount, flushPromises, mount } from '@vue/test-utils';
```
lfarrell, line 1: Should be able to get rid of shallowMount

static/js/vue-cdr-access/src/App.vue
```diff
@@ -3,3 +3,28 @@
     <router-view/>
   </div>
 </template>
+<style lang="scss">
```
lfarrell, line 6: If styles are used across facets. I've usually pulled them out into their own stylesheet and then import that into main.js. See https://github.com/UNC-Libraries/box-c/tree/main/static/js/admin/vue-permissions-editor/src/assets

static/js/vue-cdr-access/src/components/displayWrapper.vue
bbpennel: apparently I forgot to even set `min_created_year` to a value, I guess that's one disadvantage to shallowMounting the subcomponents (I tried to regular mount for the displayWrapper tests but it gave me some really weird errors and I backed off after a while). But yeah I can move it over.

static/js/vue-cdr-access/src/App.vue
lfarrell: I'd think adding stylesheet is fine. I find removing the scoped flag on a components CSS get a little confusing, though I did it for the slider component, otherwise the style overrides wouldn't work.

lfarrell (approved):

PR #1376, opened 2022-03-28
Add otherSubjects field

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3534

### DIFF
--- a/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
+++ b/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
@@ -84,6 +84,7 @@ public class SetDescriptiveMetadataFilter implements IndexDocumentFilter {
             this.extractCollectionId(mods, idb);
             this.extractLanguages(mods, idb);
             this.extractSubjects(mods, idb);
+            this.extractOtherSubjects(mods, idb);
             this.extractLocations(mods, idb);
             this.extractPublisher(mods, idb);
             this.extractDateCreated(mods, idb);
@@ -280,6 +281,31 @@ public class SetDescriptiveMetadataFilter implements IndexDocumentFilter {
             idb.setSubject(null);
         }
 
+    }
+
+    private void extractOtherSubjects(Element mods, IndexDocumentBean idb) {
+        List<Element> otherSubjectEls = mods.getChildren("subject", JDOMNamespaceUtil.MODS_V3_NS);
+        List<String> otherSubjects = new ArrayList<>();
+        if (!otherSubjectEls.isEmpty()) {
+            for (Element otherSubjectObj : otherSubjectEls) {
+                List<Element> subjectParts = otherSubjectObj.getChildren();
+                for (Element otherSubject : subjectParts) {
+                    String otherSubjectName = otherSubject.getName();
+                    if (otherSubjectName.equals("name") || otherSubjectName.equals("topic")) {
+                        continue;
+                    }
+
+                    if (otherSubject.getChildren().isEmpty()) {
+                        addIfNotBlank(otherSubjects, otherSubject.getValue());
+                    }
+                }
+            }
+        }
+        if (otherSubjects.size() > 0) {
+            idb.setOtherSubject(otherSubjects);
+        } else {
+            idb.setOtherSubject(otherSubjects);
+        }
     }
 
     private void extractLocations(Element mods, IndexDocumentBean idb) {
--- a/indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
+++ b/indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
@@ -157,6 +157,11 @@ public class SetDescriptiveMetadataFilterTest {
 
         assertEquals("2006-04", dateFormat.format(idb.getDateCreated()));
 
+        assertTrue(idb.getOtherSubject().contains("Canada"));
+        assertTrue(idb.getOtherSubject().contains("Explorer"));
+        assertFalse(idb.getOtherSubject().contains("Snowshoes"));
+        assertFalse(idb.getOtherSubject().contains("rules, boxy"));
+
         List<String> ids = idb.getIdentifier();
         assertTrue(ids.contains("local|abc123"));
         assertFalse(ids.contains("uri|http://example.com"));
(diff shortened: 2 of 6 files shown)


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -282,6 +283,31 @@ private void extractSubjects(Element mods, IndexDocumentBean idb) {
...
+                        continue;
+                    }
+
+                    if (otherSubject.getChildren().isEmpty()) {
+                        addIfNotBlank(otherSubjects, otherSubject.getValue());
+                    }
+                }
+            }
+        }
+        if (otherSubjects.size() > 0) {
```
bbpennel, line 304: Both sides of this conditional do the same thing, so probably don't need it

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -282,6 +283,31 @@ private void extractSubjects(Element mods, IndexDocumentBean idb) {
...
+        if (!otherSubjectEls.isEmpty()) {
+            for (Element otherSubjectObj : otherSubjectEls) {
+                List<Element> subjectParts = otherSubjectObj.getChildren();
+                for (Element otherSubject : subjectParts) {
+                    String otherSubjectName = otherSubject.getName();
+                    if (otherSubjectName.equals("name") || otherSubjectName.equals("topic")) {
+                        continue;
+                    }
+
+                    if (otherSubject.getChildren().isEmpty()) {
```
bbpennel, line 298: Just a thought, but it might make sense to merge this conditional with the one at 194 since they are all conditions if which we want to skip adding the contents of this subject

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
```diff
@@ -157,6 +157,11 @@ public void testInventory() throws Exception {
 
         assertEquals("2006-04", dateFormat.format(idb.getDateCreated()));
 
+        assertTrue(idb.getOtherSubject().contains("Canada"));
+        assertTrue(idb.getOtherSubject().contains("Explorer"));
+        assertFalse(idb.getOtherSubject().contains("Snowshoes"));
+        assertFalse(idb.getOtherSubject().contains("rules, boxy"));
```
bbpennel, line 163: Add a check for the size of the otherSubjects, and then you can remove the two false cases. Checking the size should be more full proof in case we edit any values in the test document and forget to update this.

search-api/src/main/java/edu/unc/lib/boxc/search/api/SearchFieldKey.java
```diff
@@ -57,6 +57,7 @@
     LANGUAGE("language", "language", "Language"),
     LAST_INDEXED("lastIndexed", "lastIndexed", "lastIndexed"),
     LOCATION("location", "location", "Location"),
+    OTHER_SUBJECTS("otherSubject", "otherSubjects", "Other Subjects"),
```
bbpennel, line 60: hmm, we're getting a bit inconsistent about pluralization. I'd suggest
```suggestion
    OTHER_SUBJECT("otherSubject", "otherSubject", "Other Subject"),
```
I missed previously that `OTHER_TITLES` was plural, for consistency could you also update that one not to be?

bbpennel (changes requested):

bbpennel (approved):

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -282,6 +283,30 @@ private void extractSubjects(Element mods, IndexDocumentBean idb) {
...
 
+    private void extractOtherSubjects(Element mods, IndexDocumentBean idb) {
+        List<Element> otherSubjectEls = mods.getChildren("subject", JDOMNamespaceUtil.MODS_V3_NS);
+        List<String> otherSubjects = new ArrayList<>();
+        if (!otherSubjectEls.isEmpty()) {
+            for (Element otherSubjectObj : otherSubjectEls) {
+                List<Element> subjectParts = otherSubjectObj.getChildren();
+                for (Element otherSubject : subjectParts) {
+                    String otherSubjectName = otherSubject.getName();
+                    if (!otherSubject.getChildren().isEmpty() ||
```
bbpennel, line 294: Turns out we will want to get the values from all the children rather than skipping subjects with children, so I guess we'll need a loop when children isn't empty. Anna is getting some examples for the test.

bbpennel (changes requested):

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -285,26 +285,48 @@ private void extractSubjects(Element mods, IndexDocumentBean idb) {
 
     private void extractOtherSubjects(Element mods, IndexDocumentBean idb) {
         List<Element> otherSubjectEls = mods.getChildren("subject", JDOMNamespaceUtil.MODS_V3_NS);
-        List<String> otherSubjects = new ArrayList<>();
+        List<String> otherSubjects = extractNestedSubjects(otherSubjectEls, new ArrayList<>());
```
bbpennel, line 288: This method doesn't really need a return value since you are modifying otherSubjects in place. So I'd suggest initializing otherSubjects to an empty array on the previous line and then calling the method on the next line, and removing the return type from the method.

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
```diff
@@ -161,7 +161,8 @@ public void testInventory() throws Exception {
         assertTrue(idb.getOtherSubject().contains("Germany"));
         assertTrue(idb.getOtherSubject().contains("Canada"));
         assertTrue(idb.getOtherSubject().contains("Explorer"));
-        assertEquals(3, idb.getOtherSubject().size());
+        assertTrue(idb.getOtherSubject().contains("scale"));
+        assertEquals(5, idb.getOtherSubject().size());
```
bbpennel, line 165: how did the count jump from 3 to 5 when only one new subject was added?

bbpennel (changes requested):

bbpennel (approved):

PR #1378, opened 2022-03-28
Add rights statement and rights statement uris

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3532
Relies on: https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/187


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -393,6 +394,35 @@ private void extractDateCreated(Element mods, IndexDocumentBean idb) {
...
+                String href = rightsEl.getAttributeValue("href", JDOMNamespaceUtil.XLINK_NS);
+                if (!StringUtils.isBlank(href)) {
+                    rightsUri.add(href);
+                } else {
+                    rights.add(rightsEl.getTextTrim());
+                }
+            }
+        }
+
+        if (rights.size() > 0) {
```
bbpennel, line 413: use `.isEmpty`

search-api/src/main/java/edu/unc/lib/boxc/search/api/SearchFieldKey.java
```diff
@@ -64,6 +64,8 @@
     READ_GROUP("readGroup", "readGroup", "Read Group"),
     RESOURCE_TYPE("resourceType", "type", "Resource Type"),
     RESOURCE_TYPE_SORT("resourceTypeSort", "typeSort", "Resource Type Sort"),
+    RIGHT("right", "right", "Rights"),
```
bbpennel, line 67: I don't think this currently matches the solr schema field definition, and I know its weird after our the last PR where I was saying to depluralize, but here and in the URI field I think it would be appropriate to use "rights"
```suggestion
    RIGHTS("rights", "rights", "Rights"),
```

bbpennel (changes requested):
I asked Anna a few questions on the ticket, but in the meantime there are two other things we should look into adding:

- Update the solr schema.xml in boxc to match the ansible changes
- Add in the controlled vocabulary which Anna provided in the ticket. At the moment it looks like it can just be a text doc that gets loaded into a Set during initialization, and then we check against that. Might want a warning logged if a term doesn't match the vocab

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -445,6 +468,87 @@ private void extractDateCreated(Element mods, IndexDocumentBean idb) {
...
+
+                if (hasUrl && !hasRightsText) {
+                    String trimmedHref = href.trim();
+                    if (rightsUriMap.containsKey(trimmedHref)) {
+                        String mappedRightsText = rightsUriMap.get(trimmedHref);
+                        rightsOaiPmh.add(setUrlProtocol(trimmedHref));
+                        rightsOaiPmh.add(mappedRightsText);
+                        rights.add(mappedRightsText);
+                    }
+                    rightsUri.add(trimmedHref);
```
bbpennel, line 502: We only want to add to rightsUri when the value is in the controlled vocab

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -445,6 +468,87 @@ private void extractDateCreated(Element mods, IndexDocumentBean idb) {
...
+                }
+
+                if (hasRightsText && hasUrl) {
```
bbpennel, line 505: I think we can merge the logic for `hasUrl && !hasRightsText` with this, might even be able to use this logic for all 3 cases:

1. if there is a url, check if its in vocab:
    1. if it is:
        1. add it to rightsUri
        2. set indexingText to the display label from vocab
        3. add the transformed version to rightsOaiPmh
    2. if it was not in the vocab, log a warning and discard the URI
2. if indexingText == null
    1. then, set it to the text value from the MODS field
    2. else, compare it to the value from the MODS field, and log error if they don't match
3. if indexingText != null
    1. add indexingText to `rights` and `rightsOaiPmh`

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -445,6 +468,87 @@ private void extractDateCreated(Element mods, IndexDocumentBean idb) {
...
+        }
+
+        if (!rightsUri.isEmpty()) {
+            idb.setRightsUri(rightsUri);
+        } else {
+            idb.setRightsUri(null);
+        }
+    }
+
+    private String setUrlProtocol(String uri) {
```
bbpennel, line 545: since I usually try to avoid `set` for non-setter methods, something like `convertUrlProtocol` might be better

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
```diff
@@ -159,6 +160,27 @@ public void testInventory() throws Exception {
...
+        assertTrue(idb.getRights().contains("Copyright Not Evaluated"));
+        assertTrue(idb.getRights().contains("For copyright information or permissions questions, see our " +
+                "intellectual property statement https://library.unc.edu/wilson/research/perm/"));
+        assertTrue(idb.getRights().contains("Copyright Not Evaluated"));
+        assertTrue(idb.getRights().contains("Random Rights"));
+
+        assertEquals(4, idb.getRightsUri().size());
+        assertTrue(idb.getRightsUri().contains("https://rightsstatements.org/vocab/CNE/1.0/"));
+        assertTrue(idb.getRightsUri().contains("https://creativecommons.org/licenses/by-sa/3.0/us/"));
+        assertTrue(idb.getRightsUri().contains("https://random.url/rights/no_text/"));
```
bbpennel, line 173: Should not include values from outside the vocab

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
```diff
@@ -159,6 +160,27 @@ public void testInventory() throws Exception {
...
+        assertTrue(idb.getRightsUri().contains("https://random.url/rights/text/"));
+
+        assertEquals(6, idb.getRightsOaiPmh().size());
+        assertTrue(idb.getRightsOaiPmh().contains("http://rightsstatements.org/vocab/CNE/1.0/"));
+        assertTrue(idb.getRightsOaiPmh().contains("http://creativecommons.org/licenses/by-sa/3.0/us/"));
+        assertTrue(idb.getRightsOaiPmh().contains("http://random.url/rights/text/"));
```
bbpennel, line 179: Should not include URIs from outside the vocab

bbpennel (changes requested):

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -69,6 +74,23 @@ public SetDescriptiveMetadataFilter() {
         } catch (IOException e) {
             log.error("Failed to load code language mappings", e);
         }
+
+        // URIS aren't great for property keys
+        rightsUriMap = new HashMap<>();
+        try {
+            InputStream rightsStream = getClass().getClassLoader().getResourceAsStream(
```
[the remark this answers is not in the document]
lfarrell: I tried that but I ran into a number of issues with having urls as property keys. The colon is a reserved character.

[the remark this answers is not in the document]
bbpennel: Okay, that's reasonable

bbpennel (approved):

bbpennel (approved):

PR #1379, opened 2022-03-28
LIB-16359 - Resolve issues with refreshing solr after move operations

bbpennel:
https://jira.lib.unc.edu/browse/LIB-16359
* Add status updating filter to move operation indexing, since inherited statuses were not clearing when moving out of a folder with permissions restrictions
* Invalidate path cache when an object is modified, to prevent stale value from preventing the path from updating in solr from a move operation if the same object had recently been indexed

### DIFF
--- a/services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/util/CacheInvalidatingProcessor.java
+++ b/services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/util/CacheInvalidatingProcessor.java
@@ -17,6 +17,7 @@ package edu.unc.lib.boxc.services.camel.util;
 
 import static org.slf4j.LoggerFactory.getLogger;
 
+import edu.unc.lib.boxc.model.api.services.ContentPathFactory;
 import org.apache.camel.Exchange;
 import org.apache.camel.Message;
 import org.apache.camel.Processor;
@@ -39,6 +40,7 @@ public class CacheInvalidatingProcessor implements Processor {
     private static final Logger log = getLogger(CacheInvalidatingProcessor.class);
     private RepositoryObjectLoader repoObjLoader;
     private ObjectAclFactory objectAclFactory;
+    private ContentPathFactory contentPathFactory;
 
     @Override
     public void process(Exchange exchange) throws Exception {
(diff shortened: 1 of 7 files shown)


services-camel-app/src/test/java/edu/unc/lib/boxc/services/camel/util/CacheInvalidatingProcessorTest.java
```diff
@@ -23,6 +23,7 @@
 import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.initMocks;
 
+import edu.unc.lib.boxc.model.api.services.ContentPathFactory;
```
lfarrell, line 26: Should this be with the other edu imports?

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/util/CacheInvalidatingProcessor.java
```diff
@@ -17,6 +17,7 @@
 
 import static org.slf4j.LoggerFactory.getLogger;
 
+import edu.unc.lib.boxc.model.api.services.ContentPathFactory;
```
lfarrell, line 20: Should this be with the other edu imports?

lfarrell (changes requested):
Should you have to explicitly re-index after moving an object in/out of a restricted folder to see the inherited permissions change?

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/util/CacheInvalidatingProcessor.java
bbpennel: I'm not sure why intellij put this here, guessing it didn't agree with eclipse about ordering. I've told it to organize imports

bbpennel:
No, this PR is supposed to fix that particular bug. I've deployed the branch to test and am seeing restricted and inherited flags cleared on the moved object. What steps did you take to produce the issue? Did you deploy the services-camel application?

lfarrell:
I hadn't deployed all the changes

lfarrell (approved):

PR #1381, opened 2022-03-30
Don't index full text for work objects

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3583


bbpennel (approved):
This looks good. Have you tested to make sure that a full text search still turns up work objects after reindexing?

lfarrell:
Yes, it seems to return the work. You can search "Rotisserie" on dcr-test and it should bring up one record, where the full text of the file has that word. However, all the facets appear to be empty.

PR #1389, opened 2022-04-04
Replace string_lc fields with SortableTextField

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3574


bbpennel (approved):
This looks good. I'm not going to merge it quite yet until we are ready to reindex test, but i think we'll do so soon

PR #1390, opened 2022-04-05
Add sorting by date added

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3529


static/js/vue-cdr-access/src/translations.js
```diff
@@ -26,6 +26,8 @@ export default {
             "relevance": "Sort by Relevance",
             "title_a-z": "Title A-Z",
             "title_z-a": "Title Z-A",
+            "date_added_newest": "Date Added (newest)",
```
bbpennel, line 29: Judging by the ticket, I think we want the label to be "Date Deposited"

bbpennel (changes requested):

PR #1391, opened 2022-04-07
BXC-3347 - Add unit facet

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3347

* Search/list json responses return a "filterParameters" field, which lists all the search terms/ranges/facets applied to the search
    * Includes display values for facets, particularly for Collection and Unit facets
    * Additional services to support population of this.
* Updates filterTags to work with filterParameters rather than exclusively off of values from the URL


static/js/vue-cdr-access/src/components/browseSearch.vue
```diff
@@ -19,7 +19,7 @@ Search form component displayed on full record pages, allowing keyword searches
                     <i class="fas fa-times"></i>
                 </span> {{ $t('search.clear_search')}}</a>
             <clear-facets-button></clear-facets-button>
-            <filter-tags :facet-list="facetList"></filter-tags>
+            <filter-tags :filter-parameters="this.filterParameters"></filter-tags>
```
lfarrell, line 22: Should be able to remove `this` here

static/js/vue-cdr-access/src/components/filterTags.vue
```diff
@@ -99,27 +93,36 @@ export default {
...
-                original_value: fieldValue,
-                value_text: display_text
-            };
-
-            // Return non multi-value tags
-            if (!MULTI_VALUED_FIELDS.includes(type)) {
+            if (Array.isArray(fieldValue)) {
+                // Multivalued fields
+                return fieldValue.map((value) => {
+                    let tag_info = {
```
lfarrell, line 99: Could probably move this out of the if statement, since both fields repeat in the else statement. Then in the else clause  just set the original_value `tag_info.original_value = fieldValue`

lfarrell (approved):

PR #1395, opened 2022-04-12
BXC-3563 - Revisions to full record page faceting

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3563
Also addresses https://jira.lib.unc.edu/browse/BXC-3610

* Facet results in listJson requests now include all children
    * Required fixing an issue applyCutoffs parameter, which was not being honored for the ancestorPath facet when constructing solr queries (it is the only CutoffFacet), and was incorrectly being used to prevent other facet behaviors (most likely code had moved around at some point and the conditional not been updated)
* Applying facet restrictions now switches from listJson to searchJson, just like search terms
* "Show works only" button now shown on admin unit pages
* Suppress Admin Unit facet on all full record pages, suppress Collection facet on collection and folder pages
    * Required switching possible_facet_fields over to vuex so a consistent list of facets was used after removing values from the list, which required adding in vuex
* Fix bug that caused the UUID in the url to be removed when clearing the `anywhere` term
* A few minor refactors, like using "start" everywhere, centralizing UUID regex

### DIFF
--- a/static/js/vue-cdr-access/src/store.js
+++ b/static/js/vue-cdr-access/src/store.js
@@ -13,6 +13,18 @@ const store = createStore({
     mutations: {
         removePossibleFacetFields (state, removeFacets) {
             state.possibleFacetFields = state.possibleFacetFields.filter(f => removeFacets.indexOf(f) < 0);
+        },
+        resetPossibleFacetFields (state) {
+            state.possibleFacetFields = POSSIBLE_FACET_FIELDS.slice();
+        }
+    },
+    actions: {
+        /**
+         * Used to reset the store back to its initial state, primarily for testing
+         * @param context
+         */
+        resetState (context) {
+            context.commit("resetPossibleFacetFields");
         }
     }
 });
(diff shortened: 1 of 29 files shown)


static/js/vue-cdr-access/src/store.js
```diff
@@ -0,0 +1,23 @@
...
+    state () {
+        return {
+            possibleFacetFields: POSSIBLE_FACET_FIELDS.slice()
+        }
+    },
+    mutations: {
+        removePossibleFacetFields (state, removeFacets) {
+            state.possibleFacetFields = state.possibleFacetFields.filter(f => removeFacets.indexOf(f) < 0);
+        },
+        resetPossibleFacetFields (state) {
```
lfarrell, line 17: This doesn't seem to be used.

lfarrell (changes requested):
One of the vuex methods seems to not be used anywhere.

lfarrell (approved):

PR #1397, opened 2022-04-20
Reset start row to 0, if search changes

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3592

### DIFF
--- a/static/js/vue-cdr-access/src/components/browseSearch.vue
+++ b/static/js/vue-cdr-access/src/components/browseSearch.vue
@@ -78,16 +78,27 @@ Search form component displayed on full record pages, allowing keyword searches 
 
         methods: {
             getResults() {
-                let update_params = { anywhere: encodeURIComponent(this.search_query), start: 0 };
-                this.routeWithParams(this.urlParams(update_params, 'displayRecords'));
+                const reset_search_start_row = this.resetStartRow({ anywhere: encodeURIComponent(this.search_query) });
+                const update_parameters = this.urlParams(reset_search_start_row, 'displayRecords');
+                this.routeWithParams(update_parameters);
             },
 
             clearSearch() {
-                this.routeWithParams(this.removeQueryParameters(this.allPossibleSearchParameters, true));
+                this.clear(this.allPossibleSearchParameters);
             },
 
             clearAllFacets() {
-                this.routeWithParams(this.removeQueryParameters(this.possibleFacetFields, true));
+                this.clear(this.possibleFacetFields);
+            },
+
+            /**
+             * Clears facets/searches
+             * @param param_to_clear
+             */
+            clear(param_to_clear) {
+                const reset_params = this.removeQueryParameters(param_to_clear);
+                const reset_start_row = this.resetStartRow(reset_params);
+                this.routeWithParams(reset_start_row);
             }
         },
 
--- a/static/js/vue-cdr-access/src/components/browseSearch.vue
+++ b/static/js/vue-cdr-access/src/components/browseSearch.vue
@@ -78,8 +78,8 @@ Search form component displayed on full record pages, allowing keyword searches 
 
         methods: {
             getResults() {
-                const reset_search_start_row = this.resetStartRow({ anywhere: encodeURIComponent(this.search_query) });
-                const update_parameters = this.urlParams(reset_search_start_row, 'displayRecords');
+                const encode_search = { anywhere: encodeURIComponent(this.search_query), start: 0 };
+                const update_parameters = this.urlParams(encode_search, 'displayRecords');
                 this.routeWithParams(update_parameters);
             },
 
--- a/static/js/vue-cdr-access/src/components/browseSearch.vue
+++ b/static/js/vue-cdr-access/src/components/browseSearch.vue
@@ -78,16 +78,16 @@ Search form component displayed on full record pages, allowing keyword searches 
 
         methods: {
             getResults() {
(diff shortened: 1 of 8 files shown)


static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -125,9 +125,13 @@ export default {
         /**
          * Returns a list of query parameters from the current URLs minus the specified set of parameters
          * @param param_names names of query parameters to remove
+         * @param reset_start_row Start row should be reset to 0 for searching, pagination purposes
          * @returns {LocationQuery} list of query parameters with specified parameters removed
          */
-        removeQueryParameters(param_names) {
+        removeQueryParameters(param_names, reset_start_row = false) {
+            if (reset_start_row) {
```
bbpennel, line 132: I think I would prefer to add a new method named "resetStartRow" which we could used like 
```
this.routeWithParams(this.resetStartRow(this.removeQueryParameters(this.allPossibleSearchParameters)))
```
It'd be clearer what is happening and it wouldn't add extra behaviors into this method, which is pretty clearly scoped right now

static/js/vue-cdr-access/src/components/browseSearch.vue
```diff
@@ -78,16 +78,16 @@ Search form component displayed on full record pages, allowing keyword searches
 
         methods: {
             getResults() {
-                let update_params = { anywhere: encodeURIComponent(this.search_query) };
+                let update_params = { anywhere: encodeURIComponent(this.search_query), start: 0 };
                 this.routeWithParams(this.urlParams(update_params, 'displayRecords'));
```
bbpennel, line 82: if you add it, it would be good to use `resetStartRow` whenever possible for resetting the start row in param objects

bbpennel (changes requested):

static/js/vue-cdr-access/src/components/browseSearch.vue
lfarrell: I had that at one point, but took it out.

static/js/vue-cdr-access/src/components/browseSearch.vue
```diff
@@ -78,16 +78,16 @@ Search form component displayed on full record pages, allowing keyword searches
...
             },
 
             clearSearch() {
-                this.routeWithParams(this.removeQueryParameters(this.allPossibleSearchParameters));
+                this.routeWithParams(this.removeQueryParameters(this.allPossibleSearchParameters, true));
             },
 
             clearAllFacets() {
-                this.routeWithParams(this.removeQueryParameters(this.possibleFacetFields));
```
bbpennel, line 90: I don't think this is working, if I clear all facets it stays on the current page for me on dcr-test.

For the sake of readability, I think it might be good to break this line up into 2 or 3 lines with good variable names, or good comments, since its a decent number of steps

bbpennel: There are a two other places that we should reset the page number:
* Clicking "Show only works"
* Changing the sort (this one doesn't break, but is confusing)

bbpennel (changes requested):

static/js/vue-cdr-access/src/mixins/routeUtils.js
```diff
@@ -125,9 +125,13 @@ export default {
         /**
          * Returns a list of query parameters from the current URLs minus the specified set of parameters
          * @param param_names names of query parameters to remove
+         * @param reset_start_row Start row should be reset to 0 for searching, pagination purposes
          * @returns {LocationQuery} list of query parameters with specified parameters removed
          */
-        removeQueryParameters(param_names) {
+        removeQueryParameters(param_names, reset_start_row = false) {
```
bbpennel, line 131: I don't think `reset_start_row` is ever used now, so can remove it and the block that goes with it

static/js/vue-cdr-access/src/components/searchWrapper.vue
```diff
@@ -108,12 +115,20 @@ Top level component wrapper for search pages
...
                 }).catch(function (error) {
                     console.log(error);
+                    this.query_changed = false;
                 });
             }
         },
 
+        // check if query params other than "start" have changed otherwise pagination can't get past the first page
+        beforeRouteUpdate(to, from) {
+            this.query_changed = !isEqual(this.removeQueryParameters(['start'], to.query),
```
bbpennel, line 128: I'm confused how this is working:
`this.removeQueryParameters(['start'], to.query)`
The function seems to accept an array and a boolean, i don't see a version of it that would remove `['start']` from the provided `query` object, am I missing something?

static/js/vue-cdr-access/src/components/browseSearch.vue
```diff
@@ -78,16 +78,27 @@ Search form component displayed on full record pages, allowing keyword searches
...
+                this.clear(this.possibleFacetFields);
+            },
+
+            /**
+             * Clears facets/searches
+             * @param param_to_clear
+             */
+            clear(param_to_clear) {
```
bbpennel, line 98: maybe `clearParameters` to differentiate it from other possible clearing operations (like clearing the search box or something)

bbpennel (changes requested):

static/js/vue-cdr-access/src/mixins/routeUtils.js
lfarrell: Whoops, thought I'd removed this.

bbpennel (approved):

PR #1400, opened 2022-04-25
Fix wrapping of filter tags

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3648


static/js/vue-cdr-access/src/components/clearFilters.vue
```diff
@@ -0,0 +1,67 @@
...
+
+<style scoped  lang="scss">
+    i {
+        padding-left: 5px;
+    }
+    .clear-options {
+        display: inline-flex;
+        flex-wrap: wrap;
+        margin-left: 35px;
+    }
```
bbpennel, line 66: I would suggest adding `margin-top: -16px` here so that the gap between the search input and the clear buttons is more in line with the other vertical gaps in that area

static/js/vue-cdr-access/src/components/clearFilters.vue
```diff
@@ -0,0 +1,67 @@
...
+            const reset_start_row = this.resetStartRow(reset_params);
+            this.routeWithParams(reset_start_row);
+        }
+    }
+}
+</script>
+
+<style scoped  lang="scss">
+    i {
+        padding-left: 5px;
```
bbpennel, line 60: maybe bump this up to 10px, the gap between the text and the X icon for selected filters is around 12px so it would make spacing more similar

bbpennel (changes requested):
Looks good! Left two minor spacing suggestions then i think its good to go

bbpennel (approved):

PR #1402, opened 2022-05-02
Don't index empty genre elements or those without authorities

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3651

### DIFF
--- a/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
+++ b/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
@@ -572,8 +572,34 @@ public class SetDescriptiveMetadataFilter implements IndexDocumentFilter {
     }
 
     private void extractGenre(Element mods, IndexDocumentBean idb) {
-        idb.setGenre(new ArrayList<>());
-        this.addValuesToList(idb.getGenre(), mods.getChildren("genre", JDOMNamespaceUtil.MODS_V3_NS));
+        ArrayList<String> genres = new ArrayList<>();
+        List<String> genreAttrs = Arrays.asList("authority", "authorityURI", "valueURI");
+        List<Element> genreList = mods.getChildren("genre", JDOMNamespaceUtil.MODS_V3_NS);
+
+        for (Element genre : genreList) {
+            if (hasAttribute(genre, genreAttrs)) {
+                String genreValue = genre.getTextTrim();
+                if (!StringUtils.isBlank(genreValue)) {
+                    genres.add(genreValue);
+                }
+            }
+        }
+
+        if (!genres.isEmpty()) {
+            idb.setGenre(genres);
+        } else {
+            idb.setGenre(null);
+        }
+    }
+
+    private boolean hasAttribute(Element el, List<String> attributes) {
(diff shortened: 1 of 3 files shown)


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -572,8 +572,34 @@ private void extractKeywords(Element mods, IndexDocumentBean idb) {
     }
 
     private void extractGenre(Element mods, IndexDocumentBean idb) {
-        idb.setGenre(new ArrayList<>());
-        this.addValuesToList(idb.getGenre(), mods.getChildren("genre", JDOMNamespaceUtil.MODS_V3_NS));
+        ArrayList<String> genres = new ArrayList<>();
```
bbpennel, line 575: either 
```suggestion
        var genres = new ArrayList<String>();
```
or
```suggestion
        List<String> genres = new ArrayList<>();
```

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -572,8 +572,34 @@ private void extractKeywords(Element mods, IndexDocumentBean idb) {
...
+        List<String> genreAttrs = Arrays.asList("authority", "authorityURI", "valueURI");
```
bbpennel, line 576: I'd suggest declaring this as a class level constant

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilter.java
```diff
@@ -572,8 +572,34 @@ private void extractKeywords(Element mods, IndexDocumentBean idb) {
...
+            idb.setGenre(genres);
+        } else {
+            idb.setGenre(null);
+        }
+    }
+
+    private boolean hasAttribute(Element el, List<String> attributes) {
+        for (String attr : attributes) {
+            if (el.getAttribute(attr) != null &&
+                    !StringUtils.isBlank(el.getAttributeValue(attr))) {
```
bbpennel, line 598: Think we can do one retrieval of the attribute
```suggestion
            var attrValue = el.getAttributeValue(attr);
            if (!StringUtils.isBlank(attrValue)) {
```
i

indexing-solr/src/test/resources/datastream/inventoryMods.xml
```diff
@@ -4,7 +4,19 @@
   <mods:titleInfo>
     <mods:title>Paper title</mods:title>
   </mods:titleInfo>
-  <mods:genre>Dissertation</mods:genre>
+
+  <mods:genre authority="local">Dissertation</mods:genre>
```
bbpennel, line 8: I'd suggest not using the same value with different attributes (unless you're testing for handling of duplicates), otherwise we can't tell which Dissertation, Paper or Letter genre field was the one that made it through to solr.

bbpennel (changes requested):

bbpennel (approved):

PR #1404, opened 2022-05-04
Update filetype text based on how many and what filetypes are returned

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3456

### DIFF
--- a/static/js/vue-cdr-access/tests/unit/listDisplay.spec.js
+++ b/static/js/vue-cdr-access/tests/unit/listDisplay.spec.js
@@ -161,6 +161,50 @@ describe('listDisplay.vue', () => {
         expect(wrapper.vm.getFileType(wrapper.vm.recordList[1])).toEqual('Portable Network Graphics');
     });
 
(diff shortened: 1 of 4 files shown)


static/js/vue-cdr-access/tests/unit/listDisplay.spec.js
```diff
@@ -183,6 +227,50 @@ describe('listDisplay.vue', () => {
...
+            ],
+            "fileCategory": [
+                "Image"
+            ],
+            "id": "e3931f0d-a32f-4b84-b6a4-4baf0ca9b576",
+            "updated": "2019-10-29T17:22:01.830Z",
+        })).toEqual('Various');
+    });
+
+    it("file type falls back to mimetype when there's no description and displays 'Various' if more than one fileType", () => {
```
bbpennel, line 252: Need to update the description here, its the same as the previous test

bbpennel (commented):
This approach looks good to me. One minor comment, and it sounds like we'll need to do something similar for the value shown on the full record page (in the JSP part, don't need it for the File table)

bbpennel (approved):

PR #1418, opened 2022-06-03
BXC-3658 works factory & BXC-3660 File Factory

sharonluong:
This PR introduces the `WorkFactory` and the `FileFactory`. Both have creation methods, and the `WorkFactory` also has a method for adding a file to a work (and making that file the primary object).

https://jira.lib.unc.edu/browse/BXC-3658
https://jira.lib.unc.edu/browse/BXC-3660


integration/src/test/resources/spring-test/object-factory-context.xml
```diff
@@ -46,4 +46,14 @@
         <property name="updateDescriptionService" ref="updateDescriptionService" />
         <property name="repositoryObjectLoader" ref="repositoryObjectLoader" />
     </bean>
+
+    <bean id="fileFactory" class="edu.unc.lib.boxc.integration.factories.FileFactory">
```
bbpennel, line 50: Should also initialize another bean for workFactory

integration/src/test/java/edu/unc/lib/boxc/integration/factories/WorkFactory.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package edu.unc.lib.boxc.integration.factories;
+
+import edu.unc.lib.boxc.model.api.objects.WorkObject;
+
+import java.util.Map;
+
+/**
+ * @author sharonluong
```
bbpennel, line 23: since these are test utilities, it would be good to add a brief description of what these factories are for (something like "Factory for creating test WorkObjects and their files")

integration/src/test/resources/spring-test/object-factory-context.xml
sharonluong: would it have all the same content as the file factory bean?

bbpennel: I think so, except that it should also receive a fileFactory as a property

integration/src/test/java/edu/unc/lib/boxc/integration/factories/FileFactory.java
```diff
@@ -34,30 +34,32 @@
...
  * @author snluong
  */
 public class FileFactory extends ContentObjectFactory {
-    public static final String FILE_FORMAT = "fileFormat";
-    public static final String IMAGE_FORMAT = "image";
-    public static final String TEXT_FORMAT = "text";
-    public static final String PDF_FORMAT = "pdf";
-    public static final String AUDIO_FORMAT = "audio";
-    public static final String VIDEO_FORMAT = "video";
+    public static final String FILE_FORMAT_OPTION = "fileFormat";
```
bbpennel, line 41: Sorry, I was thinking it would be `FILE_FORMAT_OPTION` and the others would stay `IMAGE_FORMAT`, etc, so that it was more clear that FILE_FORMAT_OPTION was a key in the options object, while IMAGE_FORMAT was a value.

integration/src/test/java/edu/unc/lib/boxc/integration/factories/FileFactory.java
```diff
@@ -0,0 +1,111 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * @author snluong
+ */
+public class FileFactory extends ContentObjectFactory {
+    public static final String FILE_FORMAT = "fileFormat";
```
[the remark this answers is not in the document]
bbpennel: Sorry, I was thinking it would be FILE_FORMAT_OPTION and the others would stay IMAGE_FORMAT, etc, so that it was more clear that FILE_FORMAT_OPTION was a key in the options object, while IMAGE_FORMAT was a value.

bbpennel (approved):
Looks good! We should have plenty of file types to work with for our future tests

PR #1421, opened 2022-06-09
Bxc 3661 collection factory

sharonluong:
This PR adds the CollectionFactory which now allows Works to have Collections as parents. Collections have AdminUnits as parents. With that done, we can test factory creating in the CollectionsEndpointIT file.

https://jira.lib.unc.edu/browse/BXC-3661


bbpennel (approved):
I think that covers everything, good work!

PR #1423, opened 2022-06-14
BXC-3704/3706 - Stopwords in searches and collection facet

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3704
https://jira.lib.unc.edu/browse/BXC-3706

* Changes the way the operator ("AND") is specified between search terms in solr queries to prevent query failures when stopwords are present in the query
    * Solr was receiving queries like `(map AND of AND spanish AND fort)` and then removing stopwords, producing the following query, `(map AND AND spanish AND fort)`. The "AND AND" part doesn't match anything, so the query would return no results.
* Fixes a bug in the advanced search page, where the collection facet did no work. Does this by removing a mechanism that attempted to pass along search state via session, which was not working correctly and offered no real benefits.


sharonluong (approved):
nice cleanup

PR #1428, opened 2022-07-06
Reset tested component after each test

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3725


bbpennel (approved):
Unfortunate that it needs to explicitly reset (kind of weird that its needed in the tests where wrapper gets explicitly set to a new value), but it is what it is

PR #1429, opened 2022-07-08
Bxc 3729 blank search it

sharonluong:
This adds the first searchJson test and parcels out methods from the collectionsEndpointIT so they can be used in both. some are in the parent class and the assertions are in the utility class as static methods.

https://jira.lib.unc.edu/browse/BXC-3729


integration/src/test/java/edu/unc/lib/boxc/integration/web/access/CollectionsEndpointIT.java
```diff
@@ -101,23 +48,14 @@ public void setup() throws Exception {
...
-        folderFactory.createFolder(collection, Map.of("title", "Object" + System.nanoTime()));
+        createDefaultObjects();
 
         try (var resp = httpClient.execute(getMethod)) {
             var metadata = getMetadataFromResponse(resp);
 
-            assertSuccessfulResponse(resp);
-            assertValuePresent(metadata, 0, "type", "AdminUnit");
-            assertValuePresent(metadata, 1, "type", "AdminUnit");
+            SearchEndpointTestUtility.assertSuccessfulResponse(resp);
```
bbpennel, line 56: Definitely don't need to change this, just wanted to mention that there is the option to do a `import static` of these methods, similar to how the assertEquals get imported. While it can improve readability, it can also make debugging harder since it's a little more difficult to know where the methods are coming from.

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void testBlankSearchReturnsRightNumberOfObjects() throws Exception {
+        createDefaultObjects();
+
+        try (var resp = httpClient.execute(getMethod)) {
+            var metadata = getMetadataFromResponse(resp);
+            SearchEndpointTestUtility.assertSuccessfulResponse(resp);
+            // two admin units, 1 collection (nested in the admin unit), 1 work (with nested file), and 1 folder
+            assertEquals(5, metadata.size());
```
bbpennel, line 56: while it might be good to verify more details, we could probably circle back to that as we add more tests and get a clearer idea of what specific features are important to each case. So this is probably good for now.

bbpennel (changes requested):

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointTestUtility.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Utility methods for testing endpoints in the SearchActionController
```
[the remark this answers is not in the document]
sharonluong: i can move them into the parent class instead

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointTestUtility.java
```diff
@@ -0,0 +1,53 @@
...
+        assertEquals(value, result.get(key).asText());
+    }
+
+    public static void assertValuePresent(List<JsonNode> json, int index, String key) {
+        var result = json.get(index);
+        assertNotNull(result.get(key).asText());
+    }
+
+    public static void assertSuccessfulResponse(CloseableHttpResponse response) {
+        System.out.println("response = " + response);
```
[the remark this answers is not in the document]
sharonluong: i'll remove it

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
sharonluong: yeah, that's what i was thinking!

bbpennel (approved):

PR #1431, opened 2022-07-13
BXC-3736 add integration tests searching for type

sharonluong:
This adds two tests specifying search results by type.
https://jira.lib.unc.edu/browse/BXC-3736


bbpennel (approved):
short and to the point, looks good

PR #1432, opened 2022-07-14
BXC-3734 adding sort searchJson tests

sharonluong:
This adds two tests, one that specifies a title search in normal order and another that tries a title search in reversed order.
https://jira.lib.unc.edu/browse/BXC-3734

### DIFF
--- a/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
+++ b/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
@@ -86,14 +86,14 @@ public class EndpointIT {
     protected AdminUnit adminUnit1;
 
     public void createDefaultObjects() throws Exception {
-        adminUnit1 = adminUnitFactory.createAdminUnit(Map.of("title", " Admin Object1"));
-        var adminUnit2 = adminUnitFactory.createAdminUnit(Map.of("title", " Admin Object2"));
+        adminUnit1 = adminUnitFactory.createAdminUnit(Map.of("title", "Admin Object1"));
+        var adminUnit2 = adminUnitFactory.createAdminUnit(Map.of("title", "Admin Object2"));
         collection = collectionFactory.createCollection(adminUnit1,
                 Map.of("title", "Collection Object", "readGroup", "everyone"));
         work = workFactory.createWork(collection,
-                Map.of("title", " Work Object", "readGroup", "everyone"));
+                Map.of("title", "Work Object", "readGroup", "everyone"));
         var fileOptions = Map.of(
-                "title", " File Object",
+                "title", "File Object",
                 WorkFactory.PRIMARY_OBJECT_KEY, "false",
                 FileFactory.FILE_FORMAT_OPTION, FileFactory.AUDIO_FORMAT,
                 "readGroup", "everyone");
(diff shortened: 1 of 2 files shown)


integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -156,4 +156,46 @@ public void testSearchTypeParamWithFileSpecified() throws Exception {
...
+
+        try (var resp = httpClient.execute(getMethod)) {
+            var metadata = getMetadataFromResponse(resp);
+
+            assertSuccessfulResponse(resp);
+            // make sure all items return
+            assertEquals(6, metadata.size());
+            // check for the collection, should be first
+            assertValuePresent(metadata, 0, "type", "Collection");
+            assertValuePresent(metadata, 0, "title", "A first collection");
```
bbpennel, line 177: Assuming the rest of the objects have a defined order, it would be nice to include them here so that we demonstrate its not sorting by something else that happens to also make this go first (like date created or something). Just checking the title would be sufficient. I think it'd be fine to edit the default objects titles if that was necessary to produce a reliable sort order.

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -156,4 +156,46 @@ public void testSearchTypeParamWithFileSpecified() throws Exception {
...
         }
     }
+
+    @Test
+    public void testSearchWithNormalSorting() throws Exception {
+        createDefaultObjects();
+        collectionFactory.createCollection(adminUnit1,
+                Map.of("title", "A first collection", "readGroup", "everyone"));
+
+        // rollup=false disables the FileObjects getting grouped into their associated Works
```
bbpennel, line 166: This comment looks like its leftover from a previous test. Same in the other test

bbpennel (approved):
Looks good over all, left two suggestions

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
bbpennel: There is a way, but it'd be way easier to edit them in the parent test class so its like that for all tests.

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
sharonluong: removed!

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
sharonluong: done!

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
```diff
@@ -86,20 +86,20 @@ public class EndpointIT {
     protected AdminUnit adminUnit1;
 
     public void createDefaultObjects() throws Exception {
-        adminUnit1 = adminUnitFactory.createAdminUnit(Map.of("title", "Object1"));
-        var adminUnit2 = adminUnitFactory.createAdminUnit(Map.of("title", "Object2"));
+        adminUnit1 = adminUnitFactory.createAdminUnit(Map.of("title", " Admin Object1"));
```
bbpennel, line 89: I think it gets trimmed off at some point, but were the leading spaces on a number of these titles intentional? If not, it might be good to remove them to avoid confusion.

sharonluong: nope, whoops

bbpennel (approved):

PR #1439, opened 2022-07-22
Bxc 3747 front page facet link bug

sharonluong:
It was noticed that the links for "What's in the Repository" weren't working, and also that the counts were off between the front page and the actual results. This capitalizes the facet options and makes the count only return the works, not the works and the files.

https://jira.lib.unc.edu/browse/BXC-3747

### DIFF
--- a/web-common/src/main/java/edu/unc/lib/boxc/web/common/services/SolrQueryLayerService.java
+++ b/web-common/src/main/java/edu/unc/lib/boxc/web/common/services/SolrQueryLayerService.java
@@ -182,6 +182,7 @@ public class SolrQueryLayerService extends SolrSearchService {
             query.setRows(0);
             query.addFacetField(SearchFieldKey.FILE_FORMAT_CATEGORY.getSolrField());
             query.setFacetLimit(-1);
+            query.addFilterQuery("resourceType:(Work)");
 
             QueryResponse response = this.executeQuery(query);
             FacetField facetField = response.getFacetField(SearchFieldKey.FILE_FORMAT_CATEGORY.getSolrField());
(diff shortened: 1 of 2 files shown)


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/SolrQueryLayerService.java
```diff
@@ -182,6 +182,7 @@ public Map<String, Long> getFormatCounts(AccessGroupSet accessGroups) {
             query.setRows(0);
             query.addFacetField(SearchFieldKey.FILE_FORMAT_CATEGORY.getSolrField());
             query.setFacetLimit(-1);
+            query.addFilterQuery("resourceType:(Work)");
```
bbpennel, line 185: It would be best to use the constants when possible, in case fields or types get renamed later:
`SearchFieldKey.RESOURCE_TYPE.getSolrField()`
and
`ResourceType.Work.name()`

bbpennel (changes requested):
Good work, just one suggestion related to constants

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/SolrQueryLayerService.java
sharonluong: so would it be something like `query.addFilterQuery(SearchFieldKey.RESOURCE_TYPE.getSolrField() + ":(" + ResourceType.Work.name() + ")");`?

bbpennel: yes, good old java string concatenation

bbpennel (approved):

PR #1440, opened 2022-07-25
BXC-3748 - Fix ObjectTypeMismatchException thrown during work reindexing batches

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3748

* Call solr commit directly rather than sending commit message referencing the root of fedora, which was causing object tpye mismatch errors when loading the object


sharonluong (approved):
looks good

PR #1443, opened 2022-07-28
BXC-3709 - Advanced search date range fixes

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3709

* Restricts advanced search Date Deposited and Date Created fields to YEAR only
* Date Created field now searches dateCreatedYear
* Fixed a bug when generating ranges for the solr query, which was incorrectly treating both the start and the end date in the range as end dates. This was causing Date Deposited to not return records that matched the start year.
* Update display name for "added" field on search results page to "Date Deposited" for consistency (was "Date Added")

### DIFF
--- a/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
+++ b/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
@@ -309,19 +309,6 @@ public class SearchEndpointIT extends EndpointIT {
         assertResultCountEquals(SEARCH_URL + "/?added=" + (currentYear + 1) + ",", 0);
     }
 
-    private void assertResultCountEquals(String url, int expectedCount) throws IOException {
-        assertResultCountEquals(new HttpGet(url), expectedCount);
-    }
-
-    private void assertResultCountEquals(HttpGet getMethod, int expectedCount) throws IOException {
-        try (var resp = httpClient.execute(getMethod)) {
-            var metadata = getMetadataFromResponse(resp);
-            assertSuccessfulResponse(resp);
-            // two admin units, 1 collection (nested in the admin unit), 1 work (with nested file), and 1 folder
-            assertEquals(expectedCount, metadata.size());
-        }
-    }
-
     @Test
     public void testSearchWithCreatedDateRangeSpecified() throws Exception {
         createDefaultObjects();
--- a/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
+++ b/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
@@ -15,7 +15,6 @@
  */
 package edu.unc.lib.boxc.integration.web.access;
 
-import edu.unc.lib.boxc.auth.fcrepo.models.AccessGroupSetImpl;
 import edu.unc.lib.boxc.auth.fcrepo.services.GroupsThreadStore;
 import edu.unc.lib.boxc.integration.factories.FileFactory;
 import edu.unc.lib.boxc.integration.factories.WorkFactory;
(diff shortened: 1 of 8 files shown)


integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -297,6 +299,29 @@ public void testSearchPublicUserWithMetadataOnlyPermissionCanSeeMetadata() throw
...
+        assertResultCountEquals(SEARCH_URL + "/?added=" + currentYear + "," + currentYear, 5);
+        assertResultCountEquals(SEARCH_URL + "/?added=," + currentYear, 5);
+        assertResultCountEquals(SEARCH_URL + "/?added=" + (currentYear + 1) + ",", 0);
+    }
+
+    private void assertResultCountEquals(String url, int expectedCount) throws IOException {
+        assertResultCountEquals(new HttpGet(url), expectedCount);
+    }
+
+    private void assertResultCountEquals(HttpGet getMethod, int expectedCount) throws IOException {
```
sharonluong, line 316: can you put this in the parent class where i have the other custom assert methods?

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SolrSearchService.java
```diff
@@ -562,14 +562,14 @@ private void addRangeFields(SearchState searchState, SolrQuery query) {
...
-    private String getRangeValue(String key, String value) {
+    private String getRangeValue(String key, String value, boolean endRange) {
         if (StringUtils.isBlank(value)) {
             return "*";
         }
 
         if (SearchSettings.FIELDS_DATE_SEARCHABLE.contains(key)) {
             try {
-                return DateFormatUtil.getFormattedDate(value, true, true);
+                return DateFormatUtil.getFormattedDate(value, true, endRange);
```
sharonluong, line 572: what is the end range used for?

sharonluong (changes requested):

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/SolrSearchService.java
bbpennel: the DateFormatUtil takes the input in "YYYY" format and converts it into a timestamp format, which solr requires. As a result, it needs to add a month, day, etc, to the year. Since we are doing an inclusive search here, any dates that occur in the start or end year need to be included. So if we're searching for "2020,2022", it needs to search solr for the range "2020-01-01T00:00:00Z to 2022-12-31T23:59.59". The "endRange" flag tells it whether to expand to the beginning of the year or the end of the year.

sharonluong: hmm, then maybe you could rename it to `includeEndOfYear`?

bbpennel: hmm, its slightly more complicated since it interacts with the inclusive flag as well. For example, if inclusive=false then "2020,2022" would become "2020-01-01T23:59:59.999Z to 2021-12-31T23:59:59.999Z". 

The tests help make it clearer what happens:
https://github.com/UNC-Libraries/box-c/blob/main/search-solr/src/test/java/edu/unc/lib/boxc/search/solr/utils/DateFormatUtilTest.java#L61

This method is super complicated though and could use a refactor, not to mention that this is the only place it is used and the `inclusive` flag is never false now in practice (it was used at one point).

sharonluong: i suppose the tests would make it clear if i could remember what each boolean passed in was when i looked at them =P
ok then

sharonluong (approved):

PR #1444, opened 2022-07-28
Bxc 3731 date filter search

sharonluong:
This PR tests the dateCreated search, and also fixes a bug found for the unknown search (thanks @bbpennel )
https://jira.lib.unc.edu/browse/BXC-3731

### DIFF
--- a/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
+++ b/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
@@ -240,4 +240,23 @@ public class SearchEndpointIT extends EndpointIT {
             assertEquals(3, metadata.size());
         }
     }
+
+    @Test
+    public void testSearchWithCreatedDateRangeSpecified() throws Exception {
+        var adminUnit = adminUnitFactory.createAdminUnit(
+                Map.of("title", "Admin Object", "dateCreated", "2018-07-01"));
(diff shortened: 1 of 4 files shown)


integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -295,4 +295,51 @@ public void testSearchPublicUserWithMetadataOnlyPermissionCanSeeMetadata() throw
             assertEquals(6, metadata.size());
         }
     }
+
+    @Test
+    public void testSearchWithCreatedDateRangeSpecified() throws Exception {
+        createDefaultObjects();
+        var adminUnit = adminUnitFactory.createAdminUnit(
```
bbpennel, line 302: Since all 3 of these tests depend on having the same 2 objects created, it might make sense to either create them in a private method (could return the objects in a list or array), or to combine the 3 tests so the setup only happens once. The latter is the route I ended up taking over here:
https://github.com/UNC-Libraries/box-c/pull/1443/files#diff-e3d038bc6b26620f1703cd71bf02a84da5b4ba6f5779abca61ad7edc54647b39R301

sharonluong: cool, i've made a private method. i only see two tests though so have only added it to the two. if i've missed one let me know!

bbpennel (approved):

PR #1446, opened 2022-08-05
BXC-3757 - Fix incorrect thumbnails for multifile works

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3757

* Adds an integration test to reproduce the bug, where random thumbnails were being returned instead of the first one
* Fixes the bug by adding a sort to query which looks up child ids when generating thumbnail url

* Refactor QueryFilters into their own class/interface/factory, so that AccessCopiesService can use our clients for performing searches, rather than going directly to solr (my intent would be to move more query parameters into this structure later, particular the RangePair classes)
    * Updated ExportXmlProcessor to also make use of QueryFilters, since it was the only other situation where datastream queries were being performed
* Adds more tests for AccessCopiesService to help verify some of this refactoring
* Some cleanup in factories, removes extra DerivativeService instance (which was causing nullpointerexceptions in the new test) and duplicate setter methods

### DIFF
--- a/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/MultipleSelfOwnedDatastreamsFilter.java
+++ b/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/MultipleSelfOwnedDatastreamsFilter.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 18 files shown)


search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/MultipleSelfOwnedDatastreamsFilter.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Filter which restricts results to entries which contain at least one of the listed datastreams,
+ * where the datastream is owned by the object itself rather than one of its children
+ *
+ * @author bbpennel
+ */
+public class MultipleSelfOwnedDatastreamsFilter implements QueryFilter {
```
bbpennel, line 31: Name is awful

sharonluong (approved):

PR #1447, opened 2022-08-09
Bxc 3732 facet filters search

krwong:
[https://jira.lib.unc.edu/browse/BXC-3732](https://jira.lib.unc.edu/browse/BXC-3732)

- add language and subject facets to `ModsFactory.java`
- modify `getNodeFromResponse` (previously `getMetadataFromResponse`)
- move `createDatedObjects` and add `createLanguageSubjectObjects` to `EndpointIT.java`
- add integration tests for language and subject facet filter searches to `SearchEndpointIT.java`

### DIFF
--- a/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
+++ b/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
@@ -118,7 +118,6 @@ public class EndpointIT {
     }
 
     public void createLanguageSubjectObjects() throws Exception {
-        //language facet selected
         var languageAdminUnit = adminUnitFactory.createAdminUnit(
                 Map.of("title", "Admin Object", "languageTerm", "eng"));
         collectionFactory.createCollection(languageAdminUnit,
@@ -127,13 +126,13 @@ public class EndpointIT {
                         "readGroup", "everyone"));
         folderFactory.createFolder(collection,
                 Map.of("title","A language folder","languageTerm", "eng",
-                        "subject", "NorthCarolina","readGroup", "everyone"));
+                        "topic", "North Carolina","readGroup", "everyone"));
         adminUnitFactory.createAdminUnit(Map.of(
                 "title", "English Language Admin", "languageTerm", "eng",
-                "subject", "NorthCarolina", "readGroup", "everyone"));
+                "topic", "North Carolina", "readGroup", "everyone"));
         adminUnitFactory.createAdminUnit(Map.of(
                 "title", "Cherokee Language Admin", "languageTerm", "chr",
-                "subject", "UNC", "readGroup", "everyone"));
+                "topic", "UNC", "readGroup", "everyone"));
     }
 
     public List<JsonNode> getNodeFromResponse(CloseableHttpResponse response, String fieldName) throws IOException {
--- a/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
+++ b/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
@@ -104,42 +104,15 @@ public class EndpointIT {
                 Map.of("title", "Folder Object", "readGroup", "everyone"));
     }
 
-    public List<String> createDatedObjects() throws Exception {
-        List<String> ids = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -348,16 +347,58 @@ public void testSearchWithCreatedDateRangeSpecifiedAsUnknown() throws Exception
...
+            assertValuePresent(languageFields,1,"value", "Cherokee");
+        }
+    }
+
+    @Test
+    public void testSearchWithSubjectFilterSelectedDisplaysAvailableLanguages() throws Exception {
+        createDefaultObjects();
+        createLanguageSubjectObjects();
+
+        var getMethod = new HttpGet(SEARCH_URL + "/?facetSelect=language%2Csubject&subject=North%2520Carolina&getFacets=true");
```
bbpennel, line 390: I'm not sure why the access app is double encoding the spaces, looks like it can be either `%20` or `+`, but that is indeed what it is doing, hmm. Maybe we can correct that in the future.

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -348,16 +347,58 @@ public void testSearchWithCreatedDateRangeSpecifiedAsUnknown() throws Exception
...
+
+        try (var resp = httpClient.execute(getMethod)) {
+            var facetFields = getNodeFromResponse(resp, "facetFields");
+            var languageFields = new ArrayList<JsonNode>();
+            facetFields.get(0).get("values").elements().forEachRemaining(languageFields::add);
+
+            assertSuccessfulResponse(resp);
+            // should find all available languages: English
+            assertEquals(2,facetFields.size());
```
bbpennel, line 399: Since this is the only test that is verifying two facets being selected, it'd be good to add an assertion that the subject field is also being returned.

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
```diff
@@ -103,11 +104,42 @@ public void createDefaultObjects() throws Exception {
                 Map.of("title", "Folder Object", "readGroup", "everyone"));
     }
 
-    public List<JsonNode> getMetadataFromResponse(CloseableHttpResponse response) throws IOException {
+    public List<String> createDatedObjects() throws Exception {
```
bbpennel, line 107: Were these creation methods moved up to the parent class in anticipation that they'll be needed by other tests later?

bbpennel (changes requested):

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/EndpointIT.java
sharonluong: I thought it'd be more consistent to have the object creation methods in one place, but I suppose we could put the more specialized ones in the SearchEndpointIT

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
sharonluong: I have no idea what this means

bbpennel: Just remarking on `North%2520Carolina`, which contains a space that normally would be encoded as a `+` or `%20`, but has been double encoded from `%2520` (the % got encoded to `%25`). Its not a problem with this PR, just a not great thing

bbpennel (approved):

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
sharonluong: ah, is that just something @krwong added manually? or does it actually look like this on prod?

bbpennel: It looks like that on prod, but it works with any of the 3 encodings i listed.

PR #1450, opened 2022-08-11
Draft: Update to UV 4.x

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3728

Note: Containing div must have a class named "uv" or it won't pick up styling from UV itself
```
<div id="jp2_viewer" class="jp2_imageviewer_window uv" data-url="${briefObject.id}"></div>
```


lfarrell:
Waiting on some upstream fixes

- https://github.com/UniversalViewer/universalviewer/issues/823
- https://github.com/UniversalViewer/universalviewer/issues/872

I think we're hitting a couple of problems. UV4 works with IIIF 3 and we're using 2. They're mostly compatible, but we're hitting the issues above. Plus some of the language seems to have changed so instead of <region>/<size> full/full it's now full/max. The Java IIIF library we use hasn't been updated to add v3 support and seems like it won't be. We might need to switch to jiiify-presentation.

bbpennel:
Closing, we already updated to uv4 as a part of https://github.com/UNC-Libraries/box-c/pull/1553

PR #1452, opened 2022-08-11
Initial Code Climate config

cazzerson:
* Use our checkstyle config
* Ignore some external libraries
* Enable some plugins

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,4 @@
+version: "2"
 plugins:
   checkstyle:
     enabled: true
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,13 +3,13 @@ plugins:
   checkstyle:
     enabled: true
     config:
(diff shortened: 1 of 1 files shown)


bbpennel:
I can't get codeclimate to run for this branch, it always gives the error:
```
Comparison results can't be displayed properly due to high issue volume. Please contact an admin to manually approve this PR.
```
I am wondering if the issue is that it won't run the checks using a configuration not found on main, or perhaps using new configurations is simply too resource intensive. Googling it is not turning up much. I'm inclined to try merging this PR to see if it fairs any better on main.

bbpennel (approved):

bbpennel:
I think it is working? I believe the Maintainability rating when from C to B after merging this, with 6months of work instead of a year, and I don't see our various external js libraries as topping the list of issues anymore.

cazzerson:
Thanks! I was seeing inconsistent results with it on the branch. It would first run with errors and then fail to run at all. Anyways, this sounds like a good first step to moving towards metrics worth reviewing and/or acting on.

PR #1453, opened 2022-08-12
Bxc 3771 return fields tests

krwong:
[https://jira.lib.unc.edu/browse/BXC-3771](https://jira.lib.unc.edu/browse/BXC-3771)

- add fields to `ModsFactory.java`
- add `AssertStringValuePresent` and `AssertArrayValuePresent`
- create integration tests to verify all record fields returned

### DIFF
--- a/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
+++ b/integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
@@ -535,11 +535,14 @@ public class SearchEndpointIT extends EndpointIT {
             var metadata = getMetadataFromResponse(resp);
             assertSuccessfulResponse(resp);
 
+            var workId = metadata.get(0).get("id").asText();
+
             //work assertions
             assertValuePresent(metadata, 0, "title", "Work Record");
             assertValuePresent(metadata, 0, "type", "Work");
             assertArrayValuePresent(metadata, 0, "creator", "[\"Creator?\"]");
             assertArrayValuePresent(metadata, 0, "counts");
+            assertValuePresent(metadata, 0, "rollup", workId);
 
             //file assertions
             assertStringValuePresent(metadata, 1, "thumbnail_url");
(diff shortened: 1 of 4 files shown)


integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -527,6 +525,41 @@ public void testFacetCountsReflectUserPermissions() throws Exception {
...
+
+        var getMethod = new HttpGet(SEARCH_URL + "/?types=Work,File&rollup=false");
+
+        try (var resp = httpClient.execute(getMethod)) {
+            var metadata = getMetadataFromResponse(resp);
+            assertSuccessfulResponse(resp);
+
+            assertStringValuePresent(metadata, 1, "thumbnail_url");
+            assertStringValuePresent(metadata, 1, "id");
+            assertStringValuePresent(metadata, 1, "title");
```
bbpennel, line 540: It would help with readability to group the assertions on the same object together, so it's clear what the expectations are for the work versus the file.

In addition to verifying that a field is present, I think we should check that some of the fields with known, verifiable values, are populated. This could help us avoid problems if any expectations change, like empty fields return an empty string instead of nulls, etc, or we get back an array with unexpected values. I'd suggest doing this for title, resourceType, creator, format, fileType, identifier, rollup.

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
```diff
@@ -558,4 +591,21 @@ public void createLanguageSubjectObjects() throws Exception {
                 "title", "Cherokee Language Admin", "languageTerm", "chr",
                 "topic", "UNC", "readGroup", "everyone"));
     }
+
+    public void createWorkAndFileObjects() throws Exception {
```
bbpennel, line 595: It was a little unexpected that this method creates the admin unit and collection, rather than depending on the standard method for creating the structure. I'm fine with this, it should make the test a bit faster, but could you add a comment to the method to indicate that it creates the full structure, rather than just the work and file objects mentioned in the method name?

bbpennel (changes requested):

integration/src/test/java/edu/unc/lib/boxc/integration/web/access/SearchEndpointIT.java
krwong: I checked the values for all the fields you suggested except rollup. Rollup changed every time I ran the tests.

bbpennel: The value for rollup is the id of the work, so you should be able to get it from the work object

bbpennel (approved):

PR #1455, opened 2022-08-16
Don't show user download button if they don't have access

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3694

### DIFF
--- a/static/js/public/fullRecord.js
+++ b/static/js/public/fullRecord.js
@@ -72,14 +72,11 @@ define('fullRecord', ['module', 'jquery', 'JP2Viewer', 'StructureView', 'dataTab
 	}
 
 	if ($childFilesTable.length > 0) {
-		var excluded_columns = [0, 4, 5];
+		var excluded_columns = [0, 4];
 		var column_defs = [
 			{ orderable: false, targets: excluded_columns },
 			{ searchable: false, target: excluded_columns },
 			{ type: 'file-size', targets: 3 },
-			{ width: '5%', targets: [0, 4, 5] },
-			{ width: '20%', targets: [2, 3] },
-			{ width: '40%', targets: 1 },
 			{ render: function (data, type, row, meta) {
 				var img;
 				var hasIntersectionObserver = 'IntersectionObserver' in window;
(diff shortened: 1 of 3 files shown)


static/js/public/fullRecord.js
```diff
@@ -115,20 +112,24 @@ define('fullRecord', ['module', 'jquery', 'JP2Viewer', 'StructureView', 'dataTab
...
 			{ render: function (data, type, row) { return getOriginalFileValue(row.datastream, 'file_size');  }, targets: 3 },
 			{ render: function (data, type, row) { return '<a href="/record/' + row.id + '" aria-label="View ' + row.title +'">' +
 					'<i class="fa fa-search-plus is-icon"' + ' title="View"></a>'; },
 				targets: 4
 			},
-			{ render: function (data, type, row) { return '<a href="/indexablecontent/' + row.id + '?dl=true" aria-label="Download ' + row.title +'">' +
-					'<i class="fa fa-download is-icon" title="Download"></a>'; },
+			{ render: function (data, type, row) {
+					if (row.permissions.indexOf('viewOriginal') === -1) {
+						return '<i class="fa fa-download is-icon no-download" title="Download Unavailable">';
```
bbpennel, line 124: I know the browser corrects this, but shouldn't this have a closing `</i>`? Looks like on the font-awesome site has it that way, and looks similar to the error listed here https://html.spec.whatwg.org/#syntax-errors labeled as "Errors that result in disproportionately poor performance". Looks like this is happening in a few other places in this file.

bbpennel (changes requested):

static/js/public/fullRecord.js
```diff
@@ -103,8 +103,8 @@ define('fullRecord', ['module', 'jquery', 'JP2Viewer', 'StructureView', 'dataTab
 
 					img += '<div class="thumbnail-badge thumbnail-badge-' + whichBadge + '">' +
 							'<div class="fa-stack">' +
-								'<i class="fas fa-circle fa-stack-2x background"></i>' +
-								'<i class="fas fa-' + whichBadge + ' fa-stack-1x foreground"></i>' +
+								'<i class="fas fa-circle fa-stack-2x background"></i></i>' +
```
bbpennel, line 106: Looks like this got double closed? Same with the next line

lfarrell: Updated.

bbpennel (approved):

PR #1458, opened 2022-08-18
BXC-3780 - Sanitize filenames for FITS CLI

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3780
* Sanitize filenames submitted to FITS cli by symlinking to the file, using a filename with the reserved and non-ascii characters removed


sharonluong (approved):
Looks good

PR #1459, opened 2022-08-18
BXC-3758 adding test and methods to get parentCollectionId

sharonluong:
https://jira.lib.unc.edu/browse/BXC-3758

As Ben notes, the url is getting generated with a value like:
https://dcr-test.lib.unc.edu/record/This%20is%20a%20serial%20title%7Ca168cf29-a2a9-4da8-9b8d-025855b180d5
but it should be:
https://dcr-test.lib.unc.edu/record/a168cf29-a2a9-4da8-9b8d-025855b180d5

so this PR adds methods to get the second part of the string and return it accordingly


bbpennel (approved):
Looks good!

PR #1464, opened 2022-09-01
BXC-3785 add ClearOrderValidator

sharonluong:
Adding a ClearOrderValidator that checks whether the target is a work object, with a test and adding a switch statement to the OrderValidator class.

https://jira.lib.unc.edu/browse/BXC-3785

### DIFF
--- a/operations/src/main/java/edu/unc/lib/boxc/operations/impl/order/ClearOrderValidator.java
+++ b/operations/src/main/java/edu/unc/lib/boxc/operations/impl/order/ClearOrderValidator.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.boxc.operations.impl.order;
+
+import edu.unc.lib.boxc.model.api.ResourceType;
+import edu.unc.lib.boxc.model.api.objects.RepositoryObjectLoader;
+import edu.unc.lib.boxc.operations.api.order.OrderValidator;
+import edu.unc.lib.boxc.operations.jms.order.OrderRequest;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author snluong
+ */
+public class ClearOrderValidator implements OrderValidator {
+    private RepositoryObjectLoader repositoryObjectLoader;
+    private OrderRequest request;
+    private Boolean result;
+    private List<String> errors = new ArrayList<>();
+
+    @Override
+    public boolean isValid() {
+        if (result != null) {
+            return result;
+        }
+        result = validate();
+        return result;
+    }
+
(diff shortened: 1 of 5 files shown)


operations/src/main/java/edu/unc/lib/boxc/operations/impl/order/ClearOrderValidator.java
```diff
@@ -0,0 +1,67 @@
...
+
+import edu.unc.lib.boxc.model.api.ResourceType;
+import edu.unc.lib.boxc.model.api.objects.RepositoryObjectLoader;
+import edu.unc.lib.boxc.operations.api.order.OrderValidator;
+import edu.unc.lib.boxc.operations.jms.order.OrderRequest;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
```
bbpennel, line 26: Please add a short description of this class

operations/src/main/java/edu/unc/lib/boxc/operations/impl/order/OrderValidatorFactory.java
```diff
@@ -35,11 +35,23 @@ public class OrderValidatorFactory {
...
+            return validator;
+        }
+        case CLEAR: {
+            var validator = new ClearOrderValidator();
+            validator.setRepositoryObjectLoader(repositoryObjectLoader);
+            validator.setRequest(request);
+            return validator;
+        }
+        default:
+            throw new UnsupportedOperationException("Unable to determine order for request " + request);
```
bbpennel, line 53:
```suggestion
            throw new UnsupportedOperationException("Unable to determine validator for request " + request);
```

operations/src/test/java/edu/unc/lib/boxc/operations/impl/order/OrderValidatorFactoryTest.java
```diff
@@ -58,4 +58,10 @@ public void forSetOrderRequestTest() {
         var request = OrderRequestFactory.createRequest(OrderOperationType.SET, PARENT_UUID, Arrays.asList(CHILD1_UUID));
         assertTrue(factory.createValidator(request) instanceof SetOrderValidator);
     }
+
+    @Test
+    public void forClearOrderRequestTest() {
+        var request = OrderRequestFactory.createRequest(OrderOperationType.CLEAR, PARENT_UUID, Arrays.asList(CHILD1_UUID));
```
bbpennel, line 64: can leave off the last parameter, for ClearOrder requests we only need to say what object we are clearing the order of.
```suggestion
        var request = OrderRequestFactory.createRequest(OrderOperationType.CLEAR, PARENT_UUID);
```

operations/src/main/java/edu/unc/lib/boxc/operations/impl/order/ClearOrderValidator.java
```diff
@@ -0,0 +1,67 @@
...
+        result = validate();
+        return result;
+    }
+
+    private boolean validate() {
+        var parentId = request.getParentPid().getId();
+        var parentObj = repositoryObjectLoader.getRepositoryObject(request.getParentPid());
+        if (!ResourceType.Work.equals(parentObj.getResourceType())) {
+            errors.add("Object " + parentId + " of type " + parentObj.getResourceType().name()
+                    + " does not support setting member order");
```
bbpennel, line 49: Perhaps we should remove the word "setting" here and in the other validator. In both cases we're really validating whether the object supports ordering at all, rather than whether it supports any particular operations (set, clear, etc)

bbpennel (changes requested):
Looks good, added a few minor suggestions

operations/src/main/java/edu/unc/lib/boxc/operations/impl/order/ClearOrderValidator.java
sharonluong: makes sense! i changed them both

bbpennel (approved):

PR #1465, opened 2022-09-06
Add buttons for restricted content access

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3782

### DIFF
--- a/web-access-app/src/main/webapp/WEB-INF/jsp/error/invalidRecord.jsp
+++ b/web-access-app/src/main/webapp/WEB-INF/jsp/error/invalidRecord.jsp
@@ -27,9 +27,9 @@
 
 		<p>Or if you believe the record exists and would like to get access to it,  
 			<c:if test="${empty sessionScope.user || empty sessionScope.user.userName}">
-				try <a href="<c:out value='${loginUrl}'/>">logging in</a> or
+				try <a href="<c:out value='${loginUrl}'/>">logging in (UNC Onyen)</a> or
 			</c:if>
-			you may <a href="">request</a> permission.
+			you may <a href="https://library.unc.edu/wilson/contact/?refer=https%3A%2F%2Fdcr.lib.unc.edu">Contact Wilson Library for access information</a>.
 		</p>
 	</div>
 </div>
--- a/web-access-app/src/main/webapp/WEB-INF/jsp/error/invalidRecord.jsp
+++ b/web-access-app/src/main/webapp/WEB-INF/jsp/error/invalidRecord.jsp
@@ -29,7 +29,7 @@
 			<c:if test="${empty sessionScope.user || empty sessionScope.user.userName}">
 				try <a href="<c:out value='${loginUrl}'/>">logging in (UNC Onyen)</a> or
 			</c:if>
-			you may <a href="https://library.unc.edu/wilson/contact/?refer=https%3A%2F%2Fdcr.lib.unc.edu">Contact Wilson Library for access information</a>.
+			you may <a href="${contactUrl}">Contact Wilson Library for access information</a>.
 		</p>
 	</div>
 </div>
--- a/web-access-app/src/main/webapp/WEB-INF/jsp/fullRecord/accessInfo.jsp
+++ b/web-access-app/src/main/webapp/WEB-INF/jsp/fullRecord/accessInfo.jsp
@@ -1,0 +1,33 @@
+<%--
+
+    Copyright 2008 The University of North Carolina at Chapel Hill
+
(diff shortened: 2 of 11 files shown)


web-access-app/src/main/webapp/WEB-INF/jsp/fullRecord/accessInfo.jsp
```diff
@@ -0,0 +1,31 @@
...
+<%@ taglib prefix="fmt" uri="http://java.sun.com/jsp/jstl/fmt" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn" %>
+<%@ taglib prefix="cdr" uri="http://cdr.lib.unc.edu/cdrUI"%>
+
+    <div class="restricted-access">
+        <h2>This ${fn:toLowerCase(briefObject.resourceType)} has restricted content</h2>
+        <c:if test="${permsHelper.enhancedAuthenticatedAccess(accessGroupSet, briefObject)}">
+            <div class="actionlink"><a class="button" href="${loginUrl}"><i class="fa fa-id-card"></i> Log in for access (UNC Onyen)</a></div>
+        </c:if>
+        <div class="actionlink"><a class="button" href="https://library.unc.edu/wilson/contact/?refer=https%3A%2F%2Fdcr.lib.unc.edu"><i class="fa fa-envelope"></i> Contact Wilson Library for access</a></div>
```
bbpennel, line 30: I'm realizing this is the same url as we are linking to from the Contact Us url in the footer, so you should be able to use `${contactUrl}` for the href here, like this:
https://github.com/UNC-Libraries/box-c/blob/main/web-access-app/src/main/webapp/WEB-INF/jsp/common/footer.jsp#L27
It'd be good to include the full url of the page being requested anyways.

web-access-app/src/main/webapp/WEB-INF/jsp/error/invalidRecord.jsp
```diff
@@ -27,9 +27,9 @@
 
 		<p>Or if you believe the record exists and would like to get access to it,  
 			<c:if test="${empty sessionScope.user || empty sessionScope.user.userName}">
-				try <a href="<c:out value='${loginUrl}'/>">logging in</a> or
+				try <a href="<c:out value='${loginUrl}'/>">logging in (UNC Onyen)</a> or
 			</c:if>
-			you may <a href="">request</a> permission.
+			you may <a href="https://library.unc.edu/wilson/contact/?refer=https%3A%2F%2Fdcr.lib.unc.edu">Contact Wilson Library for access information</a>.
```
bbpennel, line 32: See the comment below about using contactUrl

web-access-app/src/main/webapp/WEB-INF/jsp/fullRecord/adminUnit.jsp
```diff
@@ -42,43 +42,48 @@
 
 <c:url var="fullRecordUrl" scope="page" value="record/${briefObject.id}" />
 <div class="contentarea">
-    <div id="is-admin-unit" class="columns browse-header">
```
bbpennel, line 45: I am guessing all the changes in this file are converting the admin unit page over to following the same layout as the other pages, rather than changes directly related to the access request buttons? So much reformatting, but seems necessary.

One thing we should update though, it looks like before these changes, if I shrink my browser down to mobile widths, all the parts properly display in rows. But after these changes the breadcrumbs are getting smushed into the left hand side, and all the rest of the content is on the right. This is happening for AdminUnit and Collection pages, but not folder, work or file.

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/PermissionsHelper.java
```diff
@@ -139,6 +145,25 @@ public boolean hasEditAccess(AccessGroupSet principals, ContentObjectRecord meta
         return accessControlService.hasAccess(metadata.getPid(), principals, editDescription);
     }
 
+    /**
+     * True if authenticated permissions are greater than no access
+     * @param metadata
+     * @return
+     */
+    public boolean enhancedAuthenticatedAccess(AccessGroupSet principals, ContentObjectRecord metadata) {
```
bbpennel, line 153: To stick with naming conventions, this should begin with a verb like`has` or `allow`.

My reading of the method is that it returns true if the Authenticated group has `canViewOriginals`, rather than having permission greater than no access, since neither canViewMetadata or canViewAccessCopies return true. Maybe `allowsFullAuthenticatedAccess` would make sense? Should also update the javadoc

bbpennel (changes requested):

bbpennel (approved):

PR #1479, opened 2022-10-12
Bxc 3791 remove child job

sharonluong:
Creates functionality for REMOVE_FROM order type action and adds tests
https://jira.lib.unc.edu/browse/BXC-3791


bbpennel (approved):
Looks good. I usually default to ArrayList when I need a modifiable List, but LinkedList is fine here. I think it's okay that we're modifying the list that comes back from `getMemberOrder`, although I had to run a test to make sure the list it returns in modifiable. It wouldn't be if there is no order set, but I think the validator should catch that case before it gets here.

PR #1480, opened 2022-10-13
BXC-3792 remove job validator code and moving methods to helpers

sharonluong:
This PR adds a validator for the RemoveFromOrderJob. In doing so, methods in common with SetOrderJob and its validators/tests were moved to helpers as well.
https://jira.lib.unc.edu/browse/BXC-3792

### DIFF
--- a/operations/src/main/java/edu/unc/lib/boxc/operations/api/order/MemberOrderHelper.java
+++ b/operations/src/main/java/edu/unc/lib/boxc/operations/api/order/MemberOrderHelper.java
@@ -18,6 +18,12 @@ package edu.unc.lib.boxc.operations.api.order;
 import edu.unc.lib.boxc.model.api.ResourceType;
 import edu.unc.lib.boxc.model.api.ids.PID;
 
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
 /**
  * Helpers for member order operations
  *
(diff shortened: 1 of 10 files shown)


operations/src/main/java/edu/unc/lib/boxc/operations/api/order/MemberOrderHelper.java
```diff
@@ -46,4 +52,33 @@ public static String formatUnsupportedMessage(PID pid, ResourceType resourceType
...
+
+    /**
+     *
+     * @param parentId UUID of parent object
+     * @param reason reason child objects may not be ordered
+     * @param problemPids PIDs of child objects that are not valid to be ordered
+     * @return formatted error message
+     */
+    public static String formatErrorMessage(String parentId, String reason, Collection<PID> problemPids) {
+        return "Invalid request to set order for " + parentId
```
bbpennel, line 64: I'm trying to decide if it reads weird or not that the REMOVE_FROM operation would emit an error saying that "set order" is not supported. Maybe its fine, since what we're reporting is really that the object type doesn't support ordering at all, rather than that a particular order modification operation isn't supported. What do you think?

operations/src/test/java/edu/unc/lib/boxc/operations/impl/order/RemoveFromOrderValidatorTest.java
```diff
@@ -0,0 +1,96 @@
...
+        when(membershipService.listMembers(parentPid)).thenReturn(Arrays.asList(child1Pid, child2Pid, child3Pid));
+
+        var request = OrderRequestFactory.createRequest(OrderOperationType.REMOVE_FROM, PARENT_UUID,
+                Arrays.asList(CHILD1_UUID, CHILD2_UUID, CHILD1_UUID));
+        validator.setRequest(request);
+
+        assertFalse(validator.isValid());
+        assertHasErrors(validator,"Invalid request to set order for " + PARENT_UUID
+                + ", it contained duplicate member IDs: " + CHILD1_UUID);
+    }
```
bbpennel, line 95: We should add a test where `validator.isValid()` returns `true` and no errors

bbpennel (changes requested):
This looks good over all, i left one question and a suggestion

operations/src/main/java/edu/unc/lib/boxc/operations/api/order/MemberOrderHelper.java
sharonluong: hmm, i suppose i could change it to something like `Invalid request to amend order for...` ?

bbpennel: Hm, that sounds okay but we aren't using "amend" anywhere else related to order. Another could be "Invalid member ordering request for". Passing in the verb might be easier than coming up with verbiage that works for everything though.

sharonluong: alright, i just passed it in

bbpennel (approved):

PR #1482, opened 2022-10-21
Bxc 3790 trigger remove job

sharonluong:
This PR adds a trigger for the remove order job during the delete object job.
https://jira.lib.unc.edu/browse/BXC-3790

### DIFF
--- a/operations/src/test/java/edu/unc/lib/boxc/operations/impl/destroy/DestroyObjectsJobIT.java
+++ b/operations/src/test/java/edu/unc/lib/boxc/operations/impl/destroy/DestroyObjectsJobIT.java
@@ -50,6 +50,9 @@ import edu.unc.lib.boxc.operations.api.events.PremisLoggerFactory;
 import edu.unc.lib.boxc.operations.jms.MessageSender;
 import edu.unc.lib.boxc.operations.jms.destroy.DestroyObjectsRequest;
 import edu.unc.lib.boxc.operations.jms.indexing.IndexingMessageSender;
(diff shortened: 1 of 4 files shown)


operations/src/test/java/edu/unc/lib/boxc/operations/impl/destroy/DestroyObjectsJobIT.java
```diff
@@ -225,6 +232,9 @@ public void destroySingleFileObjectTest() throws Exception {
         verify(binaryDestroyedMessageSender).sendMessage(docCaptor.capture());
 
         assertMessagePresent(docCaptor.getAllValues(), filesToCleanup, null);
+
+        // assert that the remove order request was sent
+        verify(memberOrderRequestSender).sendToQueue(requestCaptor.capture());
```
bbpennel, line 237: If we don't need to inspect the contents of the message sent, then we can use `any()` instead of using a captor. Alternatively, we could keep the catpor and verify that the request contains the parent and child ids.

bbpennel (changes requested):
This looks good! One suggestion related to the test

operations/src/test/java/edu/unc/lib/boxc/operations/impl/destroy/DestroyObjectsJobIT.java
sharonluong: alright, i added the latter!

bbpennel (approved):

PR #1484, opened 2022-11-03
BXC-3789 add memberOrder property to archival resource for Fedora

sharonluong:
This adds the memberOrder property to the AIP properties.
https://jira.lib.unc.edu/browse/BXC-3789


bbpennel (approved):
Looks good!

PR #1499, opened 2022-12-07
UI for importing/exporting member order

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3906


static/js/admin/src/ExportMenu.js
```diff
@@ -0,0 +1,101 @@
...
+
+                    return {
+                        className: 'add_to_container_menu',
+                        events : {
+                            show: function() {
+                                this.addClass("active");
+                            },
+                            hide: function() {
+                                this.removeClass("active");
+                                sessionStorage.clear();
```
bbpennel, line 65: Will this clear everything in the sessionStorage for the dcr? Would it make sense to just clear the `exportTargets` field?

bbpennel (changes requested):
Looks good over all, left two comments

bbpennel (approved):

PR #1515, opened 2023-01-24
BXC-3941 - UTC when indexing date fields

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3941

* Treat incoming dates as being in UTC timezone when parsing unless otherwise specified. 
* When extracting year for dateCreatedYear, extract the year using an instant in UTC timezone to avoid getting the wrong year due to timezone conversions
* Change unit test to testing public method, and adjust expectations based on UTC being the default timezone.

### DIFF
--- a/common-utils/src/test/java/edu/unc/lib/boxc/common/util/DateTimeUtilTest.java
+++ b/common-utils/src/test/java/edu/unc/lib/boxc/common/util/DateTimeUtilTest.java
@@ -32,65 +32,44 @@ public class DateTimeUtilTest {
     }
 
     @Test
-    public void utcToDateTimeTest() throws Exception {
-        DateTime date = DateTimeUtil.parseUTCToDateTime("2001");
-        assertEquals(2001, date.getYear());
+    public void parseUTCToDateYearTest() throws Exception {
+        var date = DateTimeUtil.parseUTCToDate("2001");
+        assertEquals("2001-01-01T00:00:00.000Z", DateTimeUtil.formatDateToUTC(date));
+    }
 
-        date = DateTimeUtil.parseUTCToDateTime("2001-05-08");
-        assertEquals(2001, date.getYear());
-        assertEquals(5, date.getMonthOfYear());
-        assertEquals(8, date.getDayOfMonth());
+    @Test
+    public void parseUTCToDateYyyymmddTest() throws Exception {
+        var date = DateTimeUtil.parseUTCToDate("2001-05-08");
+        assertEquals("2001-05-08T00:00:00.000Z", DateTimeUtil.formatDateToUTC(date));
+    }
 
-        date = DateTimeUtil.parseUTCToDateTime("2002-02-01T12:13:14");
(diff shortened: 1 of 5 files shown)


common-utils/src/test/java/edu/unc/lib/boxc/common/util/DateTimeUtilTest.java
```diff
@@ -32,65 +32,44 @@ public void after() {
...
-    public void utcToDateTimeTest() throws Exception {
-        DateTime date = DateTimeUtil.parseUTCToDateTime("2001");
-        assertEquals(2001, date.getYear());
-
-        date = DateTimeUtil.parseUTCToDateTime("2001-05-08");
-        assertEquals(2001, date.getYear());
-        assertEquals(5, date.getMonthOfYear());
-        assertEquals(8, date.getDayOfMonth());
+    public void parseUTCToDateYearTest() throws Exception {
+        var date = DateTimeUtil.parseUTCToDate("2001");
```
sharonluong, line 36: So the DateTimeUtil method `parseUTCToDateTime` will go away when we have java 8 time implementation?

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
```diff
@@ -363,4 +362,18 @@ public void fileTitle() {
 
         assertEquals("uuid: 1234", idb.getTitle());
     }
+
+    @Test
+    public void testDateCreatedYearBxc3941() throws Exception {
```
sharonluong, line 367: is the BXC3941 the ticket number? that might be confusing in the future

sharonluong (approved):
Just a question/suggestion but looks good!

common-utils/src/test/java/edu/unc/lib/boxc/common/util/DateTimeUtilTest.java
bbpennel: yes, its only used internally and just there because we're stuck with the joda time library for the moment (DateTime is a joda time class)

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDescriptiveMetadataFilterTest.java
bbpennel: Yes, its the ticket number. I've seen other projects list the ticket that describes the bug being tested for. I could put it as a comment rather than in the name of the method if that's clearer

sharonluong: ohh i hadn't seen that convention before. interesting! i guess it's fine with me to leave it

PR #1517, opened 2023-01-26
Bxc 3945 upgrade to junit 5

krwong:
[https://jira.lib.unc.edu/browse/BXC-3945](https://jira.lib.unc.edu/browse/BXC-3945)

- added junit 5 dependencies to pom.xml and removed old junit dependencies
- modified junit-related imports
- changed @ Test annotations for specifying expectations to assertThrows
- changed other annotations (@ Before to @ BeforeEach, @ After to @ AfterEach, etc)
- create temporary folders using @ TempDir instead of @ Rule
- modified some tests so they work in junit 5
- changed the order of parameters in assertions (output message is last parameter)
- added EmbeddedActiveMQBroker test utility to web-services-app
- added mockito-junit-jupiter dependency to web-common pom.xml

The services-camel-app submodule will be updated at a later date. It requires a camel upgrade (2.25 => 3.x) to work with junit 5.


bbpennel (approved):
Great work, thank you for your perseverance!

PR #1518, opened 2023-01-27
WIP adjust codeclimate thresholds

krwong:
I can't get codeclimate to work locally so I'll adjust the thresholds this way instead


krwong:
Accidentally manually approved codeclimate and it won't make checks anymore. Have to create a new WIP PR

PR #1525, opened 2023-02-10
BXC-3951 - Update display of archival coll id

agoslen:


web-access-app/src/main/resources/recordTransformations/modsToFullRecord.xsl
```diff
@@ -920,11 +949,19 @@
 								<xsl:with-param name="field" select="*[local-name() = 'note' and not(@type = 'admin')]"/>
 							</xsl:call-template>
 
-							<xsl:call-template name="modsGroupedFieldWithType">
-								<xsl:with-param name="defaultLabel">Identifier</xsl:with-param>
-								<xsl:with-param name="field" select="*[local-name() = 'identifier']"/>
-							</xsl:call-template>
+							<xsl:variable name="collectionIdentifier" select="*[local-name() = 'identifier' and @displayLabel='Collection Number']"/>
+							<xsl:variable name="identifier" select="*[local-name() = 'identifier' and not(@displayLabel='Collection Number')]"/>
+							<xsl:if test="boolean($collectionIdentifier) or boolean($identifier)">
```
bbpennel, line 954: Is the conditional wrapping both of these template calls needed? It seems like typically the templates don't display anything if the the selector didn't return any elements. Did the display look weird without it?

bbpennel (approved):
I think these changes look good, although its kind of too bad we have to keep adding new templates. I left one question, but probably don't have to change anything.

web-access-app/src/main/resources/recordTransformations/modsToFullRecord.xsl
agoslen: It seems to work okay if I remove the variables and the conditional...I think I was just trying to make it more similar to the code in the other section. I will remove and update

PR #1526, opened 2023-02-13
Bxc 3915 migrate footer to vuejs

krwong:
[https://jira.lib.unc.edu/browse/BXC-3915](https://jira.lib.unc.edu/browse/BXC-3915)

### DIFF
--- a/static/js/vue-cdr-access/src/components/footer.vue
+++ b/static/js/vue-cdr-access/src/components/footer.vue
@@ -1,0 +1,45 @@
+<template>
+    <footer>
+        <p class="footer-menu container">
+            <a href="${accessBaseUrl}">Home</a> |
+            <a href="collections">Browse Collections</a> |
+            <a href="aboutRepository">About</a> |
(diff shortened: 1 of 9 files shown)


static/js/vue-cdr-access/src/components/footer.vue
```diff
@@ -0,0 +1,45 @@
+<template>
```
bbpennel, line 1: I know there isn't much logic in this template, but it would be good to add a simple test to ensure that the footer template loads and maybe verify that the Home link goes to where we expect it to
https://github.com/UNC-Libraries/box-c/pull/1522/files#diff-5b6a675d270ec776826682eab6299b2ba8b1d89a33587b8edfa38ebb09383acc

bbpennel (changes requested):

lfarrell (changes requested):
The footer won't display for me. I'll take a look at this. This could be a bit tricky, since it needs to display on non-vue pages too.

PR #1530, opened 2023-02-17
Bxc 4004 adjust sort in work

sharonluong:
https://jira.lib.unc.edu/browse/BXC-4004

This PR removes the initial title sort so that results come back as is from solr. This would reflect the member order if it was set. It also fixes the fact that solr was returning up to 2000 results at once. Now it returns 10 to match the pagination in datatables.


bbpennel (approved):
Looks good!

PR #1538, opened 2023-03-10
Bxc 4009 access copy api

sharonluong:
https://jira.lib.unc.edu/browse/BXC-4009

This PR updates some dependencies (adds Wiremock and moves fuseki to a more localized pom.xml). It mainly creates an API that will take a size request (either a pixel length or "full"), format the request for the IIIF server and return an input stream as the response.


web-services-app/pom.xml
```diff
@@ -161,6 +161,12 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock-jre8</artifactId>
+            <version>${wiremock.version}</version>
```
bbpennel, line 167: We don't need to specify the version here since it's already set in the main pom.xml

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Objects;
+
+/**
+ * @author snluong
```
bbpennel, line 17: Please add a description

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class DownloadImageService {
+    private String iiifBasePath;
+
+    public ResponseEntity<InputStreamResource> streamImage(String pidString, String size)
```
bbpennel, line 22: Generally we want to have javadoc for public methods, like this and getSize. We usually skip it for real getters and setters though.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
```diff
@@ -0,0 +1,67 @@
...
+                .contentType(MediaType.IMAGE_JPEG)
+                .body(resource);
+    }
+
+    /*
+    A method that builds the IIIF URL based on an assumption of full region, 0 rotation, and default quality.
+     */
+    private String buildURL(String id, String size) {
+        var formattedSize = size;
+        if (!Objects.equals(size, "full")) {
```
bbpennel, line 40: It'd be good to define a constant for `"full"` since it's used in a few places, and it needs to be the same between the service and the controller.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DownloadImageControllerIT.java
```diff
@@ -0,0 +1,144 @@
...
+
+        MvcResult result = mvc.perform(get("/downloadImage/" + pidString + "/full"))
+                .andExpect(status().is2xxSuccessful())
+                .andReturn();
+
+        // Verify content was retrieved
+        MockHttpServletResponse response = result.getResponse();
+
+        assertEquals("image/jpeg", response.getContentType());
+        assertEquals("attachment; filename=image_full.jpg", response.getHeader(CONTENT_DISPOSITION));
```
bbpennel, line 72: Is it possible to verify that the bunny image got returned as well? Or at least that a `InputStreamResource` was returned?

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DownloadImageControllerIT.java
```diff
@@ -0,0 +1,144 @@
...
+                .assertHasAccess(anyString(), eq(filePid), any(AccessGroupSetImpl.class), eq(viewOriginal));
+
+        MvcResult result = mvc.perform(get("/downloadImage/" + filePid.getId() + "/full"))
+                .andExpect(status().isForbidden())
+                .andReturn();
+
+        MockHttpServletResponse response = result.getResponse();
+        assertEquals("Insufficient permissions", response.getContentAsString());
+    }
+}
```
bbpennel, line 144: Another case worth checking could be `testGetImageAtPixelSizeBiggerThanFullNoPermission`

Also, what happens if we provide bad input like a non-numerical size, such as "large"? Does it throw an error? Or a blank size? Or a negative size? Some of those probably should throw an error.

What happens if the object doesn't have an original_file datastream? This would happen for any object type other than a File.

bbpennel (changes requested):
This looks good, I left some suggestions

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DownloadImageControllerIT.java
sharonluong: For the non-numerical size I will throw an illegal argument exception. For the case where there is no original file, how do you feel about a not found exception? @bbpennel

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
```diff
@@ -14,11 +14,17 @@
...
 /**
+ * Service to process access copy image downloads
  * @author snluong
  */
 public class DownloadImageService {
     private String iiifBasePath;
+    public static final String FULL_SIZE = "full";
+    public static final String INVALID_SIZE_MESSAGE = "Unable to determine size for access copy download";
 
+    /*
```
bbpennel, line 25: javadoc begins with `/**`, if you do that and hit enter, intellij will fill in all the boilerplate stuff for you

bbpennel (approved):

PR #1540, opened 2023-03-14
BXC-3981 - host in download event, and AnalyticsUtil cleanup

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3981

* Add dh property to supply missing host info.
* Removed unused value property.
*  Pass provided action down to the event being sent.
*  Make some constants static. 
* Add test coverage


web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/AnalyticsTrackerUtil.java
```diff
@@ -217,9 +222,6 @@ public void run() {
             if (label != null) {
                 params.add(new BasicNameValuePair("el", label));
             }
-            if (value != null) {
```
sharonluong, line 220: why was this no longer needed?

bbpennel: It was never used, so I figured I might as well remove the dead code

sharonluong (approved):

PR #1543, opened 2023-03-16
Draft: Aggregate updates

lfarrell:
https://jira.lib.unc.edu/browse/BXC-3911


static/js/vue-cdr-access/src/mixins/fullRecordUtils.js
```diff
@@ -87,6 +86,28 @@ export default {
...
+        formatDate(value) {
+            return format(value, 'yyyy-MM-dd');
+        },
+
+        editDescriptionUrl(id) {
+            return `${window.location.host}/describe/${id}`;
+        },
+
+        hasAccess(permission, user_type= 'authenticated') {
+            const group_roles = this.recordData.briefObject.groupRoleMap;
```
bbpennel, line 104: I thought the public UI wasn't returning the groupRole map? We can return it if we need to, or even produce an abbreviated version of it with the minimal permissions needed by the UI listed. We seem to be doing a bunch of these checks for patron permissions, so it seems like it might be necessary to return more than just whether there is edit description permission.

I would be happy to contribute some form of returning a reduced version of the role groups to this PR if that would be helpful

static/js/vue-cdr-access/tests/unit/aggregateRecord.spec.js
```diff
@@ -0,0 +1,447 @@
...
+        wrapper.vm.loadMetadata();
+
+        moxios.wait(() => {
+            expect(wrapper.vm.metadata).toEqual(html_data);
+            expect(wrapper.find('#mods_data_display').text()).toEqual(expect.stringContaining('Listen for real'));
+            done();
+        });
+
+        moxios.uninstall();
+    });
```
bbpennel, line 446: Might be good to test the behaviors of the download/request access buttons. Wouldn't hurt to add a test to verify that the finding aid field does and does not show depending on if a value is populated or not.

bbpennel (changes requested):
I haven't tried running it locally yet, but this looks good to me over all.

static/js/vue-cdr-access/src/mixins/fullRecordUtils.js
lfarrell: The group roles map does get return. What it doesn't do is inherit permissions, so if the required permission for a field is `viewMetadata` and the group has `viewOriginals` access they're going to be able to view that field.

static/js/vue-cdr-access/src/components/displayWrapper.vue
```diff
@@ -7,51 +7,55 @@ Top level component for full record pages with searching/browsing, including Adm
...
-                    <browse-search :object-type="container_metadata.type"></browse-search>
-                </div>
-                <div class="column is-2" v-if="showWidget">
-                    <browse-sort browse-type="display"></browse-sort>
-                </div>
-                <div class="column is-2 container-note" v-if="showWidget">
-                    <works-only></works-only>
+            <aggregate-record v-if="container_info.resourceType === 'Work'" :onyen="onyen" :record-data="container_info"></aggregate-record>
+
+            <div v-if="container_info.resourceType !== 'Work'">
```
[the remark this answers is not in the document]
lfarrell: Maybe we can make that a separate ticket. There's already a lot going on in this pr.

static/js/vue-cdr-access/src/mixins/fullRecordUtils.js
```diff
@@ -87,6 +89,35 @@ export default {
...
+
+        formatDate(value) {
+            return formatInTimeZone(value, 'America/New_York', 'yyyy-MM-dd');
+        },
+
+        editDescriptionUrl(id) {
+            return `https://${window.location.host}/admin/describe/${id}`;
+        },
+
+        hasGroupRole(permission, user_type= 'everyone') {
```
bbpennel, line 106: I'd suggest not naming the first parameter here `permission`, its confusing since there is `hasPermission` directly below it. How about `role` or `expected_role`?

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -271,7 +271,7 @@ public String getFullRecord(String pidString, Model model, HttpServletRequest re
      * @return
      */
     @GetMapping(path = "/{pid}/json", produces = APPLICATION_JSON_VALUE)
-    public @ResponseBody String handleRequest(@PathVariable("pid")String pidString) {
+    public @ResponseBody String handleRequest(@PathVariable("pid")String pidString, HttpServletResponse response) {
```
bbpennel, line 274:
```suggestion
    public @ResponseBody String handleRequest(@PathVariable("pid") String pidString, HttpServletResponse response) {
```

static/js/vue-cdr-access/tests/unit/aggregateRecord.spec.js
```diff
@@ -0,0 +1,735 @@
...
+        expect(wrapper.find('a.download').exists()).toBe(false);
+    });
+
+    it("allows users to download the work if a download link is present", async () => {
+        let updated_record = cloneDeep(record);
+        updated_record.briefObject.dataFileUrl = 'https://download-link.lib.unc.edu';
+        await wrapper.setProps({
+            recordData: updated_record
+        });
+        expect(wrapper.find('a.download').exists()).toBe(true);
```
bbpennel, line 625: it'd be good to verify that the download link goes to the correct url with a pid and stuff.

bbpennel (changes requested):

static/js/vue-cdr-access/src/components/full_record/aggregateRecord.vue
```diff
@@ -0,0 +1,179 @@
...
+                        <bread-crumbs :object-path="recordData.briefObject.objectPath">
+                        </bread-crumbs>
+                    </div>
+                </div>
+                <div class="columns">
+                    <div class="column">
+                        <h2>{{ recordData.briefObject.title }}</h2>
+                        <div class="columns columns-resize aggregate-info">
+                            <div class="column is-narrow-tablet" :class="isDeleted">
+                                <thumbnail :thumbnail-data="recordData"
```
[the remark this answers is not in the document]
lfarrell: I'm thinking this should probably be a separate ticket.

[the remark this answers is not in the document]
bbpennel: https://jira.lib.unc.edu/browse/BXC-4090

bbpennel (approved):

PR #1546, opened 2023-03-23
BXC-4025 full record abstract component

krwong:
[https://jira.lib.unc.edu/browse/BXC-4025](https://jira.lib.unc.edu/browse/BXC-4025)


static/js/vue-cdr-access/src/components/full_record/abstract.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+    <p v-if="truncateAbstract" class="abstract">{{ truncatedAbstractText }}... <a class="abstract-text" @click.prevent="toggleAbstractDisplay()"
+                                                                                  href="#">{{ abstractLinkText }}</a></p>
+    <p v-else class="abstract">{{ recordData.briefObject.abstractText }}</p>
+</template>
+
+<script>
```
bbpennel, line 7: I think you might need a `props: {` section so that this component can receive a briefObject. I think you can also safely just pass in `briefObject` and not the wrapping `recordData` if you don't actually need anything from outside briefObject.

static/js/vue-cdr-access/src/components/full_record/abstract.vue
```diff
@@ -0,0 +1,19 @@
...
+import fullRecordUtils from "../../mixins/fullRecordUtils";
```
bbpennel, line 8: I'm guessing this is just being included for `toggleAbstractDisplay`? If this component is the only place using that method now, I would suggest moving it into the `abstract.vue` component and dropping this import

static/js/vue-cdr-access/tests/unit/abstract.spec.js
```diff
@@ -0,0 +1,182 @@
...
+import { createRouter, createWebHistory } from 'vue-router';
+import abstract from '@/components/full_record/abstract.vue';
+import displayWrapper from '@/components/displayWrapper.vue';
+import {createI18n} from 'vue-i18n';
+import translations from '@/translations';
+import cloneDeep from 'lodash.clonedeep';
+
+const recordData = {
+    pageSubtitle: 'testAdminUnit',
+    briefObject: {
```
bbpennel, line 11: the structure of this object is out of date since we realize we were returning the briefObject in the wrong form, but i don't think it'll impact what you're doing too much. For your tests, all I think you need is:
```
{
    abstractText: 'Lorem ipsum dolor sit amet',
    id: '9e3b1754-8e39-494d-8cc4-44e0ddf264d9'
}
```

static/js/vue-cdr-access/tests/unit/abstract.spec.js
```diff
@@ -0,0 +1,182 @@
...
+
+        await wrapper.setProps({
+            recordData: longAbstract
+        });
+
+        let show_more = wrapper.find('.abstract-text');
+        expect(wrapper.vm.truncateAbstract).toBe(true);
+        expect(show_more.exists()).toBe(true);
+        expect(wrapper.vm.abstractLinkText).toEqual('Read more');
+    });
```
bbpennel, line 181: It would be good to also test that clicking the "Read more" link expands the abstract, and clicking the "Read less" button hides the extra.

bbpennel (changes requested):

static/js/vue-cdr-access/src/components/full_record/abstract.vue
krwong: It's included for `toggleAbstractDisplay`, `truncateAbstract`, `truncateAbstractText`, and `AbstractLinkText`

bbpennel (approved):

PR #1552, opened 2023-04-13
BXC-4066 - Prevent CDM Rights statements from getting indexed

bbpennel:
https://jira.lib.unc.edu/browse/BXC-4066

* Prevent CDM Rights statements from getting indexed
* Also split out the rights related assertions into their own tests so they are easier to detect issues with.


lfarrell (commented):
Should there be a test assertion to check "CDM rights"

lfarrell (approved):

PR #1556, opened 2023-04-24
Update search pages to use thumbnail component 

lfarrell:
- Update search pages to use thumbnail component 
- Move permissions into their own util file

https://jira.lib.unc.edu/browse/BXC-4090

### DIFF
--- a/static/js/vue-cdr-access/tests/unit/permissionUtils.spec.js
+++ b/static/js/vue-cdr-access/tests/unit/permissionUtils.spec.js
@@ -115,11 +115,16 @@ describe('permissionUtils', () => {
 
     it("checks for groups", () => {
         expect(wrapper.vm.hasGroups(recordData)).toBe(true);
+    });
 
+    it("checks for groups when none are defined", () => {
         let updatedRecord = cloneDeep(recordData);
         updatedRecord.briefObject.groupRoleMap = undefined;
         expect(wrapper.vm.hasGroups(updatedRecord)).toBe(false);
+    });
 
+    it("checks for groups when the group role map is empty", () => {
+        let updatedRecord = cloneDeep(recordData);
         updatedRecord.briefObject.groupRoleMap = {};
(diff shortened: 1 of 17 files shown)


static/js/vue-cdr-access/tests/unit/permissionUtils.spec.js
```diff
@@ -0,0 +1,147 @@
...
+                    RouterLink: RouterLinkStub
+                }
+            },
+            props: {
+                recordData: recordData
+            }
+        });
+    });
+
+    it("checks for groups", () => {
```
bbpennel, line 116: up to you, but if its not expensive to run more tests here it might help with readability/make expectations clearer to split this into 3 tests, 1 for each definition of the groupRoleMap. Same with the group roles and permissions tests.

static/js/vue-cdr-access/tests/unit/thumbnail.spec.js
```diff
@@ -189,9 +188,27 @@ describe('thumbnail.vue', () => {
...
+        await wrapper.setProps({ thumbnailData: updatedData });
         expect(wrapper.find('.fa-lock').exists()).toBe(false);
         expect(wrapper.find('.fa-trash').exists()).toBe(false);
+
+        // No lock icon if record has public access status
+        let updatedRecordData = cloneDeep(recordData);
+        updatedRecordData.briefObject.roleGroup = undefined;
+        updatedRecordData.briefObject.status.push('Public Access');
+        await wrapper.setProps({ thumbnailData: updatedRecordData });
+        expect(wrapper.find('.fa-lock').exists()).toBe(false);
```
bbpennel, line 211: Might be good to check again to make sure there's no trash icon

bbpennel (changes requested):
This looks great, much cleaner and I like that permissions utilities are their own mixin now. Just left a couple questions/suggestions

bbpennel (approved):

PR #1557, opened 2023-05-01
Add page views and event tracking for Google Analytics from Vue

lfarrell:
https://jira.lib.unc.edu/browse/BXC-4134

### DIFF
--- a/static/js/vue-cdr-access/src/mixins/gaUtils.js
+++ b/static/js/vue-cdr-access/src/mixins/gaUtils.js
@@ -1,0 +1,28 @@
+export default {
+    methods: {
+        pageEvent(recordData) {
+            let collection = recordData.briefObject.parentCollectionName || '';
+
+            if (collection === '' && recordData.briefObject.type === 'Collection') {
+                collection = recordData.briefObject.title;
+            }
+            if (collection === '') {
+                collection = '(no collection)';
+            }
+
+            this.$gtag.event('record', {
+                'event_category': recordData.briefObject.parentCollectionId,
+                'event_label': `${recordData.briefObject.title}|${recordData.briefObject.id}`,
+                'value': `${recordData.briefObject.title}|${recordData.briefObject.id}`
+            });
+        },
+
+        pageView(title) {
+            this.$gtag.pageview({
+                page_title: `Digital Collections Repository - ${title}`,
+                page_path: this.$route.path,
+                page_location: window.location.href
+            })
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


static/js/vue-cdr-access/src/mixins/gaUtils.js
```diff
@@ -0,0 +1,28 @@
...
+                collection = recordData.briefObject.title;
+            }
+            if (collection === '') {
+                collection = '(no collection)';
+            }
+
+            this.$gtag.event('record', {
+                'event_category': recordData.briefObject.parentCollectionId,
+                'event_label': `${recordData.briefObject.title}|${recordData.briefObject.id}`,
+                'value': `${recordData.briefObject.title}|${recordData.briefObject.id}`
```
bbpennel, line 16: Is the `value` needed here? Its being assigned the same as the even_label, and my understanding is that the `value` is intended for numerical values https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide

static/js/vue-cdr-access/src/mixins/gaUtils.js
bbpennel, line 1: I wonder if we should name this `analyticsUtils.js` so they we could use the same file as the API for triggering both ga and matomo events?

static/js/vue-cdr-access/src/mixins/gaUtils.js
bbpennel: Also, shouldn't the event have an action of "record"?

static/js/vue-cdr-access/tests/fixtures/displayWrapFixtures.js
bbpennel, line 1: glad to see this, but it should probably be named `displayWrapperFixtures` if its intended to go with the displayWrapper in particular

bbpennel (changes requested):
Looks good over all, just a few suggestions

static/js/vue-cdr-access/tests/fixtures/displayWrapFixtures.js
lfarrell: I'd intended to use it with the analyticsUtil tests as well, but it didn't really fit.

static/js/vue-cdr-access/src/mixins/gaUtils.js
lfarrell: The first parameter is the action, unless I'm reading this wrong, https://matteo-gabriele.gitbook.io/vue-gtag/methods/events

bbpennel: ah, i missed the first parameter, ignore that

bbpennel (approved):

PR #1563, opened 2023-05-05
Move breadcrumbs up a level

lfarrell:
- Move breadcrumbs up a level
- Fix display  of related items on smaller screens

### DIFF
--- a/static/js/vue-cdr-access/src/components/displayWrapper.vue
+++ b/static/js/vue-cdr-access/src/components/displayWrapper.vue
@@ -3,349 +3,355 @@ Top level component for full record pages with searching/browsing, including Adm
 -->
 <template>
     <header-small/>
-    <div>
-        <div v-if="is_page_loading" class="loading-icon">
-            <img :src="nonVueStaticImageUrl('ajax-loader-lg.gif')" alt="data loading icon">
-        </div>
-        <div v-if="!is_page_loading">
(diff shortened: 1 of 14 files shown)


static/js/vue-cdr-access/src/components/displayWrapper.vue
```diff
@@ -3,53 +3,58 @@ Top level component for full record pages with searching/browsing, including Adm
...
-                    <div class="column is-narrow-tablet" v-if="showWidget">
-                        <view-type></view-type>
-                    </div>
+        <admin-unit v-if="container_info.resourceType === 'AdminUnit'" :record-data="container_info"></admin-unit>
+        <collection-folder v-if="container_info.resourceType === 'Collection' || container_info.resourceType === 'Folder'"
+                           :record-data="container_info"></collection-folder>
+        <aggregate-record v-if="container_info.resourceType === 'Work'" :record-data="container_info"></aggregate-record>
+        <file-record v-if="container_info.resourceType === 'File'" :record-data="container_info"></file-record>
+
+        <div v-if="container_info.resourceType !== 'Work' && container_info.resourceType !== 'File'" style="background-color: white">
```
bbpennel, line 22: Should probably use `background-white` instead of inlining this

bbpennel (changes requested):
looks like you adjusted a bunch of the spacing around the breadcrumbs and the header to line things up better? seems like an improvement. The only weirdness is with the "Show only works" button where the text got pushed far away from the checkbox

bbpennel (approved):

PR #1564, opened 2023-05-08
Vue bug fixes

bbpennel:
* Ensure search results retrieved after brief object on page creation/mounting
* Fix issue where facets would be lost and not come back when navigating between admin units, collections and folders.
* Fix spacing around header in work/files page


lfarrell (changes requested):
Looks good overall. The way I had it working was a little weird. The only thing I notice is the gray header color is bleeding down too far.
Screen Shot 2023-05-09 at 8 35 50 AM

lfarrell (approved):

PR #1567, opened 2023-05-19
Add routing and display for not found, restricted content and 503 errors

lfarrell:
https://jira.lib.unc.edu/browse/BXC-4164

Depends on: https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/226

### DIFF
--- a/static/js/vue-cdr-access/src/components/collectionBrowseWrapper.vue
+++ b/static/js/vue-cdr-access/src/components/collectionBrowseWrapper.vue
@@ -3,7 +3,7 @@ Top level component used for the collection browse page
 -->
 <template>
     <header-small/>
-    <div>
+    <div v-if="!is_loading && !show_404 && !show_503">
         <div class="search-query-text collection-list has-text-centered">
             <h2>{{ $t('collecting_units') }}</h2>
         </div>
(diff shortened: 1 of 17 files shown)


static/js/vue-cdr-access/src/components/searchWrapper.vue
```diff
@@ -3,7 +3,7 @@ Top level component wrapper for search pages
 -->
 <template>
     <header-small/>
-    <div>
+    <div v-if="!show_404 && !show_503">
```
bbpennel, line 6: Alternatively we could put the error pages first in the template and make the actual page the else case, but i think its fine either way

static/js/vue-cdr-access/src/mixins/errorUtils.js
```diff
@@ -0,0 +1,31 @@
...
+    data() {
+        return {
+            show_404: false,
+            show_503: false,
+        }
+    },
+
+    methods: {
+        /**
+         * An empty response with a 200 return code signals that the uuid used in a query is valid, but not in the DCR.
```
bbpennel, line 13: Huh, this is really weird, I don't think this is intentional. I've created followup ticket to look into why we're getting 200 responses, my reading of the FullRecordController is that its supposed to either fail on the access control check or throw a NotFoundException, so maybe the error handling isn't working as expected
https://jira.lib.unc.edu/browse/BXC-4180

static/js/vue-cdr-access/src/components/collectionBrowseWrapper.vue
```diff
@@ -14,20 +14,28 @@ Top level component used for the collection browse page
             </div>
         </div>
     </div>
+    <not-found v-if="show_404" :display-header="false"></not-found>
```
bbpennel, line 17: You may have already looked at this, but I was wondering if it'd make sense to have the not-found and not-available modules up at the App.vue root level rather than in each submodule? Sort of like this example:
https://stackoverflow.com/a/60524363

lfarrell: It seems like we'd have to move the data retrieval queries up into the router since we don't necessarily know if there's an issue until we try to fetch the data.

static/js/vue-cdr-access/src/components/searchWrapper.vue
lfarrell: I'm kind of inclined to leave it as it is, as they're at least tucked out of the way a bit.

bbpennel (approved):

PR #1570, opened 2023-06-02
BXC-4189 - Add route to redirect list urls to record

bbpennel:
https://jira.lib.unc.edu/browse/BXC-4189

### DIFF
--- a/static/js/vue-cdr-access/src/router.js
+++ b/static/js/vue-cdr-access/src/router.js
@@ -31,7 +31,7 @@ const router = createRouter({
       }
     },
     { // Old style DCR list urls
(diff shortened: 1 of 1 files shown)


static/js/vue-cdr-access/src/router.js
```diff
@@ -30,6 +30,12 @@ const router = createRouter({
         return { path: to.path.replace('uuid:', '') }
       }
     },
+    { // Old style DCR list urls
+      path: `/list/:patchMatch(.*)*`,
```
cazzerson, line 34:
```suggestion
      path: `/list/:pathMatch(.*)*`,
```

cazzerson (approved):

PR #1575, opened 2023-06-09
BXC-4207 - Loris retry connections

bbpennel:
https://jira.lib.unc.edu/browse/BXC-4207

* Perform loris retry in a loop rather than recursively so that the connections will close before the next try starts, which should prevent the retries from getting stuck waiting for a connection to come available. 
* Set a shorter max time to wait for a connection to become available
* Removed a parameter that wasn't actually used, it previously just thought it was used because it was passed into a recursive call to the same method.

### DIFF
--- a/web-common/src/main/java/edu/unc/lib/boxc/web/common/services/LorisContentService.java
+++ b/web-common/src/main/java/edu/unc/lib/boxc/web/common/services/LorisContentService.java
@@ -61,6 +61,7 @@ public class LorisContentService {
 
         RequestConfig requestConfig = RequestConfig.custom()
                 .setConnectTimeout(2000)
+                .setConnectionRequestTimeout(5000)
                 .build();
 
         this.httpClient = HttpClients.custom()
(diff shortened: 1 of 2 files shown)


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/LorisContentService.java
```diff
@@ -80,42 +81,43 @@ public void getMetadata(String simplepid, String datastream, OutputStream outStr
...
-                }
-            } else {
-                if ((statusCode == 500 || statusCode == 404) && retryServerError > 0) {
-                    getMetadata(simplepid, datastream, outStream, response, retryServerError - 1);
-                } else {
-                    LOG.error("Unexpected failure: {}", httpResp.getStatusLine());
-                    LOG.error("Path was: {}", method.getURI());
+        int statusCode = -1;
+        String statusLine = null;
+        do {
```
sharonluong, line 86: i'm not familiar with do loops, how does it know when to stop the loop?

bbpennel: its the same as a while loop, except that it checks the condition at the end of the iteration rather than the beginning. See line 119.

sharonluong (approved):

PR #1576, opened 2023-06-12
BXC-4199 vuejs hide FA field if blank

agoslen:
Should the `no_finding_aid` line be removed from translation.js?


bbpennel:
Yes, it isn't used anywhere else so removing it would be good

bbpennel (approved):

PR #1581, opened 2023-06-20
Setup matomo with Vue

lfarrell:
https://jira.lib.unc.edu/browse/BXC-4148
Depends on: https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/230

### DIFF
--- a/static/js/vue-cdr-access/src/components/displayWrapper.vue
+++ b/static/js/vue-cdr-access/src/components/displayWrapper.vue
@@ -208,7 +208,11 @@ Top level component for full record pages with searching/browsing, including Adm
                return get(`${link}json`).then((response) => {
                    this.emptyJsonResponseCheck(response);
                    this.container_info = response.data;
-                   this.pageEvent(response.data);
+                   // Matomo
+                   this.matomoPageView(`Digital Collections Repository - ${this.container_info.pageSubtitle}`);
+                  // this.matomoPageEvent(this.container_info);
+
+                   // GA
                    this.pageView(this.container_info.pageSubtitle)
                    if (this.needsSearchResults) {
                        this.adjustFacetsForRetrieval();
@@ -216,7 +220,8 @@ Top level component for full record pages with searching/browsing, including Adm
                        this.is_page_loading = false;
                    }
                 }).catch(error => {
-                   this.setErrorResponse(error);
+                    console.log(error)
+                  // this.setErrorResponse(error);
                    this.is_page_loading = false;
                    console.log(error);
                });
--- a/static/js/vue-cdr-access/src/components/displayWrapper.vue
+++ b/static/js/vue-cdr-access/src/components/displayWrapper.vue
@@ -210,6 +210,8 @@ Top level component for full record pages with searching/browsing, including Adm
                    this.container_info = response.data;
 
                    this.pageView(this.container_info.pageSubtitle)
+                   this.pageEvent(response.data);
+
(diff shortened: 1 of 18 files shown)


static/js/vue-cdr-access/src/mixins/analyticsUtils.js
```diff
@@ -1,18 +1,27 @@
...
     methods: {
         pageEvent(recordData) {
-            let collection = recordData.briefObject.parentCollectionName || '';
+            this.$gtag.event('record', {
+                'event_category': recordData.briefObject.parentCollectionId,
+                'event_label': `${recordData.briefObject.title}|${recordData.briefObject.id}`
+            });
+        },
 
+        matomoPageEvent(recordData) {
```
bbpennel, line 10: What would you think of having pageEvent call matomoPageEvent rather than specifying the implementation everywhere we want to trigger an event? Same with having pageView call matomoPageView internally. Then we could probably roll back the updates to most of the other components when there are pageView and matomoPageView calls side by side.

static/js/vue-cdr-access/src/mixins/analyticsUtils.js
```diff
@@ -1,18 +1,27 @@
...
             if (collection === '') {
                 collection = '(no collection)';
             }
 
-            this.$gtag.event('record', {
-                'event_category': recordData.briefObject.parentCollectionId,
-                'event_label': `${recordData.briefObject.title}|${recordData.briefObject.id}`,
+            window._mtm = window._mtm || [];
+            window._mtm.push({
+                recordUUID: recordData.briefObject.id,
```
bbpennel, line 21: Is `recordUUID` something specified by matomo or are the keys up to us? If its up to us, I'd suggestion recordId instead, rather than specifying the algorithm used to produce the ID

static/js/vue-cdr-access/src/components/displayWrapper.vue
```diff
@@ -208,15 +208,20 @@ Top level component for full record pages with searching/browsing, including Adm
...
+
+                   // GA
                    this.pageView(this.container_info.pageSubtitle)
                    if (this.needsSearchResults) {
                        this.adjustFacetsForRetrieval();
                    } else {
                        this.is_page_loading = false;
                    }
                 }).catch(error => {
-                   this.setErrorResponse(error);
```
bbpennel, line 219: Is this a change we want to keep or was it commented out for debugging?

bbpennel (dismissed):
Looks good, left a few suggestions

static/js/vue-cdr-access/src/mixins/analyticsUtils.js
sharonluong: i chose the name, it's a variable i'm using in matomo which can be anything. we can change it to recordId, i just have to change it in the matomo config as well.

static/js/vue-cdr-access/src/components/displayWrapper.vue
lfarrell: I've uncommented it. There was an error from matomo that was being eaten by this.

static/js/vue-cdr-access/src/mixins/analyticsUtils.js
sharonluong: I changed it to `recordId` in the matomo configs for test, QA, and prod, so we should be good to change it here in the code. just need to redeploy to test after

PR #1583, opened 2023-06-26
BXC-4229 - Remove download button from the top of the work full record page

bbpennel:
https://jira.lib.unc.edu/browse/BXC-4229

* Remove download button from the top of the work full record page
* Separate embargo displaying logic from download button logic, since a work can be embargoed.

### DIFF
--- a/static/js/vue-cdr-access/src/components/full_record/restrictedContent.vue
+++ b/static/js/vue-cdr-access/src/components/full_record/restrictedContent.vue
@@ -12,20 +12,19 @@
         <div v-if="hasPermission(recordData, 'editDescription')" class="actionlink">
             <a class="edit button action" :href="editDescriptionUrl(recordData.briefObject.id)"><i class="fa fa-edit"></i> {{ $t('full_record.edit') }}</a>
         </div>
-        <template v-if="recordData.dataFileUrl">
+        <template v-if="recordData.resourceType === 'File'">
             <template v-if="hasPermission(recordData, 'viewOriginal')">
-                <file-download v-if="recordData.resourceType === 'File' || recordData.resourceType === 'Work'"
-                               :download-link="downloadLink"
+                <file-download :download-link="downloadLink"
                                :brief-object="recordData.briefObject"></file-download>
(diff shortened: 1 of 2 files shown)


static/js/vue-cdr-access/src/components/full_record/restrictedContent.vue
```diff
@@ -12,20 +12,19 @@
         <div v-if="hasPermission(recordData, 'editDescription')" class="actionlink">
             <a class="edit button action" :href="editDescriptionUrl(recordData.briefObject.id)"><i class="fa fa-edit"></i> {{ $t('full_record.edit') }}</a>
         </div>
-        <template v-if="recordData.dataFileUrl">
+        <template v-if="recordData.resourceType === 'File'">
             <template v-if="hasPermission(recordData, 'viewOriginal')">
-                <file-download v-if="recordData.resourceType === 'File' || recordData.resourceType === 'Work'"
-                               :download-link="downloadLink"
+                <file-download :download-link="downloadLink"
```
sharonluong, line 17: can you explain how this hides the button? it looks like it's still there?

bbpennel: Previously the `if` statement was allowing Files and Works in, now its only allowing Files in. The result is the works can still get some of the other functionality that the restrictedContent component provides (like telling the user if there are restrictions), but they will not get a Download button for works. I moved resourceType check up two lines

sharonluong: ahh ok, so right now this code is only for the works page, therefore it will not show up?

bbpennel: the`restricted-content` component is used on collection, work and file pages at the moment. Only file pages will get the download link though because of the `recordData.resourceType === 'File'` check. It might be that we could remove the download links from restrictedContent, but it might be tricky since the button has to be in a grouping component with stuff generated by this component.

lfarrell (approved):

sharonluong (approved):

PR #1593, opened 2023-07-18
Add download dropdown lists to  dataTables

lfarrell:
https://jira.lib.unc.edu/browse/BXC-4011


static/js/vue-cdr-access/src/components/full_record/fileList.vue
```diff
@@ -129,19 +134,21 @@ export default {
                 { type: 'file-size', targets: 3 },
                 {
                     render: (data, type, row) => {
+                        this.brief_object = row;
+                        console.log(JSON.stringify(this.brief_object))
```
bbpennel, line 138: Can remove

static/js/vue-cdr-access/src/components/full_record/fileList.vue
```diff
@@ -129,19 +134,21 @@ export default {
                 { type: 'file-size', targets: 3 },
                 {
                     render: (data, type, row) => {
+                        this.brief_object = row;
```
bbpennel, line 137: I'm not sure I understand this change. It seems like we are storing the current row while iterating through the list of file records to a component wide variable. Is this just so the row doesn't have to be passed as a parameter into various helper methods, or is there another purpose? If not, I would prefer to stick with passing it around, since I worry there might be unexpected side effects of relying on an instance variable instead of using the loop variable, and it seems like a mismatch of scope

static/js/vue-cdr-access/src/components/full_record/fileList.vue
```diff
@@ -240,9 +247,79 @@ export default {
...
+        },
+
+        closeDropdownLists() {
+            document.querySelectorAll('.show-list').forEach(element => {
+                element.setAttribute('aria-hidden', 'true');
+                element.classList.remove('show-list');
+            });
+        },
+
+        downloadButtonHtml() {
```
bbpennel, line 270: I am guessing this had to be duplicated because the html of the datatable is dynamically generated, so there wasn't a place to add in the fileDownload.vue component? It looks like we do have access to the current vue component from inside of the datatable though, so we may be able to cut down on duplication.

One option would be to replace fileDownload.vue with the fileDownloadUtils mixin, using this method (receiving the brief_object as a param) to generate the html instead of a template everywhere. Then I think we could call downloadButtonHtml where ever we need that element. Preferably the mixin could include the dropdown closing logic too.

It might also be possible to dynamically add another component via generated html being added to the page, either by some programmatic means of getting vue to import a component at a particular place in the html, or maybe even just including the `<fileDownload></fileDownload>` tag in the generated HTML if its going to be added to the page inside of the area controlled by vue. My impression is that vuejs is supposed to evaluate insertions to the HTML in these scenarios, but I'm not totally sure.

static/js/vue-cdr-access/src/mixins/fileDownloadUtils.js
```diff
@@ -0,0 +1,43 @@
...
+        },
+
+        largestImageEdge() {
+            const file_info = this.getOriginalFile.split('|');
+            const edge_sizes = file_info[file_info.length - 1].split('x');
+            return edge_sizes[0] > edge_sizes[1] ? edge_sizes[0] : edge_sizes[1];
+        },
+
+        showImageDownload() {
+            return this.hasPermission(this.brief_object, 'viewAccessCopies') &&
```
bbpennel, line 29: I think this should be viewOriginal since for the time being we are disallowing all the download options unless the user can access the original (which is why https://jira.lib.unc.edu/browse/BXC-4220 has been on hold while RCMT figures out what behavior we want)

static/js/vue-cdr-access/src/components/full_record/fileList.vue
```diff
@@ -240,9 +247,79 @@ export default {
...
+                   html += `<a href="${this.imgDownloadLink('800')}" class="dropdown-item">${this.$t('full_record.small') } JPG (800px)</a>`;
+                }
+                if (this.validSizeOption(1600)) {
+                    html += `<a href="${this.imgDownloadLink('1600')}" class="dropdown-item">${this.$t('full_record.medium') } JPG (1600px)</a>`;
+                }
+                if (this.validSizeOption(2500)) {
+                    html += `<a href="${this.imgDownloadLink('2500')}" class="dropdown-item">${this.$t('full_record.large') } JPG (2500px)</a>`;
+                }
+
+                if (this.viewOriginal) {
```
bbpennel, line 295: viewOriginal permission should be based off the permissions of the file object rather that of the work. I'm not really sure if fileList.vue needs to receive the work objects viewOriginal?

bbpennel (changes requested):
Looks good over all, left some suggestions and questions

static/js/vue-cdr-access/src/components/full_record/fileList.vue
lfarrell: It was so it didn't have to be passed around.

static/js/vue-cdr-access/src/components/full_record/fileList.vue
bbpennel: What about using `downloadButtonHtml` instead of fileDownload.vue everywhere? So that we would just have one implementation of it.

bbpennel (approved):

PR #1597, opened 2023-07-19
Bxc 4225 access copy download filename

sharonluong:
https://jira.lib.unc.edu/browse/BXC-4225

This PR makes the downloaded filename clearer, with the original filename and the size included.

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
@@ -94,6 +94,12 @@ public class DownloadImageService {
         return size;
     }
 
+    /**
+     * Formats the original filename to include size for access download filename
+     * @param record solr record of the file
+     * @param size validated size string from getSize
+     * @return a filename for the download like "filename_full.jpg" or "filename_800px.jpg
+     */
     public String getFilename(ContentObjectRecord record, String size) {
         var formattedSize = Objects.equals(size, FULL_SIZE) ?  FULL_SIZE : size + "px";
 
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
@@ -4,6 +4,8 @@ import static edu.unc.lib.boxc.model.fcrepo.ids.RepositoryPaths.idToPath;
 
 import edu.unc.lib.boxc.model.api.DatastreamType;
 import edu.unc.lib.boxc.search.api.models.ContentObjectRecord;
+import edu.unc.lib.boxc.search.api.models.Datastream;
+import org.apache.commons.io.FilenameUtils;
 import org.springframework.core.io.InputStreamResource;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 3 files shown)


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
```diff
@@ -30,15 +32,15 @@ public class DownloadImageService {
      * @return a response entity which contains headers and content of the access copy image
      * @throws IOException
      */
-    public ResponseEntity<InputStreamResource> streamImage(String pidString, String size)
+    public ResponseEntity<InputStreamResource> streamImage(String pidString, String size, String filename)
```
bbpennel, line 35: Just a thought, but if we changed the method signature to 
```
public ResponseEntity<InputStreamResource> streamImage(ContentObjectRecord contentObjectRecord, String size)
```
we could call getFilename directly inside this method since that seems like an internal consideration (and previously was since it used to be responsible for coming up with the filename), and it would have a more consistent API with getSize (using the ContentObjectRecord instead of a string).

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
```diff
@@ -93,6 +94,26 @@ public String getSize(ContentObjectRecord contentObjectRecord, String size) {
         return size;
     }
 
+    /**
+     * Formats the original filename to include size for access download filename
+     * @param record solr record of the file
```
bbpennel, line 99: update param name

bbpennel (changes requested):

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
sharonluong: I do need the `pidString` for the `buildUrl` method but i took your suggestion to move the filename functionality to the service instead of here in the controller.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
sharonluong: fixed!

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
```diff
@@ -27,17 +27,19 @@ public class DownloadImageService {
...
      * @param size a string which is either "full" for full size or a pixel length like "1200"
+     * @param pidString the UUID of the file
      * @return a response entity which contains headers and content of the access copy image
      * @throws IOException
      */
-    public ResponseEntity<InputStreamResource> streamImage(String pidString, String size, String filename)
+    public ResponseEntity<InputStreamResource> streamImage(ContentObjectRecord contentObjectRecord, String size, String pidString)
             throws IOException {
 
         String url = buildURL(pidString, size);
```
bbpennel, line 39: get the id from contentObjectRecord instead of passing in pidString on its own.

sharonluong: ok!

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/DownloadImageService.java
sharonluong: updated the code to figure out pidString fron the ContentObjectRecord

bbpennel (approved):

PR #1605, opened 2023-09-06
Bxc 4062 assign thumbnail api

sharonluong:
This PR introduces an API endpoint to assign a thumbnail from a child object to the parent object. To do this, I moved the existing thumbnail-related controller to the `modify` package and generalized the naming so I could add the new endpoint. There are also related tests and context files.

https://unclibrary.atlassian.net/browse/BXC-4062

### DIFF
--- a/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/thumbnail/ThumbnailRequestSender.java
+++ b/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/thumbnail/ThumbnailRequestSender.java
@@ -1,0 +1,30 @@
+package edu.unc.lib.boxc.operations.jms.thumbnail;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import edu.unc.lib.boxc.operations.jms.MessageSender;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * Service for sending requests to update child object used as thumbnail
+ *
+ * @author snluong
+ */
+public class ThumbnailRequestSender extends MessageSender {
+    private static final Logger log = LoggerFactory.getLogger(ThumbnailRequestSender.class);
+    private static final ObjectWriter MAPPER = new ObjectMapper().writerFor(ThumbnailRequest.class);
+
+    /**
+     * Send a ThumbnailRequest to the configured JMS queue
+     * @param request
+     * @throws IOException
+     */
+    public void sendToQueue(ThumbnailRequest request) throws IOException {
+        String messageBody = MAPPER.writeValueAsString(request);
+        sendMessage(messageBody);
+        log.info("Job to update thumbnail has been queued for {}", request.getAgent().getUsername());
+    }
+}
--- a/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/thumbnail/ThumbnailRequestSender.java
+++ b/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/thumbnail/ThumbnailRequestSender.java
@@ -25,6 +25,7 @@ public class ThumbnailRequestSender extends MessageSender {
     public void sendToQueue(ThumbnailRequest request) throws IOException {
         String messageBody = MAPPER.writeValueAsString(request);
         sendMessage(messageBody);
-        log.info("Job to update thumbnail has been queued for {}", request.getAgent().getUsername());
+        log.info("Job to update thumbnail has been queued for {} with file {}",
+                request.getAgent().getUsername(), request.getFilePidString());
     }
 }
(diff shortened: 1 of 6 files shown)


operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/thumbnail/ThumbnailRequestSender.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Send a ThumbnailRequest to the configured JMS queue
+     * @param request
+     * @throws IOException
+     */
+    public void sendToQueue(ThumbnailRequest request) throws IOException {
+        String messageBody = MAPPER.writeValueAsString(request);
+        sendMessage(messageBody);
+        log.info("Job to update thumbnail has been queued for {}", request.getAgent().getUsername());
```
bbpennel, line 28: Might be nice to include the id of the item in the request too

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
```diff
@@ -0,0 +1,114 @@
...
+    @Autowired
+    private ThumbnailRequestSender thumbnailRequestSender;
+    @Autowired
+    private AccessControlService aclService;
+    @Autowired
+    private RepositoryObjectLoader repositoryObjectLoader;
+
+    @PostMapping(value = "edit/displayThumbnail/{pid}")
+    public @ResponseBody
+    ResponseEntity<Object> ImportThumbnail(@PathVariable("pid") String pid,
```
bbpennel, line 59: I know this isn't new to this pr, but we should change this to importThumbnail since java conventions are for method names to start with the lowercase, which I missed in the original PR.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        result.put("destination", pid);
+
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @PutMapping(value = "/assignThumbnail/{pidString}")
+    @ResponseBody
+    public ResponseEntity<Object> AssignThumbnail(@PathVariable("pidString") String pidString) {
```
bbpennel, line 89: we should rename this to `assignThumbnail`.

Also, future us would probably find it helpful to add javadoc to this method to describe what the difference between it and the displayThumbnail method

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
```diff
@@ -0,0 +1,114 @@
...
+        PID pid = PIDs.get(pidString);
+
+        AccessGroupSet principals = getAgentPrincipals().getPrincipals();
+        aclService.assertHasAccess("Insufficient permissions to download access copy for " + pidString,
```
bbpennel, line 93: we'll want to update the message here to match the action that's being performed

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        var agent = AgentPrincipalsImpl.createFromThread();
+        var request = new ThumbnailRequest();
+        request.setAgent(agent);
+        request.setFilePidString(pidString);
+        try {
+            thumbnailRequestSender.sendToQueue(request);
+        } catch (IOException e) {
+            log.error("Error assigning file {} as thumbnail: {}", request.getFilePidString(), e);
```
bbpennel, line 109:
```suggestion
            log.error("Error assigning file {} as thumbnail", request.getFilePidString(), e);
```
when an exception is provided as the last parameter, the logger automatically formats it into message as the message + stacktrace.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
```diff
@@ -0,0 +1,114 @@
...
+            result.put("error", e.getMessage());
+            return new ResponseEntity<>(result, HttpStatus.UNSUPPORTED_MEDIA_TYPE);
+        }
+
+        result.put("destination", pid);
+
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @PutMapping(value = "/assignThumbnail/{pidString}")
```
bbpennel, line 87:
```suggestion
    @PutMapping(value = "edit/assignThumbnail/{pidString}")
```
we're using the edit prefix on almost all the paths for controllers that modify data, so we should stick with it here

bbpennel (changes requested):
This looks good! Just a few suggestions

bbpennel (approved):

PR #1609, opened 2023-09-08
Bxc 4063 clear thumbnail api

sharonluong:
This PR adds a controller endpoint to delete the assigned thumbnail for a work, along with some tests. I also made the Request object more flexible by letting it accept a general PID instead of a File PID.

https://unclibrary.atlassian.net/browse/BXC-4063

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
@@ -11,6 +11,7 @@ import edu.unc.lib.boxc.auth.api.services.AccessControlService;
 import edu.unc.lib.boxc.model.api.ids.PID;
 import edu.unc.lib.boxc.model.api.objects.FileObject;
 import edu.unc.lib.boxc.model.api.objects.RepositoryObjectLoader;
+import edu.unc.lib.boxc.model.api.objects.WorkObject;
 import edu.unc.lib.boxc.model.fcrepo.ids.PIDs;
 import edu.unc.lib.boxc.operations.jms.thumbnail.ThumbnailRequest;
 import edu.unc.lib.boxc.operations.jms.thumbnail.ThumbnailRequestSender;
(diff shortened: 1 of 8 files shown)


web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailIT.java
```diff
@@ -209,6 +211,48 @@ public void assignThumbnailPidIsNotAFile() throws Exception {
...
+        var workPidString = pid.getId();
+        var work = repositoryObjectFactory.createWorkObject(pid, null);
+        when(repositoryObjectLoader.getRepositoryObject(pid)).thenReturn(work);
+        mvc.perform(delete("/edit/deleteThumbnail/" + workPidString))
+                .andExpect(status().is2xxSuccessful())
+                .andReturn();
+
+        verify(thumbnailRequestSender).sendToQueue(requestCaptor.capture());
+        ThumbnailRequest request = requestCaptor.getValue();
+        assertEquals(workPidString, request.getPidString());
```
bbpennel, line 226: should probably verify that the right action was set too

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
```diff
@@ -112,13 +114,42 @@ public ResponseEntity<Object> assignThumbnail(@PathVariable("pidString") String
...
+    @ResponseBody
+    public ResponseEntity<Object> deleteThumbnail(@PathVariable("pidString") String pidString) {
+        PID pid = PIDs.get(pidString);
+
+        AccessGroupSet principals = getAgentPrincipals().getPrincipals();
+        aclService.assertHasAccess("Insufficient permissions to assign thumbnail for " + pidString,
+                pid, principals, Permission.editDescription);
+
+        var object = repositoryObjectLoader.getRepositoryObject(pid);
+        if (!(object instanceof WorkObject)) {
```
bbpennel, line 137: I'm not sure if this matches up with the functionality in the processor. It seems like that is expecting the id to be the file rather than the work. We could make it so that the user can pass in either the work or the file object's id when deleting the thumbnail. Or update the processor to expect a work idea for delete requests. Or make the processor flexible about what it receives.

bbpennel (changes requested):

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ThumbnailController.java
sharonluong: Thanks for catching that. I ended up changing it so that it expects the File ID in all cases.

bbpennel (approved):

PR #1611, opened 2023-09-25
Bxc 4082 thumbnail solr index

sharonluong:
This PR allows the assigned thumbnail to be indexed in solr for a work. If the content object is a work object, it looks to see if there is a thumbnail object relation. if there is, then it selectively adds the thumbnail streams of the assigned file object to the work object's datastreams.

https://unclibrary.atlassian.net/browse/BXC-4082

### DIFF
--- a/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetContentStatusFilter.java
+++ b/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetContentStatusFilter.java
@@ -53,6 +53,9 @@ public class SetContentStatusFilter implements IndexDocumentFilter{
             if (parentResc.hasProperty(Cdr.primaryObject, resc)) {
                 status.add(FacetConstants.IS_PRIMARY_OBJECT);
             }
+            if (parentResc.hasProperty(Cdr.useAsThumbnail, resc)) {
+                status.add(FacetConstants.IS_ASSIGNED_THUMBNAIL);
+            }
         }
 
         return status;
@@ -69,6 +72,12 @@ public class SetContentStatusFilter implements IndexDocumentFilter{
             status.add(FacetConstants.MEMBERS_ARE_ORDERED);
         } else {
             status.add(FacetConstants.MEMBERS_ARE_UNORDERED);
+        }
+
+        if (resource.hasProperty(Cdr.useAsThumbnail)) {
+            status.add(FacetConstants.THUMBNAIL_ASSIGNED);
+        } else {
+            status.add(FacetConstants.NO_THUMBNAIL_ASSIGNED);
         }
     }
 }
--- a/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
+++ b/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
@@ -234,14 +234,7 @@ public class SetDatastreamFilter implements IndexDocumentFilter {
             }
 
             String owner = (ownedByOtherObject ? pid.getId() : null);
-            String name = type.getId();
-            String mimetype = type.getMimetype();
-            String extension = type.getExtension();
-            File file = deriv.getFile();
-            Long filesize = file.length();
-            String filename = file.getName();
-            var derivative = new DatastreamImpl(owner, name, filesize, mimetype, filename, extension, null, null);
-            addDerivativeToList(dsList, derivative);
+            dsList.add(createDatastream(deriv, owner));
         });
     }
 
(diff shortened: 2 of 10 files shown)


indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilterTest.java
```diff
@@ -315,6 +316,62 @@ public void workObjectWithoutPrimaryObjectTest() throws Exception {
...
+        filter.filter(dip);
+
+        assertNotNull(idb.getDatastream());
+        assertNull(idb.getFilesizeSort());
+        assertNotNull(idb.getFilesizeTotal());
+        assertThumbnailDatastreams(fileId);
+    }
+
+    @Test
+    public void workObjectTestWithPrimaryAndThumbnailObjects() throws Exception {
```
bbpennel, line 341:
```suggestion
    public void workObjectWithPrimaryAndThumbnailObjectsTest() throws Exception {
```
This test does raise a good question, what happens if the primary object has thumbnails (which is not set up in this test or the other primary object test at the moment, but is implemented) and there is a thumbnail object set? 

I'd say that we would only want thumbnails from the thumbnail object. So for the time being, maybe when adding thumbnail datastreams we should first delete any existing thumbnail datastreams from the list. Primary objects are semi-defunct at this point, but i don't think we're ready to totally remove them yet.

model-fcrepo/src/main/java/edu/unc/lib/boxc/model/fcrepo/objects/WorkObjectImpl.java
```diff
@@ -214,4 +202,26 @@ public List<PID> getMemberOrder() {
                 .map(PIDs::get)
                 .collect(Collectors.toList());
     }
+
+    @Override
+    public FileObject getThumbnailObject() {
+        return getObject(Cdr.useAsThumbnail);
+    }
+
+    private FileObject getObject(Property property) {
```
bbpennel, line 211: how about `getFileObjectByProperty` or relation? `getObject` is very generic in java (everything in java is an Object, except for a few things like ints)

services-camel-app/src/main/webapp/WEB-INF/solr-indexing-context.xml
```diff
@@ -80,6 +80,7 @@
             <list>
                 <ref bean="setRecordDatesFilter" />
                 <ref bean="setDatastreamFilter" />
+                <ref bean="setContentStatusFilter" />
```
bbpennel, line 83: Adding this is good, but according to the ticket we still need to update SetContentStatusFilter to add a new "THUMBNAIL_ASSIGNED" status. I suspect we should also add a "IS_THUMBNAIL" property to the file itself, just like how IS_PRIMARY_OBJECT is getting set right now, since I think it would make it easier for users to tell what item is the thumbnail in the admin ui.

model-fcrepo/src/main/java/edu/unc/lib/boxc/model/fcrepo/objects/WorkObjectImpl.java
```diff
@@ -214,4 +202,26 @@ public List<PID> getMemberOrder() {
                 .map(PIDs::get)
                 .collect(Collectors.toList());
     }
+
+    @Override
+    public FileObject getThumbnailObject() {
```
bbpennel, line 207: Please add a unit test for this new method to WorkObjectTest

bbpennel (changes requested):
This looks good so far, there were a couple other things I think we needed to cover

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
```diff
@@ -227,41 +229,56 @@ private void addDerivatives(List<Datastream> dsList, PID pid, boolean ownedByOth
...
+            if (THUMBNAIL_DS_TYPES.contains(type)) {
+                datastreams.remove(datastream);
+            }
+        }
+        return datastreams;
+    }
 
-        dsList.add(new DatastreamImpl(owner, name, filesize, mimetype, filename, extension, null, null));
+    private void addDerivativeToList(List<Datastream> dsList, DatastreamImpl derivative) {
+        dsList.add(derivative);
```
bbpennel, line 281: I think it'd make sense to remove this helper method at this point since it is just wrapping a `.add` call

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetContentStatusFilter.java
```diff
@@ -53,6 +53,9 @@ private List<String> determineContentStatus(DocumentIndexingPackage dip)
             if (parentResc.hasProperty(Cdr.primaryObject, resc)) {
                 status.add(FacetConstants.IS_PRIMARY_OBJECT);
             }
+            if (parentResc.hasProperty(Cdr.useAsThumbnail, resc)) {
```
bbpennel, line 56: It'd be good to add tests to `SetContentStatusFilterTest` for the new status values being added

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
```diff
@@ -227,41 +229,56 @@ private void addDerivatives(List<Datastream> dsList, PID pid, boolean ownedByOth
...
-        Long filesize = file.length();
-        String filename = file.getName();
+    /**
+     *  There may be thumbnail streams from the primary object, so we'll clear those
+     *  before adding the assigned thumbnail datastreams
+     *
+     * @param datastreams full list of datastreams to index for the work object
+     * @return modified list of datastreams without thumbnail datastreams
+     */
+    private List<Datastream> clearPreviousThumbnailDatastreams(List<Datastream> datastreams) {
```
bbpennel, line 270: We should update `SetDatastreamFilterTest.workObjectTestWithPrimaryAndThumbnailObjects` to verify that this function works by giving the primaryObject thumbnail datastreams

bbpennel (changes requested):
Looks good, mostly a few tests cases we need to add

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
sharonluong: I'll rework the methods to streamline it a bit more

bbpennel (approved):

PR #1612, opened 2023-10-05
Bxc 4067 thumbnail retrieval

sharonluong:
This PR fixes some spelling discrepancies, matching with the complementary one in boxc-ansible: https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/237

It also adds in configs in necessary context files to initialize things.
https://unclibrary.atlassian.net/browse/BXC-4067


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -6,6 +6,7 @@
 import edu.unc.lib.boxc.model.api.DatastreamType;
 import edu.unc.lib.boxc.model.api.ResourceType;
 import edu.unc.lib.boxc.model.api.ids.PID;
+import edu.unc.lib.boxc.model.api.objects.WorkObject;
```
bbpennel, line 9: Seems like we can roll back the changes to this file, since it's just adding an unused import

bbpennel (changes requested):
Looks good, one minor cleanup thing

bbpennel (approved):

PR #1622, opened 2023-11-06
BXC-3284 add controller and service files

sharonluong:
This PR adds a controller and service file for the image server proxy (which will currently be Cantaloupe) but hopefully in the future we can switch it to any other server.

https://unclibrary.atlassian.net/browse/BXC-3284


bbpennel (changes requested):
Looks good over all, just some cleanup

PR #1626, opened 2023-11-22
Don't show download button to users without download access

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4318


bbpennel (approved):
Looks good!

PR #1630, opened 2023-11-29
Only show thumbnails if user has viewAccessCopies permissions

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4315


web-common/src/main/java/edu/unc/lib/boxc/web/common/services/PermissionsHelper.java
```diff
@@ -107,6 +109,16 @@ public boolean hasDatastreamAccess(AccessGroupSet principals, DatastreamType dat
         return accessControlService.hasAccess(metadata.getPid(), principals, permission);
     }
 
+    public boolean hasThumbnailPreviewAccess(AccessGroupSet principals, DatastreamType datastream,
```
bbpennel, line 112: If we want to keep this as public, we should add javadoc. Its only used internal to the permissions helper though at the moment, so it'd also be reasonable to make it private.

So this is roughly the same as hasDatastreamAccess but without the check to see if the datastream exists, right? In that case, I don't think we need to check for both the LARGE and SMALL thumbnail in hasThumbnailAccess, since they have the same permission and it doesn't verify if it exists. We could probably also simplify things a bit and remove the `datastream` input parameter, since internally we can figure out what the thumbnail datastream(s) are.

bbpennel (changes requested):
This looks good over all, just one part I had questions about

bbpennel (approved):

PR #1638, opened 2023-12-07
Create UI for single use links

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4363


static/js/vue-cdr-access/src/translations.js
```diff
@@ -42,12 +42,17 @@ export default {
             available_date: "Available after {available_date}",
             collection_id: "Archival Collection ID",
             contains: "Contains",
+            copied_link: "Download URL, ${text}, copied to clipboard",
+            copied_link_failed: "Unable to copy download URL, ${text}, to clipboard",
```
bbpennel, line 46: I don't think the `$`'s are necessary, it is displaying a $ in front of the url in the message

static/js/vue-cdr-access/src/components/full_record/restrictedContent.vue
```diff
@@ -20,6 +20,9 @@
                         <i class="fa fa-search" aria-hidden="true"></i> View</a>
                 </div>
             </template>
+            <template v-if="hasPermission(recordData, 'viewHidden')">
```
bbpennel, line 23: We should add tests to verify that the single-use-link component shows up with viewHidden permissions, and not without them

static/js/vue-cdr-access/tests/unit/singleUseLink.spec.js
```diff
@@ -0,0 +1,65 @@
...
+                writeText: jest.fn().mockImplementation(() => Promise.resolve()),
+            },
+        });
+
+        await wrapper.setData({ single_use_links: [response_date] });
+        await wrapper.find('.download-link-wrapper a').trigger('click');
+        expect(window.navigator.clipboard.writeText)
+            .toHaveBeenCalledWith(response_date.link);
+    });
+});
```
bbpennel, line 65: If it's not too much of a pain, it'd be good to also add a test for when it fails to generate a link since there's a decent amount of logic there.

PR #1639, opened 2023-12-07
BXC-4327 export member order

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4327](https://unclibrary.atlassian.net/browse/BXC-4327)

- add "export member order" option
- add "export" option and put "export csv" and "export member order" in it


static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -196,10 +196,15 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 		}
 
 		// Export actions
-		if (!isContentRoot) {
-			items["sepexport"] = "";
+		if ($.inArray('viewHidden', metadata.permissions) !== -1) {
+			items["export"] = {name: "Export", items: {}}
+
+			if (!isContentRoot && metadata.type !== 'File' && $.inArray('viewHidden', metadata.permissions) != -1) {
```
bbpennel, line 202: You don't need to check for the 'viewHidden' permission again here or in the following conditional since this whole block is wrapped in a check already. It actually looks like you could merge together this conditional and the Export Member Order one, then wrap Export CSV with `!isContentRoot`

static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -196,10 +196,15 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
...
+				items['export']['items']["exportCSV"] = {name: "Export CSV"};
+			}
+
 			if (metadata.type !== 'File' && $.inArray('viewHidden', metadata.permissions) != -1) {
-				items["exportCSV"] = {name : 'Export as CSV'};
+				items['export']['items']["exportMemberOrder"] = {name: "Export Member Order"};
```
bbpennel, line 207: the "Member Order" export doesn't appear to do anything for me when i click on it here. For these menu items to perform an action, they have to have a corresponding callback in this section:
https://github.com/UNC-Libraries/box-c/blob/main/static/js/admin/src/ResultObjectActionMenu.js#L247
It looks like the behavior for triggering the export is defined here:
https://github.com/UNC-Libraries/box-c/blob/44b2d5a3f6a6031d0d865cc65cac4d72f2cf5340/static/js/admin/src/ExportMenu.js#L71-L76
So you will like need to add something like this in the ResultObjectActionMenu callbacks. The main difference is that in ExportMenu it is being given a list of ids to export, while here it is being given a single id.

bbpennel (changes requested):

bbpennel (approved):

PR #1642, opened 2023-12-13
BXC-4376 - Virus scan for large files

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4376

Use streaming scanning for files which are larger than the max scanning size, so that the first N bytes will be scanned.

### DIFF
--- a/deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/VirusScanJob.java
+++ b/deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/VirusScanJob.java
@@ -42,7 +42,7 @@ public class VirusScanJob extends AbstractConcurrentDepositJob {
             .getLogger(VirusScanJob.class);
 
     private static final int MAX_RETRIES = 5;
-    private long maxStreamSize = 0;
+    private long maxStreamSize;
 
     private ClamAVClient clamClient;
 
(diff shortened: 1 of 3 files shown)


deposit-app/src/main/webapp/WEB-INF/deposit-jobs-context.xml
```diff
@@ -191,6 +191,7 @@
         <property name="clamClient" ref="clamClient" />
         <property name="maxQueuedJobs" value="${job.fileValidation.maxQueuedJobs:5}" />
         <property name="executorService" ref="fileValidationExecutor" />
+        <property name="maxStreamSize" value="${clamd.maxStreamSize:64000000}" />
```
sharonluong, line 194: what is this for?

deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/VirusScanJob.java
```diff
@@ -42,6 +42,7 @@ public class VirusScanJob extends AbstractConcurrentDepositJob {
             .getLogger(VirusScanJob.class);
 
     private static final int MAX_RETRIES = 5;
+    private long maxStreamSize = 0;
```
sharonluong, line 45: is 0 the default?

sharonluong (approved):
just a few questions

deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/VirusScanJob.java
bbpennel: Yes, which would mean it'd send everything to streaming

deposit-app/src/main/webapp/WEB-INF/deposit-jobs-context.xml
bbpennel: that's what sets the maxStreamSize. I copy and pasted its value from the clam client. It pulls from the clamd.maxStreamSize variable is set, and falls back to 64mb if not

PR #1644, opened 2023-12-19
Bxc-4358 manage links service

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4358

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
@@ -33,8 +33,8 @@ public class SingleUseKeyService {
      */
     public String generate(String id) {
         var key = getKey();
-        var expirationInMilliseconds = System.currentTimeMillis() + DAY_MILLISECONDS;
         lock.lock();
+        var expirationInMilliseconds = System.currentTimeMillis() + DAY_MILLISECONDS;
         try (var csvPrinter = createCsvPrinter(CSV_HEADERS, csvPath)) {
             csvPrinter.printRecord(id, key, expirationInMilliseconds);
         } catch (Exception e) {
@@ -49,16 +49,19 @@ public class SingleUseKeyService {
      * Determines if a key is valid by seeing if it is in the CSV and if the expiration timestamp has not passed
      * @param key access key for single use link
      * @return true if key is in the CSV, otherwise false
-     * @throws IOException
      */
-    public boolean keyIsValid(String key) throws IOException {
-        var csvRecords = parseCsv(CSV_HEADERS, csvPath);
-        var currentMilliseconds = System.currentTimeMillis();
-        for (CSVRecord record : csvRecords) {
-            if (key.equals(record.get(ACCESS_KEY))) {
-                var expirationTimestamp = Long.parseLong(record.get(TIMESTAMP));
-                return currentMilliseconds <= expirationTimestamp;
+    public boolean keyIsValid(String key) {
+        try {
+            var csvRecords = parseCsv(CSV_HEADERS, csvPath);
+            var currentMilliseconds = System.currentTimeMillis();
+            for (CSVRecord record : csvRecords) {
+                if (key.equals(record.get(ACCESS_KEY))) {
+                    var expirationTimestamp = Long.parseLong(record.get(TIMESTAMP));
+                    return currentMilliseconds <= expirationTimestamp;
+                }
             }
+        } catch (IOException e) {
+            throw new RepositoryException("Failed to determine if key is valid in Single Use Key CSV", e);
         }
         return false;
     }
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
@@ -19,7 +19,7 @@ import static edu.unc.lib.boxc.web.services.utils.CsvUtil.parseCsv;
 public class SingleUseKeyService {
     public static final String ID = "UUID";
     public static final String ACCESS_KEY = "Access Key";
-    private static final String TIMESTAMP = "Timestamp";
+    private static final String TIMESTAMP = "Expiration Timestamp";
     public static final String[] CSV_HEADERS = new String[] {ID, ACCESS_KEY, TIMESTAMP};
     public static final long DAY_MILLISECONDS = 86400000;
     private Path csvPath;
@@ -29,12 +29,12 @@ public class SingleUseKeyService {
      * @param id UUID of the record
      * @return generated access key
      */
-    public String generate(String id) {
+    public String generate(String id, long expirationInMilliseconds) {
         var lock = new ReentrantLock();
         var key = getKey();
         lock.lock();
         try (var csvPrinter = createCsvPrinter(CSV_HEADERS, csvPath)) {
-            csvPrinter.printRecord(id, key, System.currentTimeMillis());
+            csvPrinter.printRecord(id, key, expirationInMilliseconds);
         } catch (Exception e) {
             throw new RepositoryException("Failed to write new key to Single Use Key CSV", e);
         } finally {
@@ -54,9 +54,8 @@ public class SingleUseKeyService {
         var csvRecords = parseCsv(CSV_HEADERS, csvPath);
         for (CSVRecord record : csvRecords) {
             if (accessKeyMatchesUuid(record, id, key)) {
-                var startTime = Long.parseLong(record.get(TIMESTAMP));
-                long endTime = startTime + DAY_MILLISECONDS;
-                return currentMilliseconds >= startTime && currentMilliseconds <= endTime;
+                var expirationTimestamp = Long.parseLong(record.get(TIMESTAMP));
+                return currentMilliseconds <= expirationTimestamp;
             }
         }
         return false;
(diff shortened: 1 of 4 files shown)


web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+    @Test
+    public void testKeyIsNotValidCurrentTimeIsMoreThan24hLater() throws IOException {
+        var key = SingleUseKeyService.getKey();
+        generateDefaultCsv(key);
+        var currentMilliseconds = System.currentTimeMillis() + (3 * DAY_MILLISECONDS);
+        assertFalse(singleUseKeyService.keyIsValid(UUID_TEST, key, currentMilliseconds));
+    }
+
+    @Test
+    public void testInvalidate() throws IOException {
```
bbpennel, line 90: It would be good to add a test for when the key does not exist in the csv file to define what that behavior should be. Having nothing happen (current behavior?) is fine, since I think both of the cases we're going to use invalidate will have already checked that the key existed right before invalidating it, but its still useful to define the expected outcome in a test. Also, we could skip overwriting to the file if nothing was removed.

bbpennel (changes requested):

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Determines if a key is valid by seeing if it is in the CSV, connected to the proper ID
+     * @param id uuid of the box-c record
+     * @param key access key for single use link
+     * @return true if key is in the CSV, otherwise false
+     * @throws IOException
+     */
+    public boolean keyIsValid(String id, String key, long currentMilliseconds) throws IOException {
```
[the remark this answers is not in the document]
sharonluong: Ah, i wasn't sure if we should check to make sure the key matched the ID of the file being downloaded. @lfarrell and I thought it wouldn't hurt so I added it. I'll remove it.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Determines if a key is valid by seeing if it is in the CSV and if the expiration timestamp has not passed
+     * @param key access key for single use link
+     * @return true if key is in the CSV, otherwise false
+     * @throws IOException
+     */
+    public boolean keyIsValid(String key) throws IOException {
+        var csvRecords = parseCsv(CSV_HEADERS, csvPath);
```
bbpennel, line 55: since the other major public methods in this class all throw RepositoryException when an IOException is thrown, it would be consistent to do the same on this line and remove the throws from this method

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyServiceTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    public void testGenerateNoCsvExists() throws IOException {
+        var key = singleUseKeyService.generate(UUID_TEST);
+        var newRecords = parseCsv(CSV_HEADERS, csvPath);
+        var currentMilliseconds = System.currentTimeMillis();
+        assertCsvContainsCorrectEntry(newRecords, UUID_TEST, key, currentMilliseconds);
+    }
+
+    @Test
+    public void testGenerateMultipleCallsForDifferentIds() throws IOException {
```
bbpennel, line 53: I ran this test and inspected the CSV file it generates, and it looks like CsvUtil needs an adjustment to prevent it from adding headers for already existing files. At the moment the CSV from this test is:
```
UUID,Access Key,Expiration Timestamp
f277bb38-272c-471c-a28a-9887a1328a1f,fa2aba6ed5aa4718af1d497778cfaf792593d477356f3,1704575328754
UUID,Access Key,Expiration Timestamp
ba70a1ee-fa7c-437f-a979-cc8b16599652,b60c1430637f46d8b765dc2885d219d52593d4a01cbc7,1704575328775
UUID,Access Key,Expiration Timestamp
83c2d7f8-2e6b-4f0b-ab7e-7397969c0682,44fc9960d7434cbbb307f504f10f00f92593d4a0c11df,1704575328776
```

To make sure we don't end up in a situation like this later, we should also add an assertion on the number of records in the CSV.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/utils/CsvUtil.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Creates a CSV printer and determines whether new rows should be appended to existing CSV
+     * @param headers header values of the CSV
+     * @param csvPath path of the CSV
+     * @return
+     * @throws IOException
+     */
+    public static CSVPrinter createCsvPrinter(String[] headers, Path csvPath) throws IOException {
+        var file = csvPath.toFile();
+        if (file.exists()) {
```
bbpennel, line 38: We don't need to convert to a file here, we can test `Files.exists(csvPath)`

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyServiceTest.java
```diff
@@ -0,0 +1,159 @@
...
+        for (CSVRecord record : csvRecords) {
+            if (value.equals(record.get(column))) {
+                fail("Entry found!");
+                return;
+            }
+        }
+    }
+
+    private void assertCsvContainsCorrectEntry(List<CSVRecord> csvRecords,
+                                               String id, String key, long currentTimestamp) {
```
bbpennel, line 135: I think it'd be okay to drop `currentTimestamp` from this method signature, since it's always being passed `System.currentTimeMillis()`, and the expiration delay is much longer than the amount of time that would pass between each assertion.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyServiceTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    public void testKeyIsNotValidCurrentTimeIsMoreThan24hLater() throws IOException {
+        var key = SingleUseKeyService.getKey();
+        var pastTimestamp = System.currentTimeMillis() - (2 * DAY_MILLISECONDS);
+        generateDefaultCsv(key, pastTimestamp);
+        assertFalse(singleUseKeyService.keyIsValid(key));
+    }
+
+    @Test
+    public void testInvalidate() throws IOException {
```
bbpennel, line 104: We should add assertions about the number of entries into all the invalidate tests too, and also add a test that invalidates 2 keys in a row to make sure it doesn't run into any weirdness with the headers.

bbpennel (changes requested):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyServiceTest.java
```diff
@@ -131,15 +148,11 @@ private void assertDoesNotContainValue(List<CSVRecord> csvRecords, String column
...
-    private void assertCsvContainsCorrectEntry(List<CSVRecord> csvRecords,
-                                               String id, String key, long currentTimestamp) {
+    private void assertCsvContainsCorrectEntry(List<CSVRecord> csvRecords, String id, String key) {
         for (CSVRecord record : csvRecords) {
             if (key.equals(record.get(SingleUseKeyService.ACCESS_KEY))) {
                 assertEquals(id, record.get(SingleUseKeyService.ID));
-                // timestamp must not be null and must be in the future
-                var expirationTimestamp = record.get(SingleUseKeyService.TIMESTAMP);
-                assertNotNull(expirationTimestamp);
-                assertTrue(Long.parseLong(expirationTimestamp) > currentTimestamp);
```
bbpennel: Sorry i didn't mean to remove the check about the expiration timestamp being in the future, i just meant that you could set currentTimestamp inside of this method rather than passing it in from outside

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/utils/CsvUtil.java
```diff
@@ -34,11 +34,10 @@ public static List<CSVRecord> parseCsv(String[] headers, Path csvPath) throws IO
      * @throws IOException
      */
     public static CSVPrinter createCsvPrinter(String[] headers, Path csvPath) throws IOException {
-        var file = csvPath.toFile();
-        if (file.exists()) {
+        if (Files.exists(csvPath)) {
             var writer = Files.newBufferedWriter(csvPath, StandardOpenOption.APPEND);
-            return new CSVPrinter(writer, CSVFormat.DEFAULT
-                    .withHeader(headers));
+            //return new CSVPrinter(writer, CSVFormat.DEFAULT.withHeader(headers));
```
bbpennel, line 39: Can remove this commented out line

bbpennel (approved):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyServiceTest.java
sharonluong: i thought that was weird but just went with it =D

PR #1646, opened 2023-12-22
BXC-4353 - catalog id link

agoslen:
- added template for identifiers
- for catalog id identifiers, generate link

### DIFF
--- a/web-access-app/src/main/resources/recordTransformations/modsToFullRecord.xsl
+++ b/web-access-app/src/main/resources/recordTransformations/modsToFullRecord.xsl
@@ -301,14 +301,7 @@
 								<xsl:value-of select="$defaultLabel"/>
 							</xsl:when>
 							<xsl:otherwise>
-								<xsl:choose>
-									<xsl:when test="$groupKey = 'doi' or $groupKey = 'issn'">
-										<xsl:value-of select="upper-case($groupKey)"/>
-									</xsl:when>
-									<xsl:otherwise>
-										<xsl:value-of select="concat(upper-case(substring($groupKey,1,1)), substring($groupKey,2))"/>
-									</xsl:otherwise>
-								</xsl:choose>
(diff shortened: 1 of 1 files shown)


web-access-app/src/main/resources/recordTransformations/modsToFullRecord.xsl
```diff
@@ -354,27 +347,59 @@
...
 						</xsl:choose>
 					</th><xsl:value-of select="$newline"/>
 					<td>
 						<xsl:for-each select="current-group()">
-							<xsl:value-of select="text()"/><br/><xsl:value-of select="$newline"/>
+							<xsl:choose>
+								<xsl:when test="$groupKey = 'Catalog ID'">
+									<xsl:variable name="catalog_number">
+										<xsl:choose>
+											<xsl:when test="contains(., 'catalog.lib.unc.edu/')">
```
bbpennel, line 380: Does this field ever have URLs other than catalog.lib.unc.edu? I was just wondering if it'd make sense to leave any URLs in this field untouched rather than reducing them down to an ID and then formatting it back into a URL. Then the variable would be "catalog_url" instead, and the `otherwise` case would build a URL instead of a number. We don't necessary need to change this, just wondering about it.

agoslen: Hmm...not in practice but I guess in theory someone could enter a URL for the "classic catalog", for example https://afton.lib.unc.edu/record=b4386236

agoslen: @bbpennel added an alternative like what you are describing, let me know what you think

bbpennel: That looks good, but i wonder if it'd work to just check for the value beginning with "http" rather than looking for specific hosts?

agoslen: Sure!

bbpennel (approved):

PR #1653, opened 2024-01-11
BXC-4216 - Patron Role/Permission for accessing reduced quality images

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4316

* Introduces canViewReducedQuality patron role and associated canViewReducedResImages permission
    * Adds new property URI for the role
* For file record pages, if the user can access any download options (low res or original) then the download button will appear instead of the restricted content block.
* Download dropdown will now display for images if the user has viewReducedResImages permission, but will not show full size or original download options
* DownloadImage controller now uses new permission for enforcement when downloading images below the max size
* Refactoring of UserRole object to reduce repetition of permissions, remove unused methods, and add test coverage
* Removed a number of unused permission related methods
* Disables a codeclimate check that was requiring enums to be named like constants, since we have a lot of existing enums that don't follow that convention.


static/js/vue-cdr-access/src/components/full_record/restrictedContent.vue
```diff
@@ -13,12 +13,14 @@
...
         </div>
         <template v-if="recordData.resourceType === 'File'">
-            <template v-if="hasPermission(recordData, 'viewOriginal')">
+            <template v-if="hasDownloadAccess(recordData)">
                 <div class="header-button" v-html="downloadButtonHtml(recordData.briefObject)"></div>
-                <div class="actionlink">
-                    <a class="button view action" :href="recordData.dataFileUrl">
-                        <i class="fa fa-search" aria-hidden="true"></i> View</a>
-                </div>
+                <template v-if="hasPermission(recordData, 'viewOriginal')">
```
lfarrell, line 18: You should be able to put the `v-if` on `<div class="actionlink">`. I'm not sure why I did it this way, but it's not a big deal either way.

lfarrell (approved):

PR #1656, opened 2024-01-19
Bxc 4362 invalidate schedule

sharonluong:
Adding cron job and code changes to invalidate expired single use keys on a schedule (currently the first minute of every hour)
https://unclibrary.atlassian.net/browse/BXC-4362

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
@@ -86,17 +86,17 @@ public class SingleUseKeyService {
         try {
             var csvRecords = parseCsv(CSV_HEADERS, csvPath);
             var updatedRecords = new ArrayList<>();
-            var recordExists = false;
+            var recordsChanged = false;
             for (CSVRecord record : csvRecords) {
                 if (recordShouldBeInvalidated(key, record)) {
-                    recordExists = true;
+                    recordsChanged = true;
                 } else {
                     // keep this record as it is valid
                     updatedRecords.add(record);
                 }
             }
 
-            if (recordExists) {
+            if (recordsChanged) {
                 writeNewCsv(updatedRecords);
             }
         } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
```diff
@@ -70,37 +78,67 @@ public boolean keyIsValid(String key) {
...
+     * Invalidates a single use key CSV record by removing it from the CSV
+     * @param key access key of the box-c record, may be null for time-based invalidation
      */
     public void invalidate(String key) {
         lock.lock();
         try {
             var csvRecords = parseCsv(CSV_HEADERS, csvPath);
             var updatedRecords = new ArrayList<>();
-            var keyExists = false;
+            var recordExists = false;
```
bbpennel, line 89: I find the variable a bit confusing for what its used for (recordExists means we deleted a record), maybe `recordsRemoved` or `recordsChanged` might be clearer, particularly down in the conditional to decide whether to write out the CSV

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
```diff
@@ -70,37 +78,67 @@ public boolean keyIsValid(String key) {
...
     }
 
+    /**
+     * Determines if the specific record should be invalidated
+     * @param key the access key, may be null if we want a time-based eval
+     * @param record the csv record
+     * @return true if the record should be invalidated
+     */
+    private boolean recordShouldBeInvalidated(String key, CSVRecord record) {
+        var currentTime = System.currentTimeMillis();
```
bbpennel, line 116: Minor, but we can move this into the conditional to make it clearer that the time is only needed when key is null

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
```diff
@@ -70,37 +78,67 @@ public boolean keyIsValid(String key) {
...
         }
     }
 
+    /**
+     * Determines if the specific record should be invalidated
+     * @param key the access key, may be null if we want a time-based eval
+     * @param record the csv record
+     * @return true if the record should be invalidated
+     */
+    private boolean recordShouldBeInvalidated(String key, CSVRecord record) {
```
bbpennel, line 115: Huh, codeclimate is angry about some variable name here. Best I can guess, it doesn't like "record"? Apparently its reserved in java 14+ (which we're not using yet), but is only reserved in some contexts (seems to be for declaring classes). I guess we might as well rename it, even though we use "record" plenty of places...?
https://www.baeldung.com/java-record-keyword

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyServiceTest.java
```diff
@@ -124,6 +124,16 @@ public void testInvalidateWhenKeyIsNotPresent() throws IOException {
         assertEquals(3, records.size());
     }
 
+    @Test
+    public void testInvalidateTimeBased() throws IOException {
```
bbpennel, line 128: It'd be good to add a test that does an invalid based on time, but doesn't delete anything (or at least leaves one item left) so we can demonstrate it doesn't just delete everything.

bbpennel (changes requested):
Looks good, just a few minor suggestions

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/SingleUseKeyService.java
sharonluong: good point, thanks!

bbpennel (approved):

PR #1659, opened 2024-01-22
Clear single use links if the url path change

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4423


sharonluong (approved):
nifty fix!

PR #1662, opened 2024-01-24
BXC-4437 - Hide download button when no options available

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4437

 * Hide the download button if the user doesn't have any download options, such as if it is a very small image and they only have reduced quality access
 * A few extra tests to confirm that the download button and render partial option lists.


lfarrell (approved):

lfarrell:
In the rare case where there are no download options, but the user has download access a blank column will display in the file list. It's the file download column, just with no button.

PR #1663, opened 2024-01-25
WIP: Basic UI setup for setting page view display

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4432

### DIFF
--- a/static/js/admin/src/PagedDisplayForm.js
+++ b/static/js/admin/src/PagedDisplayForm.js
@@ -1,0 +1,53 @@
+define('PagedDisplayForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteStateChangeMonitor', 'tpl!../templates/admin/pagedDisplayForm',
+        'ModalLoadingOverlay', 'AbstractForm', 'AlertHandler'],
+    function($, ui, _, RemoteStateChangeMonitor, pagedDisplayForm, ModalLoadingOverlay, AbstractForm) {
+
+        var defaultOptions = {
+            title : 'Edit View Settings',
+            createFormTemplate : pagedDisplayForm,
+            submitMethod: 'PUT'
+        };
+
+        function PagedDisplayForm(options) {
+            this.options = $.extend({}, defaultOptions, options);
+        }
+
+        PagedDisplayForm.prototype.constructor = PagedDisplayForm;
+        PagedDisplayForm.prototype = Object.create(AbstractForm.prototype);
+
+        PagedDisplayForm.prototype.preprocessForm = function(resultObject) {
+            var newViewSetting = $("#view_settings_change", this.$form).val();
+            var pid = resultObject.metadata.id;
+            this.action_url = "/services/api/edit/view_settings/" + pid + "?direction=" + encodeURIComponent(newViewSetting);
+        };
+
+        PagedDisplayForm.prototype.validationErrors = function() {
+            var errors = [];
+            var viewSetting = $("#view_settings_change", this.$form).val();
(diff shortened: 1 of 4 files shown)


static/templates/admin/pagedDisplayForm.html
```diff
@@ -0,0 +1,20 @@
+<iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
+<form id="edit_view_settings" class="edit_form edit_view_settings_form">
+    <div class="form_field">
+        <label for="view_settings_change">Page Display:</label>
+        <select id="view_settings_change">
```
bbpennel, line 5: I think viewBehavior and direction should be separate inputs since they are independent. So we should have a view behavior input with values "individual" and "paged" (and maybe "continous"). Direction doesn't necessarily have to be addressed in this ticket, I didn't include it in any of the currently created tickets since it was lower priority. I'm pretty sure direction affects "individual" display in UV as well though.

static/js/admin/src/action/UpdatePagedDisplayBatchAction.js
```diff
@@ -0,0 +1,68 @@
...
+    function UpdatePagedDisplayBatchAction(context) {
+        this._create(context);
+    }
+
+    UpdatePagedDisplayBatchAction.prototype.constructor = UpdatePagedDisplayBatchAction;
+    UpdatePagedDisplayBatchAction.prototype = Object.create(AbstractBatchAction.prototype);
+
+    UpdatePagedDisplayBatchAction.prototype.isValidTarget = function (target) {
+        return target.isSelected() && target.isEnabled()
+            && $.inArray('changePatronAccess', target.metadata.permissions) !== -1
```
bbpennel, line 11: should be `editViewSettings` permission

static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -195,6 +195,10 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 			items["ingestSourceFilesOnly"] = {name : 'Add Files from Server'};
 		}
 
+		if (!isContentRoot && metadata.type === 'Work' && $.inArray('editViewSettings', metadata.permissions) !== -1) {
```
bbpennel, line 198: I don't think the `!isContentRoot` is necessary here, since a work can't be the contentRoot, although I see that we're doing the same thing in the `if` statement right before this, i'm not sure why.

static/js/admin/src/PagedDisplayForm.js
```diff
@@ -0,0 +1,53 @@
...
+            this.options = $.extend({}, defaultOptions, options);
+        }
+
+        PagedDisplayForm.prototype.constructor = PagedDisplayForm;
+        PagedDisplayForm.prototype = Object.create(AbstractForm.prototype);
+
+        PagedDisplayForm.prototype.preprocessForm = function() {
+            let newViewSetting = $('#view_settings_change', this.$form).val();
+            let pids = $('#paged_display_targets', this.$form).val();
+            this.action_url = `/services/api/edit/view_settings?targets=${encodeURIComponent(pids)}&direction=${encodeURIComponent(newViewSetting)}`;
```
bbpennel, line 21: related to my comment below, there should be a `behavior` parameter. We might want to leave out direction for now since none of the other tickets include implementing it, but we can add it in easily enough later.

bbpennel (changes requested):

bbpennel (approved):

PR #1668, opened 2024-02-06
BXC-3390 - Skip MODS updates when content is unmodified

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-3390

* VersionedDatastreamService now supports flag to tell it to skip updates if the SHA1 of the new version and existing version are the same
* UpdateDescriptionService uses this flag to prevent unnecessary updates when the incoming data is unchanged
* Adds utility for digest calculation on a stream
* Adds test coverage for MultiDigestInputStreamWrapper

### DIFF
--- a/persistence/src/test/java/edu/unc/lib/boxc/persist/impl/MultiDigestInputStreamWrapperTest.java
+++ b/persistence/src/test/java/edu/unc/lib/boxc/persist/impl/MultiDigestInputStreamWrapperTest.java
@@ -1,0 +1,79 @@
+package edu.unc.lib.boxc.persist.impl;
+
+import edu.unc.lib.boxc.persist.api.DigestAlgorithm;
+import edu.unc.lib.boxc.persist.api.exceptions.InvalidChecksumException;
+import org.apache.commons.io.IOUtils;
+import org.junit.jupiter.api.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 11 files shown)


persistence/src/test/java/edu/unc/lib/boxc/persist/impl/MultiDigestInputStreamWrapperTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private static String CONTENT = "Something to digest";
+
+    private static String MD5 = "7afbf05666feeebe7fbbf1c9071584e6";
+    private static URI MD5_URI = URI.create("urn:md5:" + MD5);
+
+    private static String SHA1 = "23d51c61a578a8cb00c5eec6b29c12b7da15c8de";
+    private static URI SHA1_URI = URI.create("urn:sha1:" + SHA1);
+
+    private static String SHA512 =
```
lfarrell, line 32: Lines 32-376 don't appear to be used. Are they needed?

lfarrell (changes requested):
I left one comment

persistence/src/test/java/edu/unc/lib/boxc/persist/impl/MultiDigestInputStreamWrapperTest.java
bbpennel: Ah yes, thanks, this test was partially copy and pasted from Fedora and I forgot to clean it up

lfarrell (approved):

PR #1672, opened 2024-02-08
Split folder and collections into their own components, and fix issues with display of collection names and ids

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4460


bbpennel (changes requested):

static/js/vue-cdr-access/src/components/full_record/folderRecord.vue
```diff
@@ -1,16 +1,77 @@
...
+                    {{ recordData.briefObject.title }}
+                    <span class="item_container_count">{{ displayChildCount }}</span>
+                </h2>
+                <p v-if="fieldExists(recordData.briefObject.added)">
+                    <strong>{{ $t('full_record.date_added') }}: </strong>
+                    {{ formatDate(recordData.briefObject.added) }}
+                </p>
+                <p class="parent_collection" v-if="fieldExists(recordData.briefObject.parentCollectionName)">
+                    <strong>{{ $t('full_record.collection') }}: </strong> <a :href="collectionLink">{{ recordData.briefObject.parentCollectionName }}</a>
+                </p>
```
bbpennel, line 16: I was looking at how this is implemented in the file and work record pages, and they are both like this:
```
<li v-if="fieldExists(recordData.briefObject.parentCollectionName)">
    <span class="has-text-weight-bold">{{ $t('display.collection') }}: </span>
    <router-link class="parent-collection" :to="parentUrl">{{ recordData.briefObject.parentCollectionName }}</router-link>
</li>
```
I was wondering if we should be using a router-link here too?

Also, it looks like there is already a `parentUrl()` method in `fullRecordUtil` that we can probably use here instead of adding a new collectionLink method (looks like it does the same thing, but with a leading slash).

And I guess it holds true for files, works and folders, but they should always have a parent collection, so we could probably leave off the fieldExists check. It would be a data integrity issue if they didn't.

static/js/vue-cdr-access/src/translations.js
```diff
@@ -41,7 +41,6 @@ export default {
             additional_metadata: "View Additional Metadata",
             available_date: "Available after {available_date}",
             collection: "Collection",
-            collection_id: "Archival Collection ID",
```
bbpennel, line 43: I assume you're going to need this back for https://unclibrary.atlassian.net/browse/BXC-4463 ? Not that you have to add it back in this PR.

bbpennel (changes requested):

bbpennel (approved):

PR #1682, opened 2024-02-20
BXC-4470 - De-Lorisify iiif v2 implementation

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4470

* Remove all Loris references, renaming to IIIF v2 theme
* Move iiif v2 endpoint from access app to services, since that is where v3 resides
* Change URL pattern for iiif v2 to match v3's
* Removes iiif v2 manifest integration test, which is mostly covered by a unit test now (test coverage is definitely not 100% for the v2 implementation, but wasn't before either)
* Remaps "full/full" jp2 region requests to "full/max" since cantaloupe iiif v2 doesn't appear to be able to handle "full/full" even though that is what is in the v2 iiif spec.
* iiif related property names now specify v2 or v3


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/ImageServerV2Service.java
```diff
@@ -117,20 +118,24 @@ public void getMetadata(String simplepid, OutputStream outStream,
...
-            String format, String datastream, OutputStream outStream, HttpServletResponse response,
+            String format, OutputStream outStream, HttpServletResponse response,
             int retryServerError) {
 
-        StringBuilder path = new StringBuilder(getLorisPath());
-        path.append(idToPath(simplepid, 4, 2)).append(simplepid).append(".jp2")
+        StringBuilder path = new StringBuilder(getImageServerProxyBasePath());
+        // Remap "full" size to "max" since cantaloupe's v2 api doesn't seem to be able to handle full/full correctly
+        if ("full".equals(size)) {
+            size = "max";
```
sharonluong, line 136: i think there is a constant for this in the image server util class

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DownloadImageControllerIT.java
```diff
@@ -14,7 +14,6 @@
 import edu.unc.lib.boxc.search.solr.models.ContentObjectSolrRecord;
 import edu.unc.lib.boxc.web.common.services.SolrQueryLayerService;
 import edu.unc.lib.boxc.web.services.processing.DownloadImageService;
-import edu.unc.lib.boxc.web.services.processing.ImageServerProxyService;
```
sharonluong, line 17: was this just an unused import?

sharonluong (changes requested):
just small things

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DownloadImageControllerIT.java
bbpennel: Yeah, it got cleaned up by intellij when I was moving things around

PR #1684, opened 2024-02-22
Bxc 4430 new get endpoint

sharonluong:
This PR adds an endpoint to get the view settings of a work object. Currently the only view setting available is `viewBehavior`
https://unclibrary.atlassian.net/browse/BXC-4430

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingController.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingController.java
@@ -3,17 +3,27 @@ package edu.unc.lib.boxc.web.services.rest.modify;
 import edu.unc.lib.boxc.auth.api.Permission;
 import edu.unc.lib.boxc.auth.api.models.AccessGroupSet;
 import edu.unc.lib.boxc.auth.api.services.AccessControlService;
+import edu.unc.lib.boxc.model.api.exceptions.InvalidOperationForObjectType;
 import edu.unc.lib.boxc.model.api.ids.PID;
+import edu.unc.lib.boxc.model.api.objects.RepositoryObjectLoader;
+import edu.unc.lib.boxc.model.api.objects.WorkObject;
+import edu.unc.lib.boxc.model.api.rdf.CdrView;
 import edu.unc.lib.boxc.model.fcrepo.ids.PIDs;
+import org.apache.jena.rdf.model.Property;
+import org.apache.jena.rdf.model.Resource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.CrossOrigin;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.ResponseBody;
+
+import java.util.HashMap;
+import java.util.Map;
 
 import static edu.unc.lib.boxc.auth.fcrepo.services.GroupsThreadStore.getAgentPrincipals;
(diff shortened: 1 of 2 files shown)


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingController.java
```diff
@@ -0,0 +1,68 @@
...
+
+@Controller
+public class ViewSettingController {
+    private static final Logger LOG = LoggerFactory.getLogger(ViewSettingController.class);
+    @Autowired
+    private AccessControlService accessControlService;
+    @Autowired
+    private RepositoryObjectLoader repositoryObjectLoader;
+
+    @CrossOrigin
```
bbpennel, line 39: We shouldn't need cross origin here, it's just going to be the DCR calling this endpoint, whereas with the iiif stuff we would expect it to be called by other websites

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingController.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static edu.unc.lib.boxc.auth.fcrepo.services.GroupsThreadStore.getAgentPrincipals;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@Controller
```
bbpennel, line 31: Could use some class level javadoc

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingIT.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @AfterEach
+    void closeService() throws Exception {
+        closeable.close();
+    }
+
+    @Test
+    public void testGetViewSetting() throws Exception {
+        var paged = ViewSettingRequest.ViewBehavior.PAGED.toString();
```
bbpennel, line 83: Just wanted to note that this will set the value of the property to "PAGED", while the UI code currently has "paged" as the form input values. It's not going to cause any issues in the tests, but we'll have to settle on a value at some point

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingController.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ViewSettingController.class);
+    @Autowired
+    private AccessControlService accessControlService;
+    @Autowired
+    private RepositoryObjectLoader repositoryObjectLoader;
+
+    @CrossOrigin
+    @GetMapping(value = "/edit/viewSettings/{id}", produces = APPLICATION_JSON_VALUE)
+    @ResponseBody
+    public ResponseEntity<Object> getViewSetting(@PathVariable("id") String id) {
```
bbpennel, line 42: should probably add javadoc for this public method too

bbpennel (changes requested):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingIT.java
sharonluong: Oh so should i just use the string `paged` in the tests?

sharonluong: nvm i just used a string

bbpennel (approved):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ViewSettingIT.java
bbpennel: at some point we could add an extra method to the enum to get the lower case form, but this is good for now

PR #1687, opened 2024-02-26
BXC-4488 fix null user id error and add test

sharonluong:
This PR generates a random UUID and sets the user id to that for analytics usage, to fix these errors in the log:

`An exception occurred while recording download event on http://dcr-bes.libint.unc.edu:8181/fcrepo/rest/content/5b/b8/10/4c/5bb8104c-c218-4ee6-bd0b-6d7a705f4f83
java.lang.NullPointerException: inputHex is marked non-null but is null
	at org.matomo.java.tracking.parameters.VisitorId.fromHex(VisitorId.java:87)
	at edu.unc.lib.boxc.web.common.utils.AnalyticsTrackerUtil.buildMatomoRequest(AnalyticsTrackerUtil.java:70)`

https://unclibrary.atlassian.net/browse/BXC-4488

### DIFF
--- a/web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/AnalyticsTrackerUtil.java
+++ b/web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/AnalyticsTrackerUtil.java
@@ -1,25 +1,23 @@
 package edu.unc.lib.boxc.web.common.utils;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URI;
-import java.util.UUID;
-
-import javax.servlet.http.Cookie;
-import javax.servlet.http.HttpServletRequest;
-
-import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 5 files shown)


web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/AnalyticsTrackerUtil.java
```diff
@@ -2,10 +2,13 @@
 
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
+import java.util.UUID;
 
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 
+import com.ctc.wstx.util.StringUtil;
```
bbpennel, line 10: I don't think this import is used

bbpennel (changes requested):

PR #1695, opened 2024-03-08
BXC-4497 - Temporarily disabling setting of a random user id

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4497


web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/AnalyticsTrackerUtil.java
```diff
@@ -167,7 +167,7 @@ public AnalyticsUserData(HttpServletRequest request) {
 
             // if it cannot be found in the cookie, generate a random 16 character hex user ID
             if (StringUtils.isBlank(uid)) {
```
sharonluong, line 169: maybe add a temporary comment as to why it's commented out?

sharonluong (approved):

PR #1696, opened 2024-03-11
BXC-4497 - Use single MatomoTracker instance

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4497

* Reuse a single instance of MatomoTracker
* Add init and close methods to AnalyticsTrackerUtil to create the MatomoTracker and shut it down
* Update to Matomo-java-tracker 3.3.0


sharonluong (approved):
thanks for looking into this and also improving the matomo tracker itself!

PR #1706, opened 2024-03-20
Bxc 4445 duracloud service

sharonluong:
This PR includes the processor, router, request, request serialization helper, and their associated tests for adding streaming properties to files. It also renames the ProcessorTestHelper since I added a method for the router tests to it (and updated the view setting router test, but did not update any other router tests as that's out of scope)
https://unclibrary.atlassian.net/browse/BXC-4445

### DIFF
--- a/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequest.java
+++ b/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequest.java
@@ -9,17 +9,23 @@ import java.util.Set;
 
 import static java.util.Arrays.asList;
 
+/**
+ * Request object for updating the streaming properties of a FileObject
+ */
 public class StreamingPropertiesRequest {
     public static final String DURACLOUD = "duracloud";
     public static final String OPEN = "open-hls";
     public static final String CLOSED = "closed-hls";
     public static final String CAMPUS = "campus-hls";
+    public static String ADD = "add";
+    public static String DELETE = "delete";
     public static Set<String> VALID_FOLDERS = new HashSet<>(asList(OPEN, CLOSED, CAMPUS));
     @JsonDeserialize(as = AgentPrincipalsImpl.class)
     private AgentPrincipals agent;
     private String filePidString;
     private String filename;
     private String folder;
+    private String action;
     private String host = DURACLOUD;
 
 
@@ -62,5 +68,13 @@ public class StreamingPropertiesRequest {
 
     public void setHost(String host) {
         this.host = host;
+    }
+
+    public String getAction() {
+        return action;
+    }
+
+    public void setAction(String action) {
+        this.action = action;
     }
 }
--- a/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequest.java
+++ b/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequest.java
@@ -1,0 +1,66 @@
+package edu.unc.lib.boxc.operations.jms.streaming;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import edu.unc.lib.boxc.auth.api.models.AgentPrincipals;
+import edu.unc.lib.boxc.auth.fcrepo.models.AgentPrincipalsImpl;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.util.Arrays.asList;
+
+public class StreamingPropertiesRequest {
+    private static final String DURACLOUD = "duracloud";
+    public static final String OPEN = "open-hls";
+    public static final String CLOSED = "closed-hls";
+    public static final String CAMPUS = "campus-hls";
+    public static Set<String> VALID_FOLDERS = new HashSet<>(asList(OPEN, CLOSED, CAMPUS));
+    @JsonDeserialize(as = AgentPrincipalsImpl.class)
+    private AgentPrincipals agent;
+    private String filePidString;
+    private String filename;
+    private String folder;
+    private String host = DURACLOUD;
+
+
+
+    public AgentPrincipals getAgent() {
+        return agent;
+    }
+
+    public void setAgent(AgentPrincipals agent) {
+        this.agent = agent;
+    }
+
+    public String getFilePidString() {
+        return filePidString;
+    }
+
+    public void setFilePidString(String filePidString) {
+        this.filePidString = filePidString;
+    }
+
+    public String getFilename() {
+        return filename;
+    }
+
+    public void setFilename(String filename) {
+        this.filename = filename;
+    }
+
+    public String getFolder() {
+        return folder;
+    }
+
+    public void setFolder(String folder) {
+        this.folder = folder;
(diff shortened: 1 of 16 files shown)


operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequest.java
```diff
@@ -0,0 +1,66 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import edu.unc.lib.boxc.auth.api.models.AgentPrincipals;
+import edu.unc.lib.boxc.auth.fcrepo.models.AgentPrincipalsImpl;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.util.Arrays.asList;
+
+public class StreamingPropertiesRequest {
```
bbpennel, line 12: javadoc

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+import edu.unc.lib.boxc.model.api.ids.PID;
+import edu.unc.lib.boxc.model.api.objects.RepositoryObjectLoader;
+import edu.unc.lib.boxc.model.api.rdf.Cdr;
+import edu.unc.lib.boxc.model.api.services.RepositoryObjectFactory;
+import edu.unc.lib.boxc.model.fcrepo.ids.PIDs;
+import edu.unc.lib.boxc.operations.jms.streaming.StreamingPropertiesRequest;
+import edu.unc.lib.boxc.operations.jms.streaming.StreamingPropertiesRequestSerializationHelper;
+import org.apache.camel.Exchange;
+import org.apache.camel.Processor;
+import org.apache.tika.utils.StringUtils;
```
bbpennel, line 15: we should use the apache commons StringUtils rather than tika, to avoid introducing tika dependencies in unexpected places.

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+        if (isValid(request, pid)) {
+            var file = repositoryObjectLoader.getFileObject(pid);
+            repositoryObjectFactory.createExclusiveRelationship(
+                    file, Cdr.streamingHost, request.getHost());
+            repositoryObjectFactory.createExclusiveRelationship(
+                    file, Cdr.streamingFile, formatFilename(request.getFilename()));
+            repositoryObjectFactory.createExclusiveRelationship(
+                    file, Cdr.streamingFolder, request.getFolder());
+        } else {
+            throw new IllegalArgumentException("Both a filename and streaming folder are required.");
```
bbpennel, line 50: alternatively, `isValid` could be `assertValid`, which would throw exceptions if something was invalid rather than returning true or false. That would allow the error message to be more specific (at the moment invalid object type and invalid folder values would be thrown with messages indicating that filename and folder are required)

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+        return true;
+    }
+
+    /**
+     * Converts a filename to appropriate format for streaming
+     * A filename like a/b/banjo_recording.mp3 would transform to banjo_recording-playlist.m3u8
+     * @param filename filename of the file object
+     * @return formatted string
+     */
+    private String formatFilename(String filename) {
```
bbpennel, line 78: i don't have a good sense of what the existing workflow for setting streaming properties is like. Will the user setting these properties only know about the "banjo_recording.mp3" filename, or will they know about the m3u8 file name? If its just a matter of what's currently stored in CDM being filenames rather than playlists, we could convert that as part of the migration, but if its going to be an ongoing situation then it makes sense to do it here

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+        var in = exchange.getIn();
+        var request = StreamingPropertiesRequestSerializationHelper.toRequest(in.getBody(String.class));
+        var agent = request.getAgent();
+        var pid = PIDs.get(request.getFilePidString());
+
+        aclService.assertHasAccess("User does not have permission to set streaming properties",
+                pid, agent.getPrincipals(), Permission.ingest);
+
+        if (isValid(request, pid)) {
+            var file = repositoryObjectLoader.getFileObject(pid);
```
bbpennel, line 42: It seems like there isn't a way to remove streaming properties from a file. Do you think that should be a separate ticket?

bbpennel (changes requested):
Looks good, left a few comments/questions

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
sharonluong: Hmm, I could probably add it like for the thumbnail request processor?

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
sharonluong: with the current usage for CDM, it just takes what's in CDM as the filename. so you could do it as part of the migration if that's the only source for boxc?

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
sharonluong: oh i didn't even realize that's the one it chose. thanks!

bbpennel (approved):

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/streaming/StreamingPropertiesRequestProcessor.java
bbpennel: What i'm wondering is if for setting these properties on File objecst in the admin ui, if the user is going to know the .mp3 file or the .m3u8 file when they input the value. But perhaps it doesn't matter too much since this method should force it into m3u8 format either way (is that format used for stream audio and video?). Although one issue is that if the user does input `banjo_recording-playlist.m3u8` then this method will turn it into `banjo_recording-playlist-playlist.m3u8`, so we should probably address that

bbpennel (changes requested):

bbpennel (approved):

PR #1711, opened 2024-04-01
BXC-4503 - Handle multiple IPs and cleanup

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4503

* Use the last IP in x-forwarded-for header if multiple are present, since it is the one set by our server rather than the client. 
    * This should resolve the "No access groups" issue we have been having.
* Increase log level when an exception is thrown while setting user access info
* Remove unused code related to setting user access info, some of which was leftover from when the application used JSP for the front end.
* Always clear agentPrincipals when setting groups


sharonluong (approved):
looks good!

PR #1712, opened 2024-04-01
Bxc 4448 api endpoint

sharonluong:
This PR adds necessary files for adding and deleting streaming properties via API endpoint. It also refactors some validations and makes the TestHelper file more generic so it's not just for Processor tests.
https://unclibrary.atlassian.net/browse/BXC-4448

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/StreamingPropertiesController.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/StreamingPropertiesController.java
@@ -33,6 +33,7 @@ import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 @Controller
 public class StreamingPropertiesController {
     private static final Logger log = LoggerFactory.getLogger(StreamingPropertiesController.class);
+    private static final String ACTION = "action";
     @Autowired
     private AccessControlService accessControlService;
     @Autowired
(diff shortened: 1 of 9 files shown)


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/StreamingPropertiesController.java
```diff
@@ -0,0 +1,94 @@
...
+    @Autowired
+    private StreamingPropertiesRequestSender streamingPropertiesRequestSender;
+
+    @PutMapping(value = "/edit/streamingProperties", produces = APPLICATION_JSON_VALUE)
+    @ResponseBody
+    public ResponseEntity<Object> updateStreamingProperties(@RequestParam Map<String,String> allParams) {
+        Map<String, Object> result = new HashMap<>();
+
+        if (hasBadParams(allParams)) {
+            result.put("error", "Streaming properties request must include action, file ID, folder, and filename");
```
bbpennel, line 48: this message isn't totally accurate for a delete request

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/StreamingPropertiesController.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final Logger log = LoggerFactory.getLogger(StreamingPropertiesController.class);
+    private static final String ACTION = "action";
+    @Autowired
+    private AccessControlService accessControlService;
+    @Autowired
+    private StreamingPropertiesRequestSender streamingPropertiesRequestSender;
+
+    @PutMapping(value = "/edit/streamingProperties", produces = APPLICATION_JSON_VALUE)
+    @ResponseBody
+    public ResponseEntity<Object> updateStreamingProperties(@RequestParam Map<String,String> allParams) {
```
bbpennel, line 44: Totally optional, but since we already have a handy request object we might actually be able to do 
```
public ResponseEntity<Object> updateStreamingProperties(StreamingPropertiesRequest streamingRequest) {
```
and have spring build the object for us (we'd still need to add the agent to it manually). Although I guess there is one mismatch with the FilePidString field being named "file" as a param currently. Could make them both the same, or rename it `id` everywhere, or just not worry about it and keep it the way it is.

PR #1721, opened 2024-04-09
BXC-4524 - Cleanup warnings when invalid facet key provided

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4524

* Add special handling of field name not existing when setting facetLimits, and reduce logging of other facet class errors.
* Increase test coverage of the SearchStateFactory createSearchState method
* Add mockito-inline jar so that we can mock static methods (this is built into mockito-core in version 5)

### DIFF
--- a/search-solr/src/test/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactoryTest.java
+++ b/search-solr/src/test/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactoryTest.java
@@ -2,8 +2,11 @@ package edu.unc.lib.boxc.search.solr.services;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertIterableEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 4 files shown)


search-solr/src/test/java/edu/unc/lib/boxc/search/solr/services/SearchStateFactoryTest.java
```diff
@@ -265,4 +277,111 @@ public void extractAdvancedSearchCollectionIdRight() {
...
+        assertTrue(searchState.getFacetLimits().isEmpty());
+    }
+
+    @Test
+    public void extractFacetLimitsInvalidFacetClass() {
+        Map<String, String[]> parameters = new LinkedHashMap<>();
+        parameters.put("anywhere", new String[]{""});
+        parameters.put("facetLimits", new String[]{"language:20"});
+
+        try (MockedStatic<SearchSettings> mockedStatic = mockStatic(SearchSettings.class)) {
```
sharonluong, line 309: what does MockedStatic do?

bbpennel: It lets you mock a static method, which is happening on the next line down. Since SearchSettings.getFacetClass is not a instance method, we can't use a regular mocked object to change its return type.

sharonluong (approved):

PR #1722, opened 2024-04-10
BXC-4537 - Upgrade to mockito 5

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4537

* Update mockito to v5
    * Replaced matchers like anySetOf with anySet, since the "Of" versions were removed
* Controller tests that injected mocks needed to be switched over to standalone server
* needed to add `type="java.lang.Class"` to context files when mocking objects
* ImageEnhancementsRouterTest is now mocking processors since they have their own test and spies in context files were causing issues
* Removed AbstractAPIIT from some tests where it wasn't needed
* Removed powermock dependency, since it was unused/unsupported and mockito now supports mocking static methods


sharonluong (approved):
that was a doozy. nice work!

PR #1723, opened 2024-04-15
BXC-4532 - Fix destroy file operation

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4532

* Destroying files had not been fully updated to work with member order 
    * Inject missing dependency, 
    * prevent indexing error when a tombstone is supposed to be indexed, 
    * improve test coverage in SolrUpdateProcessor, 
    * address codeclimate warning
* Skip splitting member order string if the string is empty, otherwise split returns an array like `[""]`


services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/solrUpdate/SolrUpdateProcessor.java
```diff
@@ -113,6 +119,12 @@ public void process(Exchange exchange) throws Exception {
         }
     }
 
+    private static boolean actionDoesNotSupportTombstones(IndexingActionType actionType) {
```
sharonluong, line 122: so only these three action types support tombstones?

bbpennel: yeah, just indexing actions that delete records from solr. We don't support having solr records for tombstones at the moment

sharonluong (approved):
looks good!

PR #1724, opened 2024-04-15
Bxc 4449 update deposit pipeline

sharonluong:
This PR makes it possible to ingest FileObjects without original files (we validate based on streaming properties). It also updates Redis so tests can run.
https://unclibrary.atlassian.net/browse/BXC-4449


deposit-app/src/main/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -376,64 +376,62 @@ private FileObject addFileToWork(WorkObject work, Resource childResc)
...
+            }
 
-        // Add the FITS report for this file
-        addFitsHistory(fileObj, childResc);
-        addFitsReport(fileObj, childResc);
+            // Add the FITS report for this file
+            addFitsHistory(fileObj, childResc);
+            addFitsReport(fileObj, childResc);
+        } else {
+            fileObj = repoObjFactory.createFileObject(childPid, aipModel);
```
bbpennel, line 432: might be helpful to our future selves to add a comment here to indicate what scenario is in which we would want to be creating a fileObject without an original_file

deposit-app/src/main/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -898,8 +905,13 @@ private void addIngestionEventForChild(ContentObject obj) throws IOException {
...
         if (obj instanceof FileObject) {
-            builder.addEventDetail("ingested as PID: {0}\n ingested as filename: {1}",
-                    obj.getPid().getQualifiedId(), ((FileObject) obj).getOriginalFile().getFilename());
+            try {
+                var originalFile = ((FileObject) obj).getOriginalFile();
+                builder.addEventDetail("ingested as PID: {0}\n ingested as filename: {1}",
+                        obj.getPid().getQualifiedId(), originalFile.getFilename());
+            } catch (Exception e) {
+                builder.addEventDetail("ingested as PID: {0}", obj.getPid().getQualifiedId());
+            }
```
bbpennel, line 914: "Exception" can cover a lot of stuff, what error was being thrown when there was no originalFile? I'd prefer switching this over to an `if` statement which detects when there is no originalFile, so we don't inadvertently end up skipping over unexpected error cases.

deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/ValidateContentModelJob.java
```diff
@@ -207,10 +210,27 @@ private void validateFileObject(Resource resc, Resource parentResc) {
...
-        } else if (!origResc.hasProperty(CdrDeposit.stagingLocation)) {
-            failJob(null, "No staging location provided for file ({0})", resc.getURI());
+        var hasOrigResc = origResc != null;
+
+        if (hasAtLeastOneStreamingProperty(resc)) {
+            if (!hasAllStreamingProperties(resc)) {
+                // if one streaming property is present, they must all be present
+                failJob(null, "All streaming properties are required for file ({0})", resc.getURI());
+            }
+        } else if (hasOrigResc) {
```
bbpennel, line 220: At the moment if there were streaming properties and an original resc, it wouldn't check that there was a staging location. I'd suggest changing this from an `else if` to a series of `if` blocks, and adding a boolean variable like `var hasStreamingOrOriginal = false` and then setting it to true in both the hasOrigResc and hasStreaming blocks, and then using that variable to decide if we want to do the "no streaming or original" error.

deposit-app/src/test/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJobIT.java
```diff
@@ -404,6 +401,48 @@ public void ingestSimpleFile() throws Exception {
...
+        String title = mWork.getResource().getProperty(DC.title).getString();
+        assertEquals(label, title, "Work title was not correctly set");
+        assertClickCount(2);
+        ingestedObjectsCount(2);
+
+        // Check the right number of members are present
+        List<ContentObject> workMembers = mWork.getMembers();
+        assertEquals(1, workMembers.size(), "Incorrect number of members in work");
+        FileObject fileObj = (FileObject) findContentObjectByPid(workMembers, filePid);
+        assertNotNull(fileObj);
```
bbpennel, line 443: we should also assert that the fileobject has the streaming properties with the correct values. We should be able to find them by checking fileObj.getResource()

deposit-app/src/test/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJobTest.java
```diff
@@ -322,29 +331,37 @@ public void ingestWorkObjectTest() throws Exception {
...
+     * Test that ingest will work if a file object is specified with no original resource,
+     * assuming it has streaming properties and passed validation
      */
     @Test
-    public void ingestWorkWithFileWithoutLocation() throws Exception {
-        Assertions.assertThrows(JobFailedException.class, () -> {
-            PID workPid = makePid(RepositoryPathConstants.CONTENT_BASE);
-            WorkObject work = mock(WorkObject.class);
-            Bag workBag = setupWork(workPid, work);
+    public void ingestWorkWithFileWithoutOriginalFileAndStagingLocation() {
```
bbpennel, line 338: The name of the test should probably reflect that the file object is being given streaming properties

deposit-app/src/test/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJobTest.java
```diff
@@ -322,29 +331,37 @@ public void ingestWorkObjectTest() throws Exception {
...
-            when(repoObjLoader.getWorkObject(eq(workPid))).thenReturn(work);
+        job.run();
 
-            job.run();
-        });
+        verify(repoObjFactory).createWorkObject(eq(workPid), any(Model.class));
+        verify(destinationObj).addMember(eq(work));
+
+        verify(jobStatusFactory, times(2)).incrCompletion(eq(jobUUID), eq(1));
     }
```
bbpennel, line 366: I think we should add a test for a file object that has both streaming properties and an original resource. I don't have an opinion whether that gets added to the unit or integration test

deposit-app/src/test/java/edu/unc/lib/boxc/deposit/validate/ValidateContentModelJobTest.java
```diff
@@ -211,6 +213,75 @@ public void missingOriginalDatastreamTest() {
...
+            objBag.addProperty(Cdr.primaryObject, childResc);
+
+            depBag.add(objBag);
+
+            job.closeModel();
+
+            job.run();
+        });
+    }
+
```
bbpennel, line 262: I'd suggest adding a test for a fileObject with all streaming properties and a valid original datastream. Maybe also one for a file object with all streaming properties plus an original datastream but no storage location.

bbpennel (changes requested):
Looks good over all, just a couple suggestions/extra cases

deposit-app/src/main/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJob.java
sharonluong: I did have it as an if statement but the test kept failing because the ingest wouldn't work. it kept failing right after the original file is null check. so i decided to catch the error instead. it didn't tell me what kind of error so i just made it an exception. we could try debugging it some more?

bbpennel (approved):

PR #1730, opened 2024-05-02
BXC-4566 removing unneeded properties

sharonluong:
This PR removes properties in favor of just one that will just point to the StreamReaper Url.
https://unclibrary.atlassian.net/browse/BXC-4566

### DIFF
--- a/model-api/src/main/java/edu/unc/lib/boxc/model/api/rdf/Cdr.java
+++ b/model-api/src/main/java/edu/unc/lib/boxc/model/api/rdf/Cdr.java
@@ -132,22 +132,10 @@ public class Cdr {
             "http://cdr.unc.edu/definitions/model#storageLocation" );
 
     /**
-     * Property which holds the streaming content filename
+     * Property which holds the StreamReaper URL for streaming
      */
-    public static final Property streamingFile = createProperty(
-            "http://cdr.unc.edu/definitions/model#streamingFile");
-
-    /**
-     *  Property which holds the folder, directory, or space of the streaming content
-     */
-    public static final Property streamingFolder = createProperty(
-            "http://cdr.unc.edu/definitions/model#streamingFolder");
-
-    /**
-     * Property that holds where the streaming content is hosted, like Duracloud
-     */
-    public static final Property streamingHost = createProperty(
-            "http://cdr.unc.edu/definitions/model#streamingHost");
+    public static final Property streamingUrl = createProperty(
+            "http://cdr.unc.edu/definitions/model#streamingUrl");
 
     /**
      *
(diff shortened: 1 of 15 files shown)


operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequest.java
```diff
@@ -13,6 +13,7 @@
  * Request object for updating the streaming properties of a FileObject
  */
 public class StreamingPropertiesRequest {
+    public static final String STREAMREAPER_PREFIX_URL = "https://durastream.lib.unc.edu/player";
     public static final String DURACLOUD = "duracloud";
     public static final String OPEN = "open-hls";
     public static final String CLOSED = "closed-hls";
```
bbpennel, line 19: Are these constants still needed?

model-api/src/main/java/edu/unc/lib/boxc/model/api/rdf/Cdr.java
```diff
@@ -132,22 +132,10 @@ public static String getURI() {
             "http://cdr.unc.edu/definitions/model#storageLocation" );
 
     /**
-     * Property which holds the streaming content filename
+     * Property which holds the StreamReaper URL for streaming
```
bbpennel, line 135: While it is true in practice, i don't know that we need to call out streamreaper here since we'd use the same property for non-stream reaper urls

operations-jms/src/test/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequestSerializationHelperTest.java
```diff
@@ -21,18 +22,16 @@ public class StreamingPropertiesRequestSerializationHelperTest {
     public void testHelperSuccess() throws IOException {
         var request = new StreamingPropertiesRequest();
         request.setAgent(agent);
-        request.setFilename("banjo_recording.mp3");
         request.setId(pid.getId());
-        request.setFolder(StreamingPropertiesRequest.OPEN);
+        request.setUrl(STREAMREAPER_PREFIX_URL + "?/params=more");
```
bbpennel, line 26: doesn't matter too much for the test, but we should remove the / in front of params since that isn't a real url pattern

bbpennel (changes requested):

operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/streaming/StreamingPropertiesRequest.java
sharonluong: nope, thanks for noticing!

PR #1733, opened 2024-05-20
Admin interface for setting/clearing access surrogates

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4545


static/js/admin/src/action/AssignAccessSurrogateAction.js
```diff
@@ -0,0 +1,64 @@
...
+            action : 'RefreshResult',
+            target : this.context.target
+        });
+        this.alertHandler.alertHandler("success", `Assignment of object "${this.context.target.metadata.title}" as the assigned access surrogate has completed.`);
+        this.context.target.enable();
+    };
+
+    AssignAccessSurrogateAction.prototype.workDone = function(data) {
+        this.completeTimestamp = data.timestamp;
+        this.oldAccessSurrogate = data.oldAccessSurrogate;
```
bbpennel, line 44: I'm not sure where oldAccessSurrogate and newAccessSurrogate are coming from, I don't see them being set in any of the server side code. Is this doing anything?

static/js/admin/src/action/ClearAccessSurrogateAction.js
```diff
@@ -0,0 +1,66 @@
...
+            action : 'RefreshResult',
+            target : this.context.target
+        });
+        this.alertHandler.alertHandler("success", "Cleared the assigned access surrogate.");
+        this.context.target.enable();
+    };
+
+    ClearAccessSurrogateAction.prototype.workDone = function(data) {
+        this.completeTimestamp = data.timestamp;
+        this.oldAccessSurrogateId = data.oldAccessSurrogateId;
```
bbpennel, line 44: same question here, what is `oldAccessSurrogateId`? I don't think the access surrogates have ids since they are just datastreams of the file object, so I don't think there would be an id that change. Some of the details of the datastream might change, but I'd guess the timestamp/version of the file object to be the main way to tell if the work is done

static/templates/admin/editAccessSurrogateForm.html
```diff
@@ -0,0 +1,19 @@
+<iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
+<form id="create_simple_object_form" method="post" action="/services/api/edit/accessSurrogate/<%= pid %>" enctype="multipart/form-data" target="upload_file_container" class="edit_form create_simple_object_form">
+        <% if (metadata.datastream.findIndex(d => d.startsWith('access_surrogate')) !== -1) { %>
```
bbpennel, line 3: Probably unindent this block once so it lines up with all the other immediate children of the form

bbpennel (changes requested):

bbpennel (approved):

PR #1735, opened 2024-05-20
BXC-4583 - Clear cantaloupe cache after changing jp2

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4583

* Call cantaloupe API endpoint to cleanup cache entries for an item when its JP2 changes.


sharonluong (approved):
looks good!

PR #1739, opened 2024-05-23
Bxc 4558 access surrogate facet

sharonluong:
This PR adds facets for access surrogates: https://unclibrary.atlassian.net/browse/BXC-4558


indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetContentStatusFilterTest.java
```diff
@@ -218,4 +233,46 @@ public void testIsAssignedThumbnail() throws Exception {
...
+
+    @Test
+    public void testWorkWithFileObjectThatHasAccessSurrogate() throws IOException {
+        Files.write(accessSurrogatePath, List.of("fake image"));
+        when(workObj.getResource()).thenReturn(resc);
+        when(dip.getContentObject()).thenReturn(workObj);
+        when(workObj.getResource()).thenReturn(resc);
+        when(fileObj.getResource()).thenReturn(fileResc);
+        workObj.addMember(fileObj);
+        when(fileObj.getParent()).thenReturn(workObj);
```
bbpennel, line 271: workObj is a mock, so calling addMember probably isn't doing anything. SetContentStatusFilter doesn't check anything about the files belonging to a work, so there doesn't seem to be anything productive we could do to set this up in a unit test. Maybe we should just remove all the fileObj stuff here and have the test just verifying that works don't get HAS_ACCESS_SURROGATE or NO_ACCESS_SURROGATE

sharonluong: done!

bbpennel (approved):

PR #1740, opened 2024-05-24
BXC-4592 - More handling of EofException during download

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4592

Specifically check the top level error to see if it is a EofException during download of files. Add a unit test to verify the various exception paths

### DIFF
--- a/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FedoraContentController.java
+++ b/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FedoraContentController.java
@@ -102,22 +102,32 @@ public class FedoraContentController {
             recordDownloadEvent(pid, datastream, principals, request);
         } catch (IOException e) {
             handleIOException(pid, datastream, e);
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
         }
     }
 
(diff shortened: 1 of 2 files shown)


web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FedoraContentController.java
```diff
@@ -154,4 +164,20 @@ public ResponseEntity<Object> handleArgumentTypeMismatch(RuntimeException ex) {
         log.debug("Argument type mismatch", ex);
         return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
     }
+
+    public void setFedoraContentService(FedoraContentService fedoraContentService) {
```
sharonluong, line 168: were these added for the tests?

sharonluong (approved):

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FedoraContentController.java
bbpennel: yes, since it's using a standalone server with a specific controller passed in, rather than being injected by spring

PR #1745, opened 2024-05-31
BXC-4604 - Handle symlink mimetypes and rank

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4604

* Adjust FITS mimetype conflict resolution to rank returned options, filter out invalid results.
* Use a temp dir for storing symlinks, since nano time can very occasionally produce the same name


krwong (approved):
Looks good!

PR #1747, opened 2024-06-05
Bxc 4600 bxc 4601 update export csv

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4600
https://unclibrary.atlassian.net/browse/BXC-4601

### DIFF
--- a/web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ExportCsvIT.java
+++ b/web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ExportCsvIT.java
@@ -208,7 +208,7 @@ public class ExportCsvIT extends AbstractAPIIT {
         List<CSVRecord> csvList = parseCsvResponse(response);
         assertEquals(3, csvList.size(), "Unexpected number of results");
 
-        assertCsvContentIsCorrect(csvList, folderObj.getPid(), pidList);
+        assertCsvContentIsCorrect(csvList, folderObj.getPid(), workPid, filePid, pidList);
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ExportCsvIT.java
```diff
@@ -793,6 +902,22 @@ private void assertCsvRecord(List<CSVRecord> csvList, ResourceType objType, PID
         fail("No CSV record with PID " + expectedPid.getId() + " present");
     }
 
+    private void assertCsvContentIsCorrect(List<CSVRecord> csvList, PID folderPid, Map<String, PID> pidList) {
```
bbpennel, line 905: pidList always contains exactly 2 pids with the same name, so I'd say we should just pass separate variables in for workPid and filePid instead, which would be consistent with folderPid.

bbpennel (changes requested):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/ExportCsvIT.java
```diff
@@ -793,6 +902,21 @@ private void assertCsvRecord(List<CSVRecord> csvList, ResourceType objType, PID
         fail("No CSV record with PID " + expectedPid.getId() + " present");
     }
 
+    private void assertCsvContentIsCorrect(List<CSVRecord> csvList, PID folderPid, PID workPid, PID filePid, Map<String, PID> pidList) {
```
bbpennel, line 905: pidList isn't used here anymore so it can be removed

sharonluong: thanks for catching!

PR #1755, opened 2024-06-18
Bxc 4446 index streaming properties

sharonluong:
This PR indexes the streaming URL property into solr.
https://unclibrary.atlassian.net/browse/BXC-4446

### DIFF
--- a/etc/solr-config/access/conf/schema.xml
+++ b/etc/solr-config/access/conf/schema.xml
@@ -216,6 +216,8 @@
     <field name="rollup" type="string" indexed="true" stored="true" required="true" docValues="true"/>
     <!-- Stores publication status and other statuses that affect viewing of this object outside of group ACLs -->
     <field name="status" type="string" indexed="true" stored="true" multiValued="true"/>
+    <!-- Stores streaming URL for files with streaming property -->
+    <field name="streamingUrl" type="string" indexed="true" stored="true"/>
     <field name="subject" type="string" indexed="true" stored="true" multiValued="true" docValues="true"/>
     <!-- Timestamp when this record was updated -->
     <field name="timestamp" type="date" indexed="true" stored="true" default="NOW" multiValued="false"/>
--- a/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetStreamingUrlFilter.java
+++ b/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetStreamingUrlFilter.java
@@ -6,6 +6,11 @@ import edu.unc.lib.boxc.model.api.rdf.Cdr;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Filter which sets a file's streaming URL. Updates the streamingUrl field.
+ *
+ * @author snluong
+ */
 public class SetStreamingUrlFilter implements IndexDocumentFilter {
     private static final Logger log = LoggerFactory.getLogger(SetStreamingUrlFilter.class);
 
--- a/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetStreamingUrlFilter.java
+++ b/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetStreamingUrlFilter.java
@@ -1,0 +1,22 @@
+package edu.unc.lib.boxc.indexing.solr.filter;
+
+import edu.unc.lib.boxc.indexing.solr.exception.IndexingException;
+import edu.unc.lib.boxc.indexing.solr.indexing.DocumentIndexingPackage;
+import edu.unc.lib.boxc.model.api.rdf.Cdr;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SetStreamingUrlFilter implements IndexDocumentFilter {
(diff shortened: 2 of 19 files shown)


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetStreamingUrlFilter.java
```diff
@@ -0,0 +1,22 @@
+package edu.unc.lib.boxc.indexing.solr.filter;
+
+import edu.unc.lib.boxc.indexing.solr.exception.IndexingException;
+import edu.unc.lib.boxc.indexing.solr.indexing.DocumentIndexingPackage;
+import edu.unc.lib.boxc.model.api.rdf.Cdr;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SetStreamingUrlFilter implements IndexDocumentFilter {
```
bbpennel, line 9: javadoc

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetStreamingUrlFilterTest.java
```diff
@@ -0,0 +1,78 @@
...
+        closeable.close();
+    }
+
+    @Test
+    public void testWithFileObject() {
+        var file = mock(FileObject.class);
+        var url = "https://streaming.url";
+        when(documentIndexingPackageDataLoader.getContentObject(dip)).thenReturn(file);
+        when(file.getResource()).thenReturn(resource);
+        when(resource.hasProperty(eq(Cdr.streamingUrl))).thenReturn(true);
```
bbpennel, line 58: We don't have to change it, but I personally find it easier and less error prone to use a real resource rather than mocking one, like in this example:
https://github.com/UNC-Libraries/box-c/blob/main/indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilterTest.java#L488

etc/solr-config/access/conf/schema.xml
```diff
@@ -216,6 +216,8 @@
     <field name="rollup" type="string" indexed="true" stored="true" required="true" docValues="true"/>
     <!-- Stores publication status and other statuses that affect viewing of this object outside of group ACLs -->
     <field name="status" type="string" indexed="true" stored="true" multiValued="true"/>
+    <!-- Stores streaming URL for files with streaming property -->
+    <field name="streamingUrl" type="string" indexed="true" stored="true"/>
```
bbpennel, line 220: We'll need to also update boxc-ansible with this

bbpennel (changes requested):

etc/solr-config/access/conf/schema.xml
sharonluong: thanks for the reminder! https://gitlab.lib.unc.edu/cdr/boxc-ansible/-/merge_requests/271

indexing-solr/src/test/java/edu/unc/lib/boxc/indexing/solr/test/MockRepositoryObjectHelpers.java
```diff
@@ -89,4 +92,16 @@ public static void addMembers(ContentContainerObject container, ContentObject...
             child.getResource().addProperty(PcdmModels.memberOf, container.getResource());
         }
     }
+
+    public static Resource makeFileResource(String name) {
+        Model model = ModelFactory.createDefaultModel();
+        Resource resc = model.getResource("http://example.com/rest/" + name);
```
bbpennel, line 98: I'd suggest using `PIDs.get(name).getRepositoryPath()`, it would be able to handle full URLs or UUIDs being passed in, and it'll produce a URI more likely to be consistent with other parts of the code that use PIDs.

Although looking at this class, I'm realizing that we could probably use `makeFileObject` instead to produce a FileObject with a resource. We don't care about the repositoryObjectLoader part, but we could have it skip that one line if the loader is null, or split it up into two methods, one that just takes a pid and does almost all the work, and the other which calls the first method and then does the one repositoryObjectLoader related line.

sharonluong: would that affect the other test that use `makeFileObject`? Or does that not matter?

sharonluong: nvm i got it!

static/js/admin/src/ResultObject.js
```diff
@@ -31,7 +31,7 @@ define('ResultObject', [ 'jquery', 'jquery-ui', 'underscore', 'ModalLoadingOverl
 
 		if (tags.length > 0) {
 			this.metadata.tags = tags.filter(function(d) {
-				return !/^(has|not|no.primary|public.access|members.are.unordered|no.assigned.thumbnail|view.behavior|access.surrogate)/i.test(d);
+				return !/^(not|no|has.primary|public.access|members.are.unordered|view.behavior)/i.test(d);
```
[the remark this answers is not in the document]
sharonluong: oh we didn't want that to show, huh? forgot about that

lfarrell (approved):

bbpennel (approved):

PR #1756, opened 2024-06-19
Admin UI for bulk CSV dowload

lfarrell:
- Admin UI for bulk csv export
- Switch request from post to get to work around some issues with ajax not working well with file downloads

https://unclibrary.atlassian.net/browse/BXC-4602

### DIFF
--- a/static/js/admin/src/ExportMenu.js
+++ b/static/js/admin/src/ExportMenu.js
@@ -16,6 +16,7 @@ define('ExportMenu', [ 'jquery', 'jquery-ui', 'underscore', 'qtip', 'cycle'],
             }
             if ($.inArray('editDescription', targets[0].metadata.permissions) !== -1) {
                 items["exportXML"] = {name : "Bulk Metadata"};
+                items["exportCSV"] = {name : "Bulk CSV"};
             }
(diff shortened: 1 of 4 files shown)


static/js/admin/src/ExportMenu.js
```diff
@@ -83,6 +84,14 @@ define('ExportMenu', [ 'jquery', 'jquery-ui', 'underscore', 'qtip', 'cycle'],
                                     });
                                     sessionStorage.removeItem('exportTargets');
                                     break;
+                                case "exportCSV" :
+                                    self.options.actionHandler.addEvent({
+                                        action : 'ChangeLocation',
+                                        url : `api/exportTree/csv?ids=${self.getMemberOrderTargetIds()}`,
```
bbpennel, line 90: I don't think `getMemberOrderTargetIds` is what we want to use here, since it only returns work ids. For this function, the user should be able to select any object types except for the root of the repository (which would get rejected in the controller anyways). You could probably just use `getTargets`

bbpennel (changes requested):

bbpennel (approved):

PR #1758, opened 2024-06-25
Streaming Admin UI

lfarrell:
- https://unclibrary.atlassian.net/browse/BXC-4615
- https://unclibrary.atlassian.net/browse/BXC-4616

### DIFF
--- a/static/js/admin/src/ResultObjectActionMenu.js
+++ b/static/js/admin/src/ResultObjectActionMenu.js
@@ -206,10 +206,10 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 
 		if (metadata.type === 'File' && $.inArray('ingest', metadata.permissions) !== -1) {
 			items["streaming"] = {name: "Streaming Properties", items: {}}
-			items["streaming"]['items']["editStreamingProperties"] = {name: "Edit Streaming URL"};
+			items["streaming"]['items']["editStreamingProperties"] = {name: "Edit Streaming Properties"};
 
 			if (metadata.streamingUrl !== undefined) {
-				items["streaming"]['items']["deleteStreamingProperties"] = {name: "Delete Streaming URL"};
+				items["streaming"]['items']["deleteStreamingProperties"] = {name: "Delete Streaming Properties"};
 			}
 		}
 
--- a/static/js/admin/src/ResultObjectActionMenu.js
+++ b/static/js/admin/src/ResultObjectActionMenu.js
@@ -205,7 +205,12 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 		}
 
 		if (metadata.type === 'File' && $.inArray('ingest', metadata.permissions) !== -1) {
-			items["editStreamingProperties"] = {name : 'Edit Streaming Properties'};
+			items["streaming"] = {name: "Streaming Properties", items: {}}
+			items["streaming"]['items']["editStreamingProperties"] = {name: "Edit Streaming URL"};
+
+			if (metadata.streamingUrl !== undefined) {
+				items["streaming"]['items']["deleteStreamingProperties"] = {name: "Delete Streaming URL"};
+			}
 		}
 
 		// Export actions
@@ -473,6 +478,13 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 						break;
 					case "editStreamingProperties":
 						self.editStreamingProperties(resultObject);
+						break;
+					case "deleteStreamingProperties":
+						self.actionHandler.addEvent({
+							action: 'DeleteStreamingPropertiesResult',
+							target : resultObject,
+							confirmAnchor : options.$trigger
+						});
 						break;
 					case "patronPermissions":
 						perms_editor_store.setPermissionType('Patron');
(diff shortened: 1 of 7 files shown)


static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -205,7 +205,12 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 		}
 
 		if (metadata.type === 'File' && $.inArray('ingest', metadata.permissions) !== -1) {
-			items["editStreamingProperties"] = {name : 'Edit Streaming Properties'};
+			items["streaming"] = {name: "Streaming Properties", items: {}}
+			items["streaming"]['items']["editStreamingProperties"] = {name: "Edit Streaming URL"};
```
bbpennel, line 209: Might make sense to keep it as Edit Streaming Properties and Delete Streaming Properties, since we're going to be adding another property shortly that will be edited at the same time.

static/js/admin/src/action/DeleteStreamingPropertiesResultAction.js
```diff
@@ -0,0 +1,50 @@
...
+    function DeleteStreamingPropertiesResultAction(context) {
+        this._create(context);
+    };
+
+    DeleteStreamingPropertiesResultAction.prototype.constructor = DeleteStreamingPropertiesResultAction;
+    DeleteStreamingPropertiesResultAction.prototype = Object.create( AjaxCallbackAction.prototype );
+
+    DeleteStreamingPropertiesResultAction.prototype._create = function(context) {
+        this.context = context;
+        console.log(this.context)
```
bbpennel, line 12: Probably don't need this

static/js/admin/src/action/DeleteStreamingPropertiesResultAction.js
```diff
@@ -0,0 +1,50 @@
...
+        var options = {
+            workMethod: "PUT",
+            workPath: `/services/api/edit/streamingProperties?id=${this.context.target.metadata.id}&url=${encodeURIComponent(this.context.target.metadata.streamingUrl)}&action=delete`,
```
bbpennel, line 15: the url isn't used for delete requests, so we should be able to remove it from the request path here

static/js/admin/src/action/DeleteStreamingPropertiesResultAction.js
```diff
@@ -0,0 +1,50 @@
...
+            workLabel: "Deleting streaming URL..",
+            followupLabel: "Deleting streaming URL...",
```
bbpennel, line 17: `Deleting streaming properties...`

static/js/admin/src/action/DeleteStreamingPropertiesResultAction.js
```diff
@@ -0,0 +1,50 @@
...
+            followupPath: `/services/api/status/item/${this.context.target.metadata.id}/solrRecord/version`
+        }
+
+        if ('confirm' in this.context && !this.context.confirm) {
+            options.confirm = false;
+        } else {
+            options.confirm = {
+                promptText : "Clear streaming URL?",
```
bbpennel, line 25: For consistency, lets go with `Delete streaming properties?` here and down below.

static/templates/admin/editStreamingPropertiesForm.html
```diff
@@ -2,14 +2,10 @@
 <form id="edit_streaming_properties" class="edit_form edit_streaming_properties_form">
     <div class="form_field">
         <label for="streaming_url"><span class="required">*</span> Streaming URL:</label>
-        <input id="streaming_url" type="text" name="streaming_url" value="<%= metadata.streamingUrl %>"/>
+        <input id="streaming_url" type="text" placeholder="Format: https://durastream.lib.unc.edu/..." name="streaming_url" value="<%= metadata.streamingUrl %>"/>
```
bbpennel, line 5: The text box isn't all the wide, so it might make sense to do one or more of these: 
* remove "Format: " from the beginning of the placeholder text
* Move the text out of the placeholder and into an external message, like adding it to the required field text down below, then the user could potentially copy the base url (although i'm not sure if that would be useful or not since you need the property names, but with more space I guess you could show all the fields)
* Make the streaming_url input wider so that the user can see a whole url at once without needing to horizontally scroll

bbpennel (changes requested):

static/templates/admin/editStreamingPropertiesForm.html
```diff
@@ -1,8 +1,13 @@
 <iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
 <form id="edit_streaming_properties" class="edit_form edit_streaming_properties_form">
+    <p class="form_notes">
+        Structure: https://durastream.lib.unc.edu/player?spaceId=[space-id]&filename=[filename-without-extension]
+    <br />
+        Example: https://durastream.lib.unc.edu/player?spaceId=open-hls&filename=04950_VT0008_0001
+    </p>
```
bbpennel, line 7: I think it might make more sense to put this below in the input rather than before, or at least that's where I'm used to explanatory text for a field being.

static/templates/admin/editStreamingPropertiesForm.html
```diff
@@ -1,8 +1,13 @@
...
     <div class="form_field">
         <label for="streaming_url"><span class="required">*</span> Streaming URL:</label>
-        <input id="streaming_url" type="text" placeholder="Format: https://durastream.lib.unc.edu/..." name="streaming_url" value="<%= metadata.streamingUrl %>"/>
+        <input id="streaming_url" type="text" name="streaming_url" value="<%= metadata.streamingUrl %>"/>
```
bbpennel, line 10: I wonder if a textarea so it can line wrap might work better since the input is fairly long? Otherwise, we might want to make the input 500 or 600px to be able to display the url all at once.

bbpennel (changes requested):

bbpennel (approved):

PR #1767, opened 2024-07-10
Index extent for audio and video files

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4628

### DIFF
--- a/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
+++ b/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
@@ -3,7 +3,6 @@ package edu.unc.lib.boxc.indexing.solr.filter;
 import edu.unc.lib.boxc.indexing.solr.exception.IndexingException;
 import edu.unc.lib.boxc.indexing.solr.indexing.DocumentIndexingPackage;
 import edu.unc.lib.boxc.indexing.solr.utils.Jp2InfoService;
-import edu.unc.lib.boxc.indexing.solr.utils.KduJp2InfoService;
 import edu.unc.lib.boxc.indexing.solr.utils.TechnicalMetadataService;
 import edu.unc.lib.boxc.model.api.DatastreamType;
 import edu.unc.lib.boxc.model.api.exceptions.FedoraException;
@@ -31,6 +30,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static edu.unc.lib.boxc.model.api.DatastreamType.JP2_ACCESS_COPY;
@@ -116,7 +117,6 @@ public class SetDatastreamFilter implements IndexDocumentFilter {
         }
 
         String fitsId = fits.getPid().getId();
-        String extent = null;
 
         try {
             var techMdDoc = technicalMetadataService.retrieveDocument(fits);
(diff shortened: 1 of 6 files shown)


indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
```diff
@@ -128,29 +128,125 @@ private String getExtent(List<BinaryObject> binList) {
...
+    private String normalizeTime(String duration) {
+        String output = "";
+        Pattern pattern = Pattern.compile("\\d+:\\d+:\\d+", Pattern.CASE_INSENSITIVE);
+        Matcher matcher = pattern.matcher(duration);
+        boolean matchFound = matcher.find();
+
+        if (matchFound) {
+            var durationParts = duration.split(":");
+            var hoursToSeconds = Integer.parseInt(durationParts[0]) * 60 * 60;
+            var minutesToSeconds = Integer.parseInt(durationParts[1]) * 60 ;
```
bbpennel, line 218: Bonus space at end of line

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
```diff
@@ -128,29 +128,125 @@ private String getExtent(List<BinaryObject> binList) {
...
+        if (duration != null) {
+            return normalizeTime(duration.getTextTrim());
+        }
+
+        return "-1";
+    }
+
+    private String normalizeTime(String duration) {
+        String output = "";
+        Pattern pattern = Pattern.compile("\\d+:\\d+:\\d+", Pattern.CASE_INSENSITIVE);
```
bbpennel, line 211: Do these durations ever have milliseconds? I can't remember if the pattern would still match in that case. This could also be a constant so that pattern doesn't have to get re-initialized every time

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
```diff
@@ -128,29 +128,125 @@ private String getExtent(List<BinaryObject> binList) {
...
+        Matcher matcher = pattern.matcher(duration);
+        boolean matchFound = matcher.find();
+
+        if (matchFound) {
+            var durationParts = duration.split(":");
+            var hoursToSeconds = Integer.parseInt(durationParts[0]) * 60 * 60;
+            var minutesToSeconds = Integer.parseInt(durationParts[1]) * 60 ;
+            var seconds = Integer.parseInt(durationParts[2]);
+
+            output += hoursToSeconds + minutesToSeconds + seconds;
```
bbpennel, line 221: Seems like output only ever gets one value put into it, so we can skip assigning it to `""` and assign it here rather than doing `+=`

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
```diff
@@ -128,29 +128,125 @@ private String getExtent(List<BinaryObject> binList) {
...
+
+        return output;
+    }
+
+    private Integer millisecondsToSeconds(String duration) {
+        var durationToSeconds = Integer.parseInt(duration) / 1000.0;
+        return (int) Math.ceil(durationToSeconds);
+    }
+
+    private String formatExtent(String imgHeight, String imgWidth, String id) {
```
bbpennel, line 234: Its `id` in this method and `pid` in some of the others. Might as well make it consistent with one or the other. `pid` might be clearer about what kind of id it is?

indexing-solr/src/test/resources/datastream/techmd_mp4.xml
```diff
@@ -0,0 +1,91 @@
...
+                        <fslastmodified toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT">1720543039916</fslastmodified>
+                    </fileinfo>
+                    <filestatus/>
+                    <metadata>
+                        <video>
+                            <location toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">/path/to/object</location>
+                            <mimeType toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">video/quicktime</mimeType>
+                            <format toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">MPEG-4</format>
+                            <formatProfile toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">Base Media</formatProfile>
+                            <duration toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">86616</duration>
```
bbpennel, line 39: It seems like FITS has already settled on a duration when video and audio disagree, maybe we can use this value instead of pulling from the tracks?

indexing-solr/src/test/resources/datastream/techmd_no_duration.xml
```diff
@@ -0,0 +1,105 @@
...
+                        <md5checksum toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT">5f7973ff9e4152827994d4149c8af39d</md5checksum>
+                        <fslastmodified toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT">1720543091535</fslastmodified>
+                    </fileinfo>
+                    <filestatus/>
+                    <metadata>
+                        <video>
+                            <location toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">/opt/data/deposits/31d00601-78cd-480f-b9e8-9b319790f7a4/data/_a49114ad-2276-4e99-afda-67c2fa8f0a24</location>
+                            <mimeType toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">video/mpg</mimeType>
+                            <format toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">MPEG-PS</format>
+                            <duration toolname="MediaInfo" toolversion="23.09" status="SINGLE_RESULT">467</duration>
```
bbpennel, line 36: If we start using the video/duration element then we will probably need to remove this element for the no duration test.

bbpennel (changes requested):

PR #1772, opened 2024-07-15
Index extent for audio and video files

lfarrell:


bbpennel (changes requested):

indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetDatastreamFilter.java
```diff
@@ -129,29 +132,120 @@ private String getExtent(List<BinaryObject> binList) {
...
+    private String formatVideoExtent(Element videoMd, String pid) {
+        var videoInfo = videoMd;
+        var trackInfo = videoMd.getChildren("track", FITS_NS);
+
+        var numTracks = trackInfo.size();
+        if (numTracks > 0) {
+            var videoTrack = 0;
+            for (int i = 0; i < numTracks; i++) {
+                var type = trackInfo.get(i).getAttributeValue("type");
+                if (type.equals(FITS_VIDEO_NAME)) {
```
bbpennel, line 172:
```suggestion
                if (FITS_VIDEO_NAME.equals(type)) {
```
Just to be safe in case FITS decides to be inconsistent, since getAttributeValue will return null if the attribute isn't present. Or we can use `Object.equals(FITS_VIDEO_NAME, type)`

bbpennel (approved):

PR #1776, opened 2024-07-23
Show streaming player for works with streaming children

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4639

### DIFF
--- a/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/HasPopulatedFieldFilter.java
+++ b/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/HasPopulatedFieldFilter.java
@@ -1,0 +1,27 @@
+package edu.unc.lib.boxc.search.solr.filters;
+
+import edu.unc.lib.boxc.search.api.SearchFieldKey;
+import edu.unc.lib.boxc.search.api.filters.QueryFilter;
+
+/**
+ * Filter which restricts results to entries which contain populated values for the given search key,
+ *
+ * @author lfarrell
+ */
+public class HasPopulatedFieldFilter implements QueryFilter {
+    private final SearchFieldKey fieldKey;
+
+    protected HasPopulatedFieldFilter(SearchFieldKey fieldKey) {
+        this.fieldKey = fieldKey;
+    }
+
+    @Override
+    public String toFilterString() {
+        return getFieldKey().getSolrField() + ":*";
+    }
+
+    @Override
+    public SearchFieldKey getFieldKey() {
+        return fieldKey;
+    }
+}
--- a/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/HasPopulatedFieldFilter.java
+++ b/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/HasPopulatedFieldFilter.java
@@ -4,7 +4,7 @@ import edu.unc.lib.boxc.search.api.SearchFieldKey;
 import edu.unc.lib.boxc.search.api.filters.QueryFilter;
 
 /**
- * Filter which restricts results to entries which contain populated values for the given search key,
+ * Filter which restricts results to entries which contain populated values for the given search key
  *
  * @author lfarrell
  */
--- a/static/js/vue-cdr-access/tests/unit/streamingPlayer.spec.js
+++ b/static/js/vue-cdr-access/tests/unit/streamingPlayer.spec.js
@@ -2,22 +2,41 @@ import { shallowMount } from '@vue/test-utils'
 import streamingPlayer from '@/components/full_record/streamingPlayer.vue';
 import cloneDeep from 'lodash.clonedeep';
 
-const streamingUrl = 'https://durastream.lib.unc.edu/player?spaceId=open-hls&filename=R1018_Audio';
-const briefObject = {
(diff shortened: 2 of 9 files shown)


static/js/vue-cdr-access/tests/unit/streamingPlayer.spec.js
```diff
@@ -27,18 +46,33 @@ describe('streamingPlayer.vue', () => {
     });
 
     // Default record url is set in jest.config.js, which is where the refUrl comes from here
-    it('has a source link for the streaming player', () => {
-        expect(wrapper.find('#streaming-player').attributes('src')).toEqual(`${streamingUrl}&refUrl=https://localhost/record/73bc003c-9603-4cd9-8a65-93a22520ef6a`);
+    it('has a source link for the streaming player', async () => {
```
bbpennel, line 49: For all three of the tests updated in this PR, I'd suggest splitting them into separate tests for the file versus work page, it should help with readability and they seem like separate test cases.

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/AccessCopiesService.java
```diff
@@ -88,6 +88,25 @@ public boolean hasViewableFiles(ContentObjectRecord briefObj, AccessGroupSet pri
         return resp.getResultCount() > 0;
     }
 
+    public ContentObjectRecord getStreamingChildren(ContentObjectRecord briefObj, AccessGroupSet principals) {
```
bbpennel, line 91: this only returns one record, so it should probably be `getStreamingChild` or `getFirstStreamingChild`

web-common/src/test/java/edu/unc/lib/boxc/web/common/services/AccessCopiesServiceTest.java
```diff
@@ -364,6 +392,51 @@ public void hasViewableFilesImageWorkTest() {
...
+        assertHasPopulatedFieldFilter(SearchFieldKey.STREAMING_TYPE);
+    }
+
+    @Test
+    public void doesNotHaveStreamingVideoTest() {
+        hasPermissions(mdObjectImg, true);
+        when(searchResultResponse.getResultCount()).thenReturn(0L);
+        assertNull(accessCopiesService.getStreamingChildren(mdObjectImg, principals));
+        assertHasPopulatedFieldFilter(SearchFieldKey.STREAMING_TYPE);
+    }
```
bbpennel, line 430: These two tests are identical, we should probably just keep one and name it "doesNotHaveStreamingTest"

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/filters/HasPopulatedFieldFilter.java
```diff
@@ -0,0 +1,27 @@
+package edu.unc.lib.boxc.search.solr.filters;
+
+import edu.unc.lib.boxc.search.api.SearchFieldKey;
+import edu.unc.lib.boxc.search.api.filters.QueryFilter;
+
+/**
+ * Filter which restricts results to entries which contain populated values for the given search key,
```
bbpennel, line 7: We can remove the trailing comma

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -273,12 +280,20 @@ public String handleUvViewerRequest(@PathVariable("pid") String pidString, Model
...
+        String streamingUrl = null;
+        String streamingType = null;
 
         boolean imageViewerNeeded = accessCopiesService.hasViewableFiles(briefObject, principals);
+        var workStreamingContent = accessCopiesService.getStreamingChildren(briefObject, principals);
+
         if (imageViewerNeeded) {
             viewerType = "uv";
         } else if (briefObject.getContentStatus().contains(FacetConstants.HAS_STREAMING)) {
             viewerType = "streaming";
```
bbpennel, line 292: Maybe we should also set streamingUrl and streamingType in viewerProperties for file records that have streaming, so its always accessible from the same place in the response? Might not need getProperty in streamingPlayer.vue then.

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -273,12 +280,20 @@ public String handleUvViewerRequest(@PathVariable("pid") String pidString, Model
     private Map<String, Object> getViewerProperties(ContentObjectRecord briefObject, AccessGroupSet principals) {
         String viewerType = null;
         String viewerPid = null;
+        String streamingUrl = null;
+        String streamingType = null;
 
         boolean imageViewerNeeded = accessCopiesService.hasViewableFiles(briefObject, principals);
+        var workStreamingContent = accessCopiesService.getStreamingChildren(briefObject, principals);
```
bbpennel, line 287: We could probably refactor this method a bit so that we only execute this query when we need it, since at the moment it will get called for uv or streaming files unnecessarily. But it might be an unnecessary optimization.

bbpennel (changes requested):

bbpennel (approved):

PR #1777, opened 2024-07-23
Bxc 4627 iiif v3 streaming

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4627


web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestControllerTest.java
```diff
@@ -141,4 +144,72 @@ public void testGetCanvas() throws Exception {
...
+
+        var result = mockMvc.perform(get("/iiif/v3/" + OBJECT_ID + "/manifest")
+                        .contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
+                .andReturn();
+
+        Map<String, Object> respMap = MvcTestHelpers.getMapFromResponse(result);
+        assertEquals("Manifest", respMap.get("type"));
+        assertEquals("http://example.com/iiif/v3/f277bb38-272c-471c-a28a-9887a1328a1f/manifest", respMap.get("id"));
+        assertEquals("Test Work", ((List) ((Map) respMap.get("label")).get("none")).get(0));
```
bbpennel, line 164: Just a note that you could use getResponseAsJson here and maybe save some casting, but that's totally up to you.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestControllerTest.java
```diff
@@ -141,4 +144,72 @@ public void testGetCanvas() throws Exception {
...
+        when(accessCopiesService.listViewableFiles(eq(OBJECT_PID), any())).thenReturn(List.of(fileObj));
+
+        var result = mockMvc.perform(get("/iiif/v3/" + OBJECT_ID + "/canvas")
+                        .contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
+                .andReturn();
+
+        var respJson = MvcTestHelpers.getResponseAsJson(result);
+        var body = respJson.get("items").get(0).get("items").get(0).get("body");
+        assertEquals(SERVICES_BASE + "file/" + OBJECT_ID, body.get("id").textValue());
```
bbpennel, line 186: Just a suggestion, but it would be good to use a consistent format for how the expected URI is supplied, since sometimes its by adding together constants and sometimes its a full string. I might have a slight preference towards listing the whole string each time since its easier to read, but either is fine.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestControllerTest.java
```diff
@@ -141,4 +144,72 @@ public void testGetCanvas() throws Exception {
...
+        assertEquals("Sound", body.get("type").textValue());
+        assertEquals(500, body.get(DURATION).intValue());
+    }
```
bbpennel, line 214: Based on the other comments, we'll probably want to add test cases for records without extent, and -1 duration.

bbpennel (changes requested):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestControllerTest.java
sharonluong: Ah, I think this was copy pasted from the test that was already there. I'll take a look and clean up, I think it's not supposed to be here?

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestControllerTest.java
sharonluong: Okay, I'll make it consistent in the whole test!

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/IiifV3ManifestService.java
```diff
@@ -215,27 +224,42 @@ private void setImageContent(ContentObjectRecord contentObj, PaintingAnnotation
...
-            if (Arrays.stream(imgDimensions).count() == 3) {
-                var duration = imgDimensions[2];
-                dimensions.put(DURATION, duration);
+            var imgDimensionsCount = Arrays.stream(imgDimensions).count();
+            if (imgDimensionsCount < 2) {
+                return null;
             }
+            return extractDimensions(imgDimensions, imgDimensionsCount);
+        }
+        return null;
```
bbpennel, line 246: Two suggestions here:
1) We should change it to using the length value rather than streaming and counting.
2) If we flip the conditional to >= 2, then we don't need the extra `return null` and can just use the one at the end of the method.
```
            if (imgDimensions.length >= 2) {
                return extractDimensions(imgDimensions, imgDimensionsCount);
            }
        }
        return null;
```

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/IiifV3ManifestService.java
```diff
@@ -167,21 +219,65 @@ private Canvas constructCanvasSection(ContentObjectRecord contentObj) {
...
+            var imgDimensionsCount = Arrays.stream(imgDimensions).count();
+            if (imgDimensionsCount < 2) {
+                return null;
+            }
+            return extractDimensions(imgDimensions, imgDimensionsCount);
+        }
+        return null;
+    }
+
+    private static HashMap<String, Integer> extractDimensions(String[] imgDimensions, long imgDimensionsCount) {
```
bbpennel, line 249:
```suggestion
    private static Map<String, Integer> extractDimensions(String[] imgDimensions) {
```

Changed HashMap to Map, and we don't really need to pass in the count since we can do imgDimensions.length

bbpennel (changes requested):

bbpennel (approved):

PR #1778, opened 2024-07-25
Add help text for streaming and access surrogates

lfarrell:
Part of investigation into why "no-streaming" and "no-access-surrogate" tags were displaying

https://unclibrary.atlassian.net/browse/BXC-4654


bbpennel (approved):

bbpennel:
Looking back at the ticket that added the access surrogate tags, i think "has access surrogate" was supposed to be suppressed, but I'm the one that wrote the ticket and don't recall if that was just my suggestion or from someone else, and honestly I think its kind of useful to display it and "has streaming" so you can easily spot when they are present since they have a significant impact on the object.

PR #1782, opened 2024-08-09
Bxc 4653 zip file download

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4653

### DIFF
--- a/operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
+++ b/operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
@@ -3,17 +3,18 @@ package edu.unc.lib.boxc.operations.impl.download;
 import edu.unc.lib.boxc.auth.api.Permission;
 import edu.unc.lib.boxc.auth.api.models.AccessGroupSet;
 import edu.unc.lib.boxc.auth.api.services.AccessControlService;
-import edu.unc.lib.boxc.model.api.ids.PID;
 import edu.unc.lib.boxc.model.api.objects.ContentObject;
 import edu.unc.lib.boxc.model.api.objects.FileObject;
-import edu.unc.lib.boxc.model.api.objects.RepositoryObject;
 import edu.unc.lib.boxc.model.api.objects.RepositoryObjectLoader;
 import edu.unc.lib.boxc.model.api.objects.WorkObject;
 import edu.unc.lib.boxc.model.fcrepo.ids.PIDs;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
 
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
@@ -54,6 +55,7 @@ public class DownloadBulkService {
             return;
         }
 
+        Map<String, Integer> duplicates = new HashMap<>();
         for (ContentObject memberObject : memberObjects ) {
             var fileObject = (FileObject) memberObject;
             if (aclService.hasAccess(memberObject.getPid(), agentPrincipals, Permission.viewOriginal)) {
@@ -64,7 +66,13 @@ public class DownloadBulkService {
                 var binaryStream = binObj.getBinaryStream();
                 var filename = binObj.getFilename();
 
-                ZipEntry zipEntry = new ZipEntry(filename);
+                // start keeping track of filenames
+                duplicates.putIfAbsent(filename, 0);
+                var copyNumber = duplicates.get(filename);
+                var zipFilename = formatFilename(filename, copyNumber);
+                duplicates.computeIfPresent(filename, (w, prev) -> prev + 1);
+
+                ZipEntry zipEntry = new ZipEntry(zipFilename);
                 zipOut.putNextEntry(zipEntry);
 
                 IOUtils.copy(binaryStream, zipOut);
@@ -77,7 +85,16 @@ public class DownloadBulkService {
     }
 
     public static String getZipFilename(String workPidString) {
-        return "ZIP-WORK-" + workPidString + ".zip";
+        return "bulk-download-" + workPidString + ".zip";
+    }
+
+    private String formatFilename(String filename, int copyNumber) {
+        if (copyNumber == 0) {
+            return filename;
+        }
+        var extension = FilenameUtils.getExtension(filename);
+        var base = FilenameUtils.removeExtension(filename);
+        return base + "(" + copyNumber + ")." + extension;
     }
 
     public void setAclService(AccessControlService aclService) {
--- a/operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
+++ b/operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
@@ -2,16 +2,16 @@ package edu.unc.lib.boxc.operations.impl.download;
 
 import edu.unc.lib.boxc.auth.api.Permission;
 import edu.unc.lib.boxc.auth.api.models.AccessGroupSet;
-import edu.unc.lib.boxc.auth.api.models.AgentPrincipals;
 import edu.unc.lib.boxc.auth.api.services.AccessControlService;
+import edu.unc.lib.boxc.model.api.DatastreamType;
 import edu.unc.lib.boxc.model.api.ids.PID;
 import edu.unc.lib.boxc.model.api.objects.ContentObject;
-import edu.unc.lib.boxc.model.api.objects.FileObject;
 import edu.unc.lib.boxc.model.api.objects.RepositoryObject;
 import edu.unc.lib.boxc.model.api.objects.RepositoryObjectLoader;
 import edu.unc.lib.boxc.model.api.objects.WorkObject;
-import edu.unc.lib.boxc.model.api.services.RepositoryObjectFactory;
 import edu.unc.lib.boxc.model.fcrepo.ids.PIDs;
+import edu.unc.lib.boxc.search.api.models.ContentObjectRecord;
+import edu.unc.lib.boxc.search.api.models.Datastream;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
```diff
@@ -0,0 +1,111 @@
...
+
+        var zipFilePath = basePath + getZipFilename(workPid.getId());
+
+        zipFiles(workObject, agentPrincipals, zipFilePath);
+        return zipFilePath;
+    }
+
+    private void zipFiles(WorkObject workObject, AccessGroupSet agentPrincipals, String zipFilePath) throws IOException {
+        var memberObjects = workObject.getMembers();
+        final FileOutputStream fos = new FileOutputStream(zipFilePath);
```
bbpennel, line 49: Just a note that if we pass zipFilePath in as a path, we can initialize a FileOutputStream directly with a Path using `Files.newOutputStream(zipFilePath)`

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+            doThrow(new AccessRestrictionException()).when(aclService).assertHasAccess(
+                    anyString(), eq(parentPid), any(), eq(Permission.viewOriginal));
+
+            service.downloadBulk(request);
+        });
+    }
+
+    @Test
+    public void notAWorkTest() {
+        when(repoObjLoader.getWorkObject(any(PID.class))).thenReturn(null);
```
bbpennel, line 103: when the requested object exists but is not a work, it'll actually throw a `ObjectTypeMismatchException` rather than returning null. I don't think the repositoryObjectLoader returns a null in other cases either, I think we get a `NotFoundException` if the object isn't found.

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+        when(repoObjLoader.getWorkObject(any(PID.class))).thenReturn(parentWork);
+        service.downloadBulk(request);
+        // the zip file should be empty
+        assertZipFile(0, List.of());
+    }
+
+    @Test
+    public void noAccessToFileObjectsTest() throws IOException {
+        when(repoObjLoader.getWorkObject(any(PID.class))).thenReturn(parentWork);
+        when(parentWork.getMembers()).thenReturn(List.of(fileObject1));
```
bbpennel, line 120: we should also construct binary so it's clear the access control filter is blocking adding the file rather than the fact there isn't a binary associated with the FileObject.

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+        service.downloadBulk(request);
+        // the zip file should be empty
+        assertZipFile( 0, List.of());
+    }
+
+    @Test
+    public void accessToOnlyOneFileObjectTest() throws IOException {
+        when(repoObjLoader.getWorkObject(any(PID.class))).thenReturn(parentWork);
+        when(parentWork.getMembers()).thenReturn(List.of(fileObject1, fileObject2));
+        makeBinaryObject(fileObject1, FILENAME1);
```
bbpennel, line 132: we should also construct binary for fileObject2, for the same reason as above.

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+        makeBinaryObject(fileObject2, "filename.txt");
+        makeBinaryObject(fileObject3, "flower.txt");
+
+        service.downloadBulk(request);
+
+        // the zip file should have two entries
+        assertZipFile(3, List.of("filename.txt", "filename(1).txt", "flower.txt"));
+    }
+
+    private void assertZipFile(int numberOfEntries, List<String> filenames) throws IOException {
```
bbpennel, line 208: We could remove the numberOfEntries param since the size of filenames should be the expected number of files (I assume we're checking for all files in all cases)

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+                eq(Permission.viewOriginal))).thenReturn(true);
+        when(aclService.hasAccess(eq(fileObject3Pid), any(),
+                eq(Permission.viewOriginal))).thenReturn(true);
+        makeBinaryObject(fileObject1, "filename.txt");
+        makeBinaryObject(fileObject2, "filename.txt");
+        makeBinaryObject(fileObject3, "flower.txt");
+
+        service.downloadBulk(request);
+
+        // the zip file should have two entries
```
bbpennel, line 204: should say 3 entries

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+        assertZipFile(3, List.of("filename.txt", "filename(1).txt", "flower.txt"));
+    }
+
+    private void assertZipFile(int numberOfEntries, List<String> filenames) throws IOException {
+        var zipFilePath = zipStorageBasePath.toString() + getZipFilename(PARENT_UUID);
```
bbpennel, line 209: Use `zipStorageBasePath.resolve(getZipFilename(PARENT_UUID))` to construct the path (generally we should prefer using Path over string concatenation for constructing paths since it is safer)

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+                eq(Permission.viewOriginal))).thenReturn(true);
+        when(aclService.hasAccess(eq(fileObject2Pid), any(),
+                eq(Permission.viewOriginal))).thenReturn(true);
+        makeBinaryObject(fileObject1, FILENAME1);
+        makeBinaryObject(fileObject2, FILENAME2);
+
+        service.downloadBulk(request);
+
+        // the zip file should have two entries
+        assertZipFile(2, List.of(FILENAME1, FILENAME2));
```
bbpennel, line 165: I'm wondering if it'd be good to verify at least one time that the file contained in the zip has the content "flower", just to ensure we aren't getting empty files copied in or something

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
```diff
@@ -0,0 +1,111 @@
...
+        ZipOutputStream zipOut = new ZipOutputStream(fos);
+
+        if (memberObjects.isEmpty()) {
```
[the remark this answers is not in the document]
sharonluong: then that doesn't return an empty zip...how to do that?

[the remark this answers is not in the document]
bbpennel: Ah, good point, I didn't realize that was what it was doing. Maybe just add in a comment then in indicate why it needs to return after the streams have been opened.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
```diff
@@ -0,0 +1,113 @@
...
+        var agentPrincipals = request.getAgent().getPrincipals();
+        aclService.assertHasAccess(
+                "User does not have permissions to view the Work for download",
+                workPid, agentPrincipals, Permission.viewOriginal);
+
+        var zipFilePath = Paths.get(basePath.toString(), getZipFilename(pidString));
+        try {
+            var workObject = repoObjLoader.getWorkObject(workPid);
+            zipFiles(workObject, agentPrincipals, zipFilePath);
+        } catch (Exception e){
```
bbpennel, line 44: I don't think we actually need this catch, repoObjLoader already throws NotFoundException when the object isn't found, and ObjectTypeMismatchException when its not a work. We could change it to catching IOException since zipFiles throws that, if that occurs then we probably would want to throw a `ServiceException`.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
```diff
@@ -0,0 +1,113 @@
...
+            throw new NotFoundException("Failed to perform bulk download for WorkObject " + pidString + " : " + e);
```
bbpennel, line 45: This might be changing to a different exception, but just wanted to note that to include the original exception during a new throw, it should be as a separate parameter:
`throw new NotFoundException("Failed to perform bulk download for WorkObject " + pidString, e);`
If just wanting to show a short description of the previous error, then we should use e.getMessage()

operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
```diff
@@ -0,0 +1,113 @@
...
+                // return an empty zip file
+                return;
+            }
+
+            Map<String, Integer> duplicates = new HashMap<>();
+            for (ContentObject memberObject : memberObjects ) {
+                var fileObject = (FileObject) memberObject;
+                if (aclService.hasAccess(memberObject.getPid(), agentPrincipals, Permission.viewOriginal)) {
+                    var binObj = fileObject.getOriginalFile();
+                    if (binObj == null ) {
```
bbpennel, line 67: bonus space

operations/src/test/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkServiceTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+            service.downloadBulk(request);
+        });
+    }
+
+    @Test
+    public void notAWorkTest() {
+        when(repoObjLoader.getWorkObject(any(PID.class)))
+                .thenThrow(ObjectTypeMismatchException.class);
+        Assertions.assertThrows(NotFoundException.class, () -> {
```
bbpennel, line 108: should be a `ObjectTypeMismatchException`

bbpennel (changes requested):

bbpennel (approved):

PR #1783, opened 2024-08-09
BXC-4664 - Handling of SolrServerExceptions

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4664

* Throw a SolrRuntimeException when a SolrServerException happens, rather than returning null (which results in a NullPointerException later on in some cases, such as the bug encountered here)
* Check whether the result is null in downloadImages controller and throw appropriate error if it is. 
* Removed unused method in SolrSearchService, and some cleanup
* Added direct test coverage of SolrSearchService


sharonluong (approved):
looks good!

PR #1787, opened 2024-08-19
Chompb pre-ingest admin UI

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4685


static/js/admin/vue-cdr-admin/src/components/chompb/preIngest.vue
```diff
@@ -0,0 +1,175 @@
...
+                { data: 'Source' },
+                { data: 'Status' },
+                { data: 'Actions' }
+            ],
+            dataSet: []
+        }
+    },
+
+    computed: {
+        /* ajaxOptions() {
```
bbpennel, line 54: while i think we are going to need ajax options at some point, we probably aren't going to need this search result retrieval logic. Maybe either remove this whole block, or at least remove all the search result stuff to avoid confusion?

static/js/admin/vue-cdr-admin/tests/unit/chompb/preIngest.spec.js
```diff
@@ -0,0 +1,56 @@
...
+        "creator": "bbpennel",
+        "indexedDate": "2024-08-15T12:38:28.662654Z",
+        "sourceFilesUpdatedDate": "2024-08-14T17:06:34.182417Z",
+        "groupMappingsUpdatedDate": "2024-08-14T17:42:49.626664Z",
+        "sipsSubmitted": [],
+        "bxcEnvironmentId": "test",
+        "projectSource": "files",
+        "status": "sources_mapped",
+        "allowedActions": ["color_bar_crop", "color_bar_report"],
+        "projectPath": "/opt/data/cdm_migrations/file_source_test"
```
bbpennel, line 19: since it's a public repo i'd rather not have real server paths, could just make it `/path/to/file_source_test`. Same in the next example

web-admin-app/src/main/java/edu/unc/lib/boxc/web/admin/controllers/ChompbController.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+/**
+ * @author lfarrell
+ */
+@Controller
+public class ChompbController {
```
bbpennel, line 11: Should add a really basic test for this. Shouldn't need any spring context files, see tests like DownloadImageControllerIT for an example, can just use
```
mvc = MockMvcBuilders.standaloneSetup(controller).build();
```

bbpennel (changes requested):

bbpennel (approved):

PR #1788, opened 2024-08-20
BXC-4684 add download zip button

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4684


lfarrell (changes requested):
You'll want to add tests for the button. They should be similar to these two just with the `viewOriginals` permission set and unset https://github.com/UNC-Libraries/box-c/blob/main/static/js/vue-cdr-access/tests/unit/restrictedContent.spec.js#L157-L168

PR #1790, opened 2024-08-22
BXC-4667 update datastream controller IT for iiif thumbnails

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4667


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/DatastreamController.java
```diff
@@ -146,7 +154,24 @@ public void getThumbnail(@PathVariable("pid") String pidString,
             }
         }
 
-        derivativeContentService.streamData(pid, thumbName, principals, false, response);
+        accessControlService.assertHasAccess("Insufficient permissions to get thumbnail for " + pidString,
```
bbpennel, line 157: I know this is the same point where it previous did the access check, but i think we should move it up to above the solr query, otherwise we're giving information about objects that the user doesn't have access to.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/DatastreamController.java
```diff
@@ -116,21 +124,21 @@ private void recordDownloadEvent(PID pid, String datastream, AccessGroupSet prin
...
-    public void getThumbnail(@PathVariable("pid") String pidString,
-            @PathVariable("size") String size, HttpServletRequest request,
-            HttpServletResponse response) throws IOException {
+    public ResponseEntity<InputStreamResource> getThumbnail(@PathVariable("pid") String pidString,
+                                                            @PathVariable("size") String size) {
 
         PID pid = PIDs.get(pidString);
         AccessGroupSet principals = getAgentPrincipals().getPrincipals();
-        String thumbName = "thumbnail_" + size.toLowerCase().trim();
+        if (!size.equals(SMALL) && !size.equals(LARGE)) {
```
bbpennel, line 139: might be overkill given that there are only two sizes, but we could make a constant mapping for the sizes, like
```
private static final Map<String, Integer> THUMB_SIZE_MAP = Map.of(SMALL, 64, LARGE, 128);
```
and then use the map to validate size name and to retrieve the pixel size. 

Also, we should convert the size to lowercase since we were previously.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DatastreamRestControllerIT.java
```diff
@@ -105,13 +119,15 @@ public class DatastreamRestControllerIT extends AbstractAPIIT {
     @BeforeEach
     public void initLocal() {
         closeable = openMocks(this);
+        connectionManager = new PoolingHttpClientConnectionManager();
```
bbpennel, line 122: This doesn't seem to be used other than to shut it down

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DatastreamRestControllerIT.java
```diff
@@ -183,53 +204,57 @@ public void testGetMultipleDatastreams() throws Exception {
...
         // Verify content was retrieved
         MockHttpServletResponse response = result.getResponse();
-        assertEquals(BINARY_CONTENT, response.getContentAsString());
-        assertEquals(BINARY_CONTENT.length(), response.getContentLength());
-        assertEquals("image/png", response.getContentType());
-        assertEquals("inline; filename=\"" + id + "." + THUMBNAIL_SMALL.getExtension() + "\"",
-                response.getHeader(CONTENT_DISPOSITION));
+        // TO DO assert correct image returned
+        assertEquals("image/jpeg", response.getContentType());
+        assertEquals("attachment; filename=file_64px.jpg", response.getHeader(CONTENT_DISPOSITION));
```
bbpennel, line 226: Previously the thumbnail wasn't being returned as an attachment (which triggers a "download" prompt in browsers), and I think we want to keep the old behavior. So we might have to update downloadImageService.streamImage to take a boolean to tell it to do attach mode or not

bbpennel (changes requested):

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/DatastreamController.java
sharonluong: ah yeah, i wasn't sure if the permissions check should be for the work or the file that is the thumbnail. i put it down here to check for the file but you're saying it should be for the work (if it's a work with a file) right?

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DatastreamRestControllerIT.java
sharonluong: oh is that not for Wiremock?

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/DatastreamController.java
bbpennel: hmm, good point. It might be that we should be checking for both the work and the file

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/DatastreamRestControllerIT.java
bbpennel: I don't think so, but it's worth trying to remove it and seeing if the test passes

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/DatastreamController.java
```diff
@@ -136,6 +139,10 @@ public ResponseEntity<InputStreamResource> getThumbnail(@PathVariable("pid") Str
 
         PID pid = PIDs.get(pidString);
         AccessGroupSet principals = getAgentPrincipals().getPrincipals();
+        accessControlService.assertHasAccess("Insufficient permissions to get thumbnail for " + pidString,
+                pid, principals, Permission.viewAccessCopies);
+
+        size = size.toLowerCase();
         if (!size.equals(SMALL) && !size.equals(LARGE)) {
```
bbpennel, line 146: just mentioning we could do this with `THUMB_SIZE_MAP.containsKey(size)`, but it's fine either way

sharonluong: ohh thanks!

bbpennel (approved):

PR #1805, opened 2024-09-19
BXC-4729 - Run enhancements for list of objects

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4729

* Extends Run Enhancements form to allow the list of ids submitted to be hand edited, and adds a checkbox for whether to run enhancements recursively
* Updates Run Enhancements server side code to make recursive execution controlled by a flag.
* RunEnhancementsRequest split into standalone class and used for passing params to service.
* Adds a unit test for run enhancements service

### DIFF
--- a/static/js/admin/src/action/RunEnhancementsBatchAction.js
+++ b/static/js/admin/src/action/RunEnhancementsBatchAction.js
@@ -30,8 +30,12 @@ define('RunEnhancementsBatchAction', [ 'jquery', 'AbstractBatchAction', "tpl!../
 		} else {
 			title = "Run enhancements on " + this.targets.length + " objects";
 		}
-		
-		var form = runEnhancementsTemplate();
+
+		var targetIds = "";
+		for (var index in this.targets) {
+			targetIds += this.targets[index].getPid() + "\n";
+		}
+		var form = runEnhancementsTemplate({ targetIds : targetIds });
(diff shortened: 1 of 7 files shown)


static/js/admin/src/action/RunEnhancementsBatchAction.js
```diff
@@ -30,8 +30,12 @@ define('RunEnhancementsBatchAction', [ 'jquery', 'AbstractBatchAction', "tpl!../
...
 			title = "Run enhancements on " + this.targets.length + " objects";
 		}
-		
-		var form = runEnhancementsTemplate();
+
+		var targetIds = "";
+		for (var index in this.targets) {
+			targetIds += this.targets[index].getPid() + "\n";
+		}
+		var form = runEnhancementsTemplate({ targetIds : targetIds });
```
sharonluong, line 38: where is `runEnhancementsTemplate` defined? how does it accept this new argument?

bbpennel: it's defined in static/templates/admin/runEnhancementsForm.html, which is imported into this file in the first line:
https://github.com/UNC-Libraries/box-c/blob/da0248538409c14e4ef93f9ea22189a41d231a3c/static/js/admin/src/action/RunEnhancementsBatchAction.js#L1
The template library that we're using can take an object as an argument, and it exposes the values in the object as parameters in the template

PR #1822, opened 2024-10-21
Add a total file size warning to work level zip download

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4755

### DIFF
--- a/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
+++ b/web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
@@ -223,7 +223,7 @@ public class FullRecordController extends AbstractErrorHandlingSearchController 
             recordProperties.put("dataFileUrl", dataFileUrl);
         }
 
-        String totalDownloadSize = null;
+        Long totalDownloadSize = null;
         if (ResourceType.Work.nameEquals(resourceType)) {
             totalDownloadSize = workFilesizeService.getTotalFilesize(briefObject, principals);
         }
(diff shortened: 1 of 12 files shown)


web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
```diff
@@ -220,6 +223,12 @@ public String getFullObjectView(String pidString) {
             recordProperties.put("dataFileUrl", dataFileUrl);
         }
 
+        String totalDownloadSize = null;
+        if (ResourceType.Work.nameEquals(resourceType)) {
```
bbpennel, line 227: do you think it'd make sense to not set the property unless its a work, rather than returning null? Doesn't seem like it would impact the views as far as i can tell, but it doesn't matter too much.

web-common/src/main/java/edu/unc/lib/boxc/web/common/services/WorkFilesizeService.java
```diff
@@ -0,0 +1,96 @@
...
+        var ONE_GIGABYTE = 1073741824L;
+        var request = buildChildrenQuery(contentObjectRecord, principals);
+        var resp = solrSearchService.getSearchResults(request);
+
+        if (resp.getResultCount() > 0) {
+            var totalFileSize = resp.getResultList().stream()
+                    .mapToLong(ContentObjectRecord::getFilesizeSort)
+                    .reduce(0, Long::sum);
+
+            if (totalFileSize > ONE_GIGABYTE) {
```
bbpennel, line 41: I would also recommend having this service always return the actual size, and then having either the full record controller or the vue code decide not to make the download available based on size. That way we could use the service as a way to get the full work size for other purposes, or show what the full bulk download size would be if they requested it from wilson. 

I'm also wondering how we want to handle blocking download of really big bulk downloads if the user put in the url directly, but that seems like a separate ticket.

bbpennel (changes requested):

web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/FullRecordController.java
lfarrell: I think I find it more confusing when properties don't show up.

bbpennel (approved):

PR #1823, opened 2024-10-21
BXC-4725 m4a derivatives for access

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4725](https://unclibrary.atlassian.net/browse/BXC-4725)

- add tests to verify that `AUDIO_ACCESS_COPY` is downloadable using the datastream controller and that the solr record has a datastream entry for `AUDIO_ACCESS_COPY`
- add `aac` to `AUDIO_MIMETYPE_REGEX` and `FILE_TYPES`
- pull main into m4a-development


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/IiifV3ManifestService.java
```diff
@@ -298,7 +299,8 @@ private boolean isVideo(String mimetype) {
     }
 
     private boolean isAudio(String mimetype) {
```
bbpennel, line 301: I think at the moment we will only get audio IIIF manifests if the original file is mp4/mp3/m4a, but we can actually support all of the audio types that we permit m4a's be generated for. I think we need to update the logic of "isAudio" to check if the ContentObjectRecord has a audio access datastream and return true if so, in addition to checking the the mimetype for cases where we haven't generated the m4a yet.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/IiifV3ManifestService.java
```diff
@@ -63,7 +64,7 @@ public class IiifV3ManifestService {
     public static final String WIDTH = "width";
     public static final String HEIGHT = "height";
     private static final List<String> FILE_TYPES = Arrays.asList(VIDEO_MP4.toString(), VIDEO_MPEG.toString(),
-            VIDEO_QUICKTIME.toString(), AUDIO_MP4.toString(), AUDIO_MPEG.toString());
+            VIDEO_QUICKTIME.toString(), AUDIO_MP4.toString(), AUDIO_MPEG.toString(), AUDIO_AAC.toString());
```
bbpennel, line 67: rather than expanding this, I think we should update `IIIFv3ViewableFilter` to add another datastream filter, so that it'll find records that have the audio access datastream (in addition to jp2s and stuff matching the mimetype)

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/IiifV3ManifestService.java
```diff
@@ -45,6 +45,7 @@
 import java.util.Objects;
```
bbpennel, line 45: we also need to update `setSoundContent` so that when there is a audio access copy, it will set the download path to the audio access copy instead of the original file. This should tell IIIF viewers to display the access copy instead of the original. We should make sure to verify that in a test as well.

bbpennel (changes requested):

bbpennel (approved):

PR #1831, opened 2024-11-05
Move file downloads into the files table

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4756


lfarrell:
Should also fix https://unclibrary.atlassian.net/browse/BXC-4778 now.

bbpennel (approved):

PR #1835, opened 2024-11-14
Bxc 4771 upper file limit

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4771

### DIFF
--- a/operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
+++ b/operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
@@ -62,10 +62,8 @@ public class DownloadBulkService {
 
             Map<String, Integer> duplicates = new HashMap<>();
             int count = 0;
-            var limit = getLoopLimit(memberObjects.size());
-//            while (count < getLoopLimit(memberObjects.size())) {
             for (ContentObject memberObject : memberObjects ) {
-                if (count == limit) {
+                if (count == getLoopLimit(memberObjects.size())) {
                     break;
                 }
                 if (!(memberObject instanceof FileObject)) {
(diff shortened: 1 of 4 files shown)


web-services-app/src/main/webapp/WEB-INF/service-context.xml
```diff
@@ -498,6 +498,7 @@
         <property name="aclService" ref="aclService"/>
         <property name="repoObjLoader" ref="repositoryObjectLoader"/>
         <property name="basePath" value="${data.dir}/zip/"/>
+        <property name="fileLimit" value="500" />
```
bbpennel, line 501: We still need to figure out what the limit should be. We can deploy this to test and try it out with works that have different numbers of files like listed in the files, and then adjust as needed

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/download/DownloadBulkService.java
```diff
@@ -60,7 +61,12 @@ private void zipFiles(WorkObject workObject, AccessGroupSet agentPrincipals, Pat
             }
 
             Map<String, Integer> duplicates = new HashMap<>();
+            int count = 0;
+            var loopLimit = getLoopLimit(memberObjects.size());
             for (ContentObject memberObject : memberObjects ) {
+                if (count == loopLimit) {
```
bbpennel, line 67: after thinking about this a bit more, I think this can be changed to just
```
if (count == fileLimit)
```
Since the loop is never going to go past the end of the array. Also, if we start count at 1, then we can set the limit to 100 instead of 99, which I think would be clearer about the intent.

bbpennel (approved):

PR #1839, opened 2024-11-26
BXC-4762 integrate jp24u jar into box-c

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4762](https://unclibrary.atlassian.net/browse/BXC-4762)

- add jp24u as submodule to box-c
- add jp24u to the build process
- add jp24u dependency to `pom.xml`
- add `runJp24u` method to `ImageDerivativeProcessor` and use `runJp24u` instead of `convertJp2.sh`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,8 +34,6 @@ jobs:
 
     - name: Checkout repository
       uses: actions/checkout@v3
-      with:
-        submodules: 'recursive'
 
     - name: Get Container ID of fedora service
       id: get_container_id
@@ -59,6 +57,9 @@ jobs:
     - name: Run solr container as command to trigger core creation
       run: |
         docker run -v /tmp/solr-config:/solr_config -d --rm -p 48983:8983 solr:9 solr-precreate access /solr_config/access
+
+    - name: Checkout submodules
+      run: git submodule update --init --recursive
 
     - name: Set up JDK 11
       uses: actions/setup-java@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,9 +71,6 @@ jobs:
         path: ~/.m2
         key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
         restore-keys: ${{ runner.os }}-m2
-    - name: Build jp24u with Maven
-      run: mvn -B -U clean install -DskipTests
-      working-directory: jp24u
     - name: Build with Maven
       run: mvn -B -U -pl '!integration' clean install -DskipTests
     - name: Set up nodejs
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,6 +72,9 @@ jobs:
         path: ~/.m2
         key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
         restore-keys: ${{ runner.os }}-m2
+    - name: Build jp24u with Maven
+      run: mvn -B -U clean install -DskipTests
+      working-directory: jp24u
     - name: Build with Maven
       run: mvn -B -U -pl '!integration' clean install -DskipTests
     - name: Set up nodejs
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
         key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
         restore-keys: ${{ runner.os }}-m2
     - name: Build with Maven
-      run: mvn -B -U -pl '!integration' -pl '!jp24u' clean install -DskipTests
+      run: mvn -B -U -pl '!integration,!jp24u' clean install -DskipTests
     - name: Set up nodejs
       uses: actions/setup-node@v2
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
         key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
         restore-keys: ${{ runner.os }}-m2
     - name: Build with Maven
-      run: mvn -B -U -pl '!integration' clean install -DskipTests
+      run: mvn -B -U -pl '!integration' -pl '!jp24u' clean install -DskipTests
     - name: Set up nodejs
       uses: actions/setup-node@v2
       with:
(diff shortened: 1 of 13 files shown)


.github/workflows/build.yml
```diff
@@ -72,6 +72,9 @@ jobs:
         path: ~/.m2
         key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
         restore-keys: ${{ runner.os }}-m2
+    - name: Build jp24u with Maven
```
bbpennel, line 74: I don't think we need this extra build, since the "Build with Maven" step at line 78 will also build jp24u since its a submodule

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/images/ImageDerivativeProcessor.java
```diff
@@ -47,6 +48,21 @@ public static boolean allowedImageType(Exchange exchange) {
...
+     * Run jp24u kdu_compress command and generate image derivatives
+     * @param exchange
+     */
+    public void runJp24u(Exchange exchange) {
+        Message in = exchange.getIn();
+        String imagePath = (String) in.getHeader(CdrFcrepoHeaders.CdrImagePath);
+        String tempPath = (String) in.getHeader(CdrFcrepoHeaders.CdrTempPath);
+        String mimetype = (String) in.getHeader(CdrFcrepoHeaders.CdrBinaryMimeType);
+
+        String[] command = new String[]{"jp24u", "kdu_compress", "-f", "\"" + imagePath + "\"",
```
bbpennel, line 61: I hadn't considered that we might still be able to interact with it like a CLI command. Calling it this way causes jp24u to execute within the same process? If so, it's probably simpler than directly calling the command classes

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/images/ImageDerivativeProcessor.java
```diff
@@ -47,6 +48,21 @@ public static boolean allowedImageType(Exchange exchange) {
         return true;
     }
 
+    /**
+     * Run jp24u kdu_compress command and generate image derivatives
+     * @param exchange
+     */
+    public void runJp24u(Exchange exchange) {
```
bbpennel, line 55: I know this is fairly short, but my suggestion would be to make a new Jp2Processor for this. That would keep the scope of ImageDerivativeProcessor narrower, and would mean we could call `.bean(jp2Processor)` in the camel route, since this method would become "process"

We should also have a test for this method/processor. I'm not sure it'd be practical to actually test that jp24u works in boxc since it requires so many dependencies including one that requires a license, but we can at least mock `CLIMain.main` (search the code base for `mockStatic` for examples)

bbpennel (changes requested):

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/images/ImageDerivativeProcessor.java
krwong: According to chatgpt it should work so I figured I'd give it a shot. It looked a lot simpler than directly calling the command classes.

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/images/ImageEnhancementsRouter.java
```diff
@@ -59,8 +62,7 @@ public void configure() throws Exception {
                 .setBody(exchange -> uuidGenerator.generateUuid())
                 .setHeader(CdrFcrepoHeaders.CdrTempPath, simple("${properties:services.tempDirectory}/${body}-access"))
                 .doTry()
-                    .recipientList(simple("exec:/bin/sh?args=${properties:cdr.enhancement.bin}/convertJp2.sh "
-                            + "${headers[CdrImagePath]} ${headers[CdrMimeType]} ${headers[CdrTempPath]}"))
+                    .recipientList(simple("bean:jp2Processor"))
```
bbpennel, line 65:
```suggestion
                    .bean(jp2Processor)
```

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/images/Jp2Processor.java
```diff
@@ -0,0 +1,26 @@
...
+
+import JP2ImageConverter.CLIMain;
+import edu.unc.lib.boxc.services.camel.util.CdrFcrepoHeaders;
+import org.apache.camel.Exchange;
+import org.apache.camel.Message;
+import org.apache.camel.Processor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Jp2Processor implements Processor {
```
bbpennel, line 11: javadoc

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/images/Jp2Processor.java
```diff
@@ -0,0 +1,26 @@
...
+    private static final Logger log = LoggerFactory.getLogger(Jp2Processor.class);
+
+    @Override
+    public void process(Exchange exchange) throws Exception {
+        Message in = exchange.getIn();
+        String imagePath = (String) in.getHeader(CdrFcrepoHeaders.CdrImagePath);
+        String tempPath = (String) in.getHeader(CdrFcrepoHeaders.CdrTempPath);
+        String mimetype = (String) in.getHeader(CdrFcrepoHeaders.CdrBinaryMimeType);
+
+        String[] command = new String[]{"jp24u", "kdu_compress", "-f", "\"" + imagePath + "\"",
```
bbpennel, line 21: My understanding is that when we are directly invoking the picocli interface like this with an array of params, it doens't need to be quoted like it would if being executed via a ProcessBuilder, since its not actually calling out to a shell. So I think we can remove all the `"\""` bits and just pass in the variables directly, since picocli knows they are separate arguments since they are separate array elements. Its worth a try at least, and would avoid us having to add in quote escaping in case paths or mimetypes had quotes in them.

bbpennel (changes requested):

services-camel-app/src/test/java/edu/unc/lib/boxc/services/camel/images/ImageEnhancementsRouterTest.java
```diff
@@ -89,7 +86,7 @@ public void testAccessCopyRouteNoForceNoFileExists() throws Exception {
         createContext(accessCopyRoute);
 
         var jp2Endpoint = getMockEndpoint("mock:bean:jp2Processor");
```
bbpennel, line 88: this doesn't exist anymore since we're calling `.bean(jp2Processor)` directly, which is why its always being called 0 times. But you should be able to do `verify(jp2Processor).process(any(Exchange.class));` instead to check if its been called or not

bbpennel (changes requested):

bbpennel (approved):

PR #1841, opened 2024-11-27
BXC-4794 - Too many files for bulk download

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4794

Add alert which prompts user to confirm they want to go through with export when there are too many files

### DIFF
--- a/static/js/vue-cdr-access/src/components/full_record/bulkDownload.vue
+++ b/static/js/vue-cdr-access/src/components/full_record/bulkDownload.vue
@@ -1,9 +1,12 @@
 <template>
     <div v-if="hasDownloadableContent" class="column pr-0 has-text-right">
-        <a v-if="showTotalFilesize" class="bulk-download bulk-download-link button action is-primary" :href="downloadBulkUrl(workId)">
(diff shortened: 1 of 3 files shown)


static/js/vue-cdr-access/src/components/full_record/bulkDownload.vue
```diff
@@ -35,17 +38,41 @@ export default {
...
         showTotalFilesize() {
             return this.hasBulkDownloadAccess && this.totalDownloadSize <= ONE_GIGABYTE;
         }
+    },
+
+    methods: {
+        handleDownload(event) {
+            if (this.hasTooManyChildren) {
+                // Prevent navigation if user cancels
+                if (!confirm("Number of files exceeds the download limit, only the first 100 will be exported, do you want continue?")) {
```
sharonluong, line 67: This is a mouthful. Perhaps a period would flow better?
"Number of files exceeds the download limit, only the first 100 will be exported. Do you want to continue?"

sharonluong (approved):
tested and works! just that one change for the sentence in the alert

PR #1857, opened 2025-01-06
BXC-4804 - Velocicroptor in the background

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4804

* Run chompb crop command as a background process rather than making the users request wait for it to complete


sharonluong (approved):
looks good, thanks!

PR #1864, opened 2025-01-13
BXC-4802 updating tests and code for zero byte JP2s

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4802
Since I updated the Test Corpus and added two new objects, I had to also update the Delete Solr Tree test accordingly.


web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/IiifV3ManifestServiceTest.java
```diff
@@ -214,7 +239,7 @@ public void buildManifestViewableFileVideoTest() {
 
     @Test
     public void buildCanvasViewableFileTest() {
```
bbpennel, line 241: Since it's new logic, it would be good to add a `buildCanvasNonViewableFileTest` to verify the new conditional in buildCanvas is working

bbpennel (changes requested):

bbpennel (approved):

PR #1866, opened 2025-01-14
BXC-4801 - Fix PDF viewer errors

bbpennel:
* Prevents errors on full record page controller endpoints if the user accesses a work with a PDF file, but that PDF file is not accessible
* Prevents error is pdfViewer endpoint is accessed for a non-PDF file.
* When setting up pdf viewer property for full record responses, ensure that the user has permission to view the PDF file. If a pdf view is requested for a non-pdf file, then clear the viewer settings.
* Adds a unit test for partial coverage of full record controller

### DIFF
--- a/web-access-app/src/test/java/edu/unc/lib/boxc/web/access/controllers/FullRecordControllerTest.java
+++ b/web-access-app/src/test/java/edu/unc/lib/boxc/web/access/controllers/FullRecordControllerTest.java
@@ -1,0 +1,254 @@
+package edu.unc.lib.boxc.web.access.controllers;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.type.MapType;
+import edu.unc.lib.boxc.auth.api.Permission;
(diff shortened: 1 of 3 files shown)


web-access-app/src/test/java/edu/unc/lib/boxc/web/access/controllers/FullRecordControllerTest.java
```diff
@@ -0,0 +1,359 @@
...
+
+        Map<String, Object> respMap = getMapFromResponse(result);
+        assertEquals(PID_1, respMap.get(VIEWER_PID));
+        assertEquals("clover", respMap.get(VIEWER_TYPE));
+        assertNull(respMap.get(STREAMING_TYPE));
+        assertNull(respMap.get(STREAMING_URL));
+        assertNotNull(respMap.get("briefObject"));
+    }
+
+    public static Map<String, Object> getMapFromResponse(MvcResult result) throws Exception {
```
krwong, line 354: Should this helper method be private?

krwong (approved):

PR #1895, opened 2025-03-04
BXC-4869 display archival collection id in additional locations

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4869](https://unclibrary.atlassian.net/browse/BXC-4869)

- add `collectionId` sortable field to admin UI
- if the `collectionId` is present, display `collectionId` in collection browse and under the collection name when collection results are returned

### DIFF
--- a/static/js/vue-cdr-access/src/components/listDisplay.vue
+++ b/static/js/vue-cdr-access/src/components/listDisplay.vue
@@ -100,7 +100,7 @@ Renders search results in a list view display format
             },
 
             showCollection(record) {
-                return record.type !== 'AdminUnit' && record.type !== 'Collection' &&
+                return record.type !== 'AdminUnit' && record.type === 'Collection' &&
                     record.objectPath[2].collectionId !== null;
             }
         }
--- a/static/js/vue-cdr-access/src/components/listDisplay.vue
+++ b/static/js/vue-cdr-access/src/components/listDisplay.vue
@@ -100,8 +100,7 @@ Renders search results in a list view display format
             },
 
             showCollection(record) {
-                return record.type !== 'AdminUnit' && record.type === 'Collection' &&
-                    record.objectPath[2].collectionId !== null;
+                return record.type !== 'AdminUnit' && record.objectPath[2].collectionId !== null;
             }
         }
     }
--- a/static/templates/admin/resultEntry.html
+++ b/static/templates/admin/resultEntry.html
@@ -36,7 +36,7 @@
 	</td>
 	<td class="collection_id">
 		<div title="<%= metadata.collectionId ? metadata.collectionId : '' %>">
-			<%= metadata.collectionId? metadata.collectionId[0] : '-' %>
+			<%= metadata.collectionId? metadata.collectionId : '-' %>
 		</div>
 	</td>
 	<td class="creator">
--- a/static/templates/admin/resultEntry.html
+++ b/static/templates/admin/resultEntry.html
@@ -34,11 +34,15 @@
 			<% } %>
 		</div>
 	</td>
-	<td class="collection_id">
-		<div title="<%= metadata.collectionId ? metadata.collectionId : '' %>">
-			<%= metadata.collectionId? metadata.collectionId : '-' %>
-		</div>
-	</td>
+	<% if (metadata.type == "Collection") { %>
+		<td class="collection_id">
+			<div title="<%= metadata.collectionId ? metadata.collectionId : '' %>">
+				<%= metadata.collectionId? metadata.collectionId : '-' %>
+			</div>
+		</td>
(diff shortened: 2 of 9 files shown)


static/js/vue-cdr-access/src/components/listDisplay.vue
```diff
@@ -100,7 +100,7 @@ Renders search results in a list view display format
             },
 
             showCollection(record) {
-                return record.type !== 'AdminUnit' && record.type !== 'Collection' &&
+                return record.type !== 'AdminUnit' && record.type === 'Collection' &&
```
bbpennel, line 103: I think this means that only Collection objects would pass now. I think your goal here is to get it to also show the archival collection id for Collection objects? If so, then I think removing `&& record.type === 'Collection'` would have that effect

web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/SerializationUtil.java
```diff
@@ -140,6 +140,10 @@ public static Map<String, Object> metadataToMap(ContentObjectRecord metadata, Ac
             result.put("type", metadata.getResourceType());
         }
 
+        if (metadata.getCollectionId() != null) {
+            result.put("collectionId", metadata.getCollectionId());
```
bbpennel, line 144: we're only doing it with some of the attributes at the moment, but generally it would be good to use the constant for these field names, so here the key could be `SearchFieldKey.COLLECTION_ID.getUrlParam()`.

bbpennel (changes requested):

web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/SerializationUtil.java
```diff
@@ -113,43 +113,43 @@ public static Map<String, Object> metadataToMap(ContentObjectRecord metadata, Ac
...
         }
 
         if (metadata.getTitle() != null) {
-            result.put("title", metadata.getTitle());
+            result.put(SearchFieldKey.TITLE.getUrlParam(), metadata.getTitle());
         }
 
         if (metadata.get_version_() != null) {
-            result.put("_version_", metadata.get_version_());
+            result.put(SearchFieldKey.VERSION.getUrlParam(), metadata.get_version_());
```
bbpennel, line 124: these aren't actually equivalent at the moment, urlParam doesn't have the _'s. And while that is what we ultimately want, we'd have to update the UI code to expect the correct name, which might be out of scope for this ticket

bbpennel (changes requested):

static/templates/admin/resultEntry.html
```diff
@@ -34,6 +34,15 @@
...
 		</div>
 	</td>
+	<% if (metadata.type == "Collection") { %>
+		<td class="collection_id">
+			<div title="<%= metadata.collectionId ? metadata.collectionId : '' %>">
+				<%= metadata.collectionId? metadata.collectionId : '-' %>
+			</div>
+		</td>
+	<% } else { %>
+		<td class="collection_id" style="display:none"></td>
```
bbpennel, line 44: Can probably remove this

static/templates/admin/resultTableView.html
```diff
@@ -5,17 +5,20 @@
 			<table class="result_table">
 				<colgroup>
 					<% _.each(resultFields, function(resultField) { %>
-						<col class="<%= resultField.colClass %>">
-					<% }); %>
+						<% if (resultField.colClass == "collection_id" && container.type != "AdminUnit") { %>
```
bbpennel, line 8: generally we avoid having empty conditionals. We could invert this so we don't need the else.

`if (!(resultField.colClass == "collection_id" && container.type != "AdminUnit"))`
 or 
`if (resultField.colClass != "collection_id" || container.type == "AdminUnit")`

static/templates/admin/resultTableView.html
```diff
@@ -5,17 +5,20 @@
...
 				</colgroup>
 				<thead>
 					<tr class="column_headers">
 						<% _.each(resultFields, function(resultField) { %>
-							<th class="<%= resultField.colClass 
-									%><%= ('sortField' in resultField)? ' sort_col' : '' %>" <%= 
-									('sortField' in resultField)? ' data-field="' + resultField.sortField + '"' : '' %> <%= 
-									('dataType' in resultField)? ' data-type="' + resultField.dataType + '"' : '' %>><a><%= resultField.name %></a></th>
-						<% }); %>
+							<% if (resultField.colClass == "collection_id" && container.type != "AdminUnit") { %>
```
bbpennel, line 16: see comment about empty conditionals

bbpennel (changes requested):

PR #1901, opened 2025-03-13
BXC-4893 update finding aid links format

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4893](https://unclibrary.atlassian.net/browse/BXC-4893)

- update finding aid links from `https://finding-aids.lib.unc.edu/40080/` to `https://finding-aids.lib.unc.edu/catalog/40080`
- update finding aid url in test resource

### DIFF
--- a/indexing-solr/src/test/resources/datastream/inventoryMods.xml
+++ b/indexing-solr/src/test/resources/datastream/inventoryMods.xml
@@ -55,7 +55,7 @@
   </mods:physicalDescription>
   <mods:relatedItem type="isReferencedBy">
     <mods:location>
-      <mods:url displayLabel="Link to finding aid">https://finding-aids.lib.unc.edu/1234</mods:url>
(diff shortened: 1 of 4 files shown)


cazzerson (approved):

cazzerson:
Looks good, but I'm not sure that the trailing slash from https://github.com/UNC-Libraries/box-c/blob/7ba1941a1e195504c8c67619ae8ba829577a8ada/web-common/src/main/java/edu/unc/lib/boxc/web/common/services/FindingAidUrlService.java#L53 is necessary. This isn't how ASpace builds the links, but the links seem valid. @bbpennel?

bbpennel:
I think the updates in this PR only affect tests. It looks like `access.properties.j2` in the boxc-ansible project will need to be updated as well, since that is where `findingaids.base.url` used in the deployed app comes from.

@cazzerson I'm not sure why that trailing slash is there. It looks like it has been there since the first commit. I'd be fine with dropping it if it is not needed. We can try it out on dcr test to see if it has any impact.

bbpennel (approved):

PR #1902, opened 2025-03-18
BXC-4896 google console for DCR

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4896](https://unclibrary.atlassian.net/browse/BXC-4896)

- add meta tag to header

### DIFF
--- a/static/js/vue-cdr-access/src/components/header/headerHome.vue
+++ b/static/js/vue-cdr-access/src/components/header/headerHome.vue
@@ -59,6 +59,7 @@
                 </div>
             </div>
         </div>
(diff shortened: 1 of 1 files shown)


static/js/vue-cdr-access/src/components/header/headerHome.vue
```diff
@@ -59,6 +59,7 @@
                 </div>
             </div>
         </div>
+      <meta name="google-site-verification" content="WO17zKdoDATg4FxnQafjjRNowoM-TDP4hEfb_H--mdo" />
```
bbpennel, line 62: I think we may actually want to that this line to this file in the "head" section with the other meta tags:
https://github.com/UNC-Libraries/box-c/blob/main/static/js/vue-cdr-access/index.html

bbpennel (approved):

PR #1906, opened 2025-03-24
Bxc 4846 export csv service test

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4846

### DIFF
--- a/web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/ExportCsvServiceTest.java
+++ b/web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/ExportCsvServiceTest.java
@@ -38,6 +38,7 @@ import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.openMocks;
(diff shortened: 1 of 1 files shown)


web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/ExportCsvServiceTest.java
```diff
@@ -0,0 +1,167 @@
...
+        when(object.getResourceType()).thenReturn(ResourceType.File.name());
+        when(object.getId()).thenReturn(pid.getId());
+        when(object.getTitle()).thenReturn("a good title");
+        when(object.getAncestorNames()).thenReturn("/Root/Unit/Collection/Folder/Work/File");
+        when(object.getContentStatus()).thenReturn(List.of(FacetConstants.CONTENT_DESCRIBED));
+        when(object.getAncestorPathFacet()).thenReturn(facet);
+        when(facet.getHighestTier()).thenReturn(1);
+        when(facet.getHighestTierNode()).thenReturn(facetNode);
+        when(facetNode.getSearchKey()).thenReturn(WORK_ID);
+        exportCsvService.streamCsv(pids, agent, outputStream);
```
bbpennel, line 138: We should probably assert something about this, other than that it runs without errors. We could do similar checks to what are in `testStreamCsv`, stating `never()` for cases where it should not be calling methods, like counting children I think.

bbpennel (changes requested):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/processing/ExportCsvServiceTest.java
sharonluong: good point!

bbpennel (approved):

PR #1918, opened 2025-04-10
BXC-4908 aspace metadata export csv

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4908](https://unclibrary.atlassian.net/browse/BXC-4908)

- `ExportMetadataCsvService`: `exportCsv` method to create a csv of metadata that will be submitted to DOMino for Aspace DO generation, helper methods
- `ref_id` is a placeholder, we still need to figure out how to get boxc to store ref_ids

### DIFF
--- a/operations/src/test/java/edu/unc/lib/boxc/operations/impl/metadata/ExportMetadataCsvServiceTest.java
+++ b/operations/src/test/java/edu/unc/lib/boxc/operations/impl/metadata/ExportMetadataCsvServiceTest.java
@@ -1,0 +1,152 @@
+package edu.unc.lib.boxc.operations.impl.metadata;
+
+import edu.unc.lib.boxc.auth.api.models.AgentPrincipals;
+import edu.unc.lib.boxc.auth.api.services.AccessControlService;
+import edu.unc.lib.boxc.auth.fcrepo.models.AccessGroupSetImpl;
+import edu.unc.lib.boxc.auth.fcrepo.models.AgentPrincipalsImpl;
+import edu.unc.lib.boxc.model.api.DatastreamType;
+import edu.unc.lib.boxc.model.api.ResourceType;
+import edu.unc.lib.boxc.model.api.ids.PID;
+import edu.unc.lib.boxc.model.fcrepo.ids.PIDs;
+import edu.unc.lib.boxc.search.api.models.ContentObjectRecord;
+import edu.unc.lib.boxc.search.solr.models.ContentObjectSolrRecord;
+import edu.unc.lib.boxc.search.solr.models.DatastreamImpl;
+import edu.unc.lib.boxc.search.solr.responses.SearchResultResponse;
+import edu.unc.lib.boxc.search.solr.services.SolrSearchService;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVRecord;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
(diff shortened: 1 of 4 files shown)


operations/src/test/java/edu/unc/lib/boxc/operations/impl/metadata/ExportMetadataCsvServiceTest.java
```diff
@@ -67,25 +75,78 @@ void closeService() throws Exception {
...
+
+        var resultPath = csvService.exportCsv(asPidList(COLLECTION_UUID), agent);
+        var csvRecords = parseCsv(ExportDominoMetadataService.CSV_HEADERS, resultPath);
+        assertContainsEntry(csvRecords, UUID1, "ref_id", "Work 1");
+        assertContainsEntry(csvRecords, UUID2, "ref_id", "Work 2");
         assertNumberOfEntries(2, csvRecords);
     }
+    
+    @Test
+    public void recordNotFoundTest() {
```
bbpennel, line 94: Is this testing that the parent is not found? It might be good to clarify that in the name.

It shouldn't be expecting an NPE to be thrown, it looks like `NotFoundException` is the expected exception for the parent being null. The NPE is actually being thrown by the mocking setup rather than running exportCsv. We don't actually need to mock anything to have the parent come back as null, but for clarity we could mock `solrSearchService.getObjectById` since that is how the parent is retrieved.

Also, `assertTrue(Files.notExists(csvPath));` will never be reached since `exportCsv` is expected to throw an exception. Since this is using a temp file for the CSV, we can't predict what the file path will be either, so we can remove the assertion.

operations/src/test/java/edu/unc/lib/boxc/operations/impl/metadata/ExportMetadataCsvServiceTest.java
```diff
@@ -67,25 +75,78 @@ void closeService() throws Exception {
...
+                    any()).getResultList());
+            var csvPath = csvService.exportCsv(asPidList(COLLECTION_UUID), agent);
+            assertTrue(Files.notExists(csvPath));
+        });
+    }
+
+    @Test
+    public void userNoPermissionTest() {
+        assertThrows(AccessRestrictionException.class, () -> {
+            doThrow(new AccessRestrictionException())
```
bbpennel, line 106: Just a safety precaution, but I would generally suggest doing setup like this outside of the assertThrows, otherwise we can have cases where the setup accidental throws an exception (like in the previous test).

Also, the `Files.notExists` assertion is unreachable here again

operations/src/test/java/edu/unc/lib/boxc/operations/impl/metadata/ExportMetadataCsvServiceTest.java
```diff
@@ -67,25 +75,78 @@ void closeService() throws Exception {
...
+        });
+    }
+
+    @Test
+    public void parentWithNoChildrenTest() {
+        var collectionRecord = makeRecord(COLLECTION_UUID, ADMIN_UNIT_UUID, ResourceType.Collection,
+                "Collection", null, null, null);
+        mockParentResults(collectionRecord);
+
+        assertThrows(NullPointerException.class, () -> {
```
bbpennel, line 120: Same feedback as `recordNotFoundTest`, except i think for the case of no children we don't expect an exception to be thrown, just an empty CSV to be returned.

operations/src/test/java/edu/unc/lib/boxc/operations/impl/metadata/ExportMetadataCsvServiceTest.java
```diff
@@ -67,25 +75,78 @@ void closeService() throws Exception {
...
+            assertTrue(Files.notExists(csvPath));
+        });
+    }
+
+    @Test
+    public void invalidIdTypeTest() throws Exception {
+        var workRecord1 = makeWorkRecord(UUID1, "Work 1");
+        mockParentResults(workRecord1);
+
+        try {
```
bbpennel, line 133: For consistency, use `var exception = assertThrows(...` then you can get the message from the returned exception

bbpennel (approved):

PR #1925, opened 2025-05-05
BXC-4934 - Prevent hanging on FITS runs

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4934

* Add a timeout for CLI FITS runs, read output in parallel to prevent deadlock, update camel to most recent version

### DIFF
--- a/common-utils/src/main/java/edu/unc/lib/boxc/common/util/CLIUtil.java
+++ b/common-utils/src/main/java/edu/unc/lib/boxc/common/util/CLIUtil.java
@@ -1,0 +1,97 @@
+package edu.unc.lib.boxc.common.util;
+
+import edu.unc.lib.boxc.common.errors.CommandException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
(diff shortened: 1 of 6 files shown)


common-utils/src/main/java/edu/unc/lib/boxc/common/util/CLIUtil.java
```diff
@@ -0,0 +1,100 @@
...
+            return List.of(outputConsumer.getContent(), errorConsumer.getContent());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new CommandException("Command interrupted", command, "", e);
+        } catch (IOException e) {
+            throw new CommandException("Command failed to execute", command, "", e);
+        }
+    }
+
+    private static String combinedOutput(StreamConsumer outputConsumer, StreamConsumer errorConsumer) {
```
sharonluong, line 61: so this method is only used when the command exits with errors? if the above method returns a list of both, does it only add them together here since the error consumer is then not null?

bbpennel: I was originally thinking I might have null checks in here, but didn't end up adding them since I think both streams will always have values, even if it's just an empty string, which should format fine. So it's mostly just making the error construction slightly shorter.

sharonluong: I guess what i'm trying to ask is, since in `executeCommand` it's a list of the outputs, this method adds them together just so it's one argument in the `CommandException` method?

sharonluong (approved):

PR #1936, opened 2025-05-16
Add admin UI form for updating aspace ref id

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4931


static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -192,6 +192,10 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 			items["editDescription"] = {name : 'Edit Description'};
 		}
 
+		if (metadata.type === 'Work' && $.inArray('editDescription', metadata.permissions) != -1) {
```
bbpennel, line 195: There's actually a new permission for this:
https://github.com/UNC-Libraries/box-c/blob/main/auth-api/src/main/java/edu/unc/lib/boxc/auth/api/Permission.java#L29
`editAspaceProperties`
I believe it should already be getting returned in the admin UI.

bbpennel (changes requested):

PR #1939, opened 2025-05-27
UI for bulk importing aspace ref ids

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-4944


bbpennel:
When testing this, I kept wanting to export bulk ref ids for the collection I was currently browsing, but I'd have to navigate up a level and then check the collection from its parent to export it. Could we add it as an option in the gear menu for appropriate object types under the Export submenu?

bbpennel (approved):

PR #1942, opened 2025-05-29
BXC-4959 update deposit pipeline to receive refIds

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4959](https://unclibrary.atlassian.net/browse/BXC-4959)

- update `populateAIPProperties` to add the `CdrAspace.RefId` property if it is present
- update `cdr-schema.ttl` to only allow the `CdrAspace.RefId` property to works

### DIFF
--- a/deposit-app/src/main/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJob.java
+++ b/deposit-app/src/main/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJob.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Set;
 
 import edu.unc.lib.boxc.model.api.exceptions.NotFoundException;
+import edu.unc.lib.boxc.model.api.rdf.CdrAspace;
 import edu.unc.lib.boxc.operations.impl.altText.AltTextUpdateService;
 import edu.unc.lib.boxc.operations.jms.altText.AltTextUpdateRequest;
 import org.apache.http.HttpStatus;
@@ -792,6 +793,9 @@ public class IngestContentObjectsJob extends AbstractDepositJob {
         }
         if (dResc.hasProperty(Cdr.streamingType)) {
             aResc.addProperty(Cdr.streamingType, dResc.getProperty(Cdr.streamingType).getString());
+        }
+        if (dResc.hasProperty(CdrAspace.refId)) {
+            aResc.addProperty(CdrAspace.refId, dResc.getProperty(CdrAspace.refId).getString());
         }
     }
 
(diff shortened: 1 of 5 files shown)


deposit-app/src/main/java/edu/unc/lib/boxc/deposit/fcrepo4/IngestContentObjectsJob.java
```diff
@@ -793,6 +794,9 @@ private void populateAIPProperties(Resource dResc, Resource aResc) {
         if (dResc.hasProperty(Cdr.streamingType)) {
             aResc.addProperty(Cdr.streamingType, dResc.getProperty(Cdr.streamingType).getString());
         }
+        if (dResc.hasProperty(CdrAspace.refId)) {
```
bbpennel, line 797: we should update one of the unit tests to verify this works

deposit-app/src/main/resources/cdr-schema.ttl
```diff
@@ -184,4 +185,11 @@ cdr-acl:unitOwner a rdfs:Property ;
     rdfs:label "Unit Owner" ;
     rdfs:domain cdr:AdminUnit ;
     rdfs:range xsd:string ;
-    rdfs:comment "User granted ownership of an administrative unit. Has all access and administrative permissions, as well as permission to permanently destroy objects, create collections, and assign staff permissions. Applies to cdr:AdminUnit objects." .
\ No newline at end of file
+    rdfs:comment "User granted ownership of an administrative unit. Has all access and administrative permissions, as well as permission to permanently destroy objects, create collections, and assign staff permissions. Applies to cdr:AdminUnit objects." .
+
+###### Aspace Properties ######
+cdr-aspace:refId a rdfs:Property ;
```
bbpennel, line 191: We should add a unit test to verify that the validator permits ref ids on works, and another to verify that it rejects ref ids on a non-work

bbpennel (changes requested):

deposit-app/src/test/java/edu/unc/lib/boxc/deposit/validate/ValidateContentModelJobTest.java
```diff
@@ -564,4 +565,43 @@ public void invalidChildAclsTest() {
...
+
+        job.closeModel();
+
+        job.run();
+    }
+
+    @Test
+    public void fileInvalidAspaceRefIdTest() {
+        Assertions.assertThrows(InvalidAssignmentException.class, () -> {
+            doThrow(new InvalidAssignmentException()).when(aclValidator).validate(any(Resource.class));
```
bbpennel, line 593: Since this test doesn't involve access control, we shouldn't need this. Ideally the job should throw an error without needing to mock anything in this case. Judging by the other tests it should be JobFailedException

PR #1947, opened 2025-06-12
BXC-4962 - PDF error handling and embedded jp2s

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4962

* App jp2 library so that pdfs containing them will render. 
* Handle ioexceptions during PDF rendering to reduce verbosity and prevent retries. This is to deal with the other exceptions that occurred during PDF image generation
* Update setup-java gh action and specify a distribution, since GH actions was randomly failing to get a JDK

### DIFF
--- a/services-camel-app/src/test/java/edu/unc/lib/boxc/services/camel/images/PdfImageProcessorTest.java
+++ b/services-camel-app/src/test/java/edu/unc/lib/boxc/services/camel/images/PdfImageProcessorTest.java
@@ -76,9 +76,10 @@ public class PdfImageProcessorTest {
         message.setHeader(CdrBinaryPath, "nonexistent.pdf");
         message.setHeader(CdrBinaryMimeType, "application/pdf");
 
-        // No exception should be thrown
+        // An IOException is thrown internally, but we are demonstrating that it does not bubble up
         processor.process(exchange);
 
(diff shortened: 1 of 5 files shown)


services-camel-app/src/test/java/edu/unc/lib/boxc/services/camel/images/PdfImageProcessorTest.java
```diff
@@ -67,6 +70,20 @@ public void testProcessPdf() throws Exception {
         assertTrue(Files.size(outputFile) > 0, "Generated TIFF file should not be empty");
     }
 
+    @Test
+    public void testLoadThrowsIOException() throws Exception {
+        // Set a non-existent PDF path
+        message.setHeader(CdrBinaryPath, "nonexistent.pdf");
+        message.setHeader(CdrBinaryMimeType, "application/pdf");
+
+        // No exception should be thrown
```
sharonluong, line 79: i guess i am confused because the title says `testLoadThrowsIOException` but then here it says no exception should be thrown. maybe something more detail is needed?

bbpennel: loading the file throws an exception, but the test is demonstrating that this kind of exception gets handled internally and doesn't bubble up. I can add a longer comment

PR #1949, opened 2025-06-19
Bxc 4943 bulk ref ids export

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4943

This PR does the following:
- adds controller endpoint for exporting ref ID csv requests
- adds an exporter class which makes the CSV from searching through the work children associated with the PID, or the object itself if it's a work
- moves shared code to an Exporter Util
- moves shared code to util and helper classes for the tests
- adds get method for Hook ID on the content object record


web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/MvcTestHelpers.java
```diff
@@ -43,4 +49,19 @@ public static SearchResultResponse createSearchResponse(List<ContentObjectRecord
...
+
+    public static List<CSVRecord> parseCsvResponse(MockHttpServletResponse response, String[] headers) throws Exception {
+        List<CSVRecord> csvList = new ArrayList<>();
+
+        CSVFormat format = CSVFormat.DEFAULT.builder()
+                .setHeader(headers)
+                .setSkipHeaderRecord(true)
+                .get();
+
+//        var format = CSVFormat.DEFAULT.withFirstRecordAsHeader();
```
bbpennel, line 61: Can remove this line

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/BulkRefIdCsvExporter.java
```diff
@@ -0,0 +1,125 @@
...
+import static java.util.Arrays.asList;
+
+public class BulkRefIdCsvExporter {
+    private static final Logger log = LoggerFactory.getLogger(BulkRefIdCsvExporter.class);
+    private static final int DEFAULT_PAGE_SIZE = 10000;
+    public static final String PID_HEADER = "pid";
+    public static final String REF_ID_HEADER = "refId";
+    public static final String HOOK_ID_HEADER = "hookId";
+    public static final String TITLE_HEADER = "title";
+    public static final String[] CSV_HEADERS = new String[] {PID_HEADER, REF_ID_HEADER, HOOK_ID_HEADER, TITLE_HEADER};
```
bbpennel, line 42: Would probably be best to name this EXPORT_CSV_HEADERS, similar to how to renamed the other one IMPORT_CSV_HEADERS in this PR

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/utils/ExporterUtil.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+public class ExporterUtil {
```
bbpennel, line 16: I'd suggest naming this something like ExporterTestUtil to make it clear that it is only used in testing, since that's not obvious at the moment when it's being used in the tests.

Also, we should add javadoc to this class

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/AspaceRefIdControllerTest.java
```diff
@@ -158,7 +178,7 @@ public void importRefIdsSuccess() throws Exception {
```
bbpennel, line 178: It would be good to update one of the existing import tests to pass in a CSV in the export format to verify it works in a round trip

bbpennel (changes requested):

PR #1958, opened 2025-07-11
Bxc 4961 state unmodified bug

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-4961

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/EditTitleController.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/EditTitleController.java
@@ -3,6 +3,7 @@ package edu.unc.lib.boxc.web.services.rest.modify;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/EditTitleController.java
```diff
@@ -51,6 +52,9 @@ public ResponseEntity<Object> editLabel(@PathVariable("id") String id, @RequestP
 
             if (e instanceof AuthorizationException || e instanceof AccessRestrictionException) {
                 return new ResponseEntity<>(result, HttpStatus.FORBIDDEN);
+            } else if (e instanceof StateUnmodifiedException) {
```
bbpennel, line 55: since this isn't really an error, it probably doesn't make sense to return the "error" field. I think making it into its own catch would work, like
```
} catch (StateUnmodifiedException e) {
    log.info("No changes were made to {}", pid.getRepositoryPath());
    result.put("status", "unchanged");
} catch (Exception e) {
   // previous error handling stuff
```

bbpennel (changes requested):

PR #2122, opened 2026-04-14
BXC-5610 update logger to include job failed exception details

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-5610

### DIFF
--- a/deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/ValidateFileAvailabilityJob.java
+++ b/deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/ValidateFileAvailabilityJob.java
@@ -79,7 +79,6 @@ public class ValidateFileAvailabilityJob extends AbstractDepositJob {
                 sbInvalid.append(" - ").append(file).append("\n");
             }
 
-            var string = sbInvalid.toString();
             failJob("Deposit references invalid files", (sbInvalid.toString()));
         }
     }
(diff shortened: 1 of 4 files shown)


deposit-app/src/main/java/edu/unc/lib/boxc/deposit/validate/ValidateFileAvailabilityJob.java
```diff
@@ -79,6 +79,7 @@ public void runJob() {
                 sbInvalid.append(" - ").append(file).append("\n");
             }
 
+            var string = sbInvalid.toString();
```
bbpennel: This line isn't being used

deposit-app/src/test/java/edu/unc/lib/boxc/deposit/validate/ValidateFileAvailabilityJobTest.java
```diff
@@ -132,7 +132,12 @@ public void invalidSourceTest() {
 
             job.closeModel();
 
-            job.run();
+            try {
+                job.run();
+            } catch (JobFailedException e) {
+                assertTrue(e.getDetails().contains("missing.pdf"));
+                throw e;
+            }
```
bbpennel: Rather than doing a try both here, we can get the exception back directly by doing  `var e = Assertions.assertThrows(JobFailedException.class, () -> {` and then inspecting the exception after the assertThrows block.

bbpennel (changes requested):

deposit-app/src/test/java/edu/unc/lib/boxc/deposit/validate/ValidateFileAvailabilityJobTest.java
sharonluong: ah that makes sense! i had copied it from another test but it did seem unnecessary. thanks!

bbpennel (approved):

PR #2128, opened 2026-04-27
Bxc 5599 mg desc api

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-5599

### DIFF
--- a/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MachineGeneratedContentService.java
+++ b/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MachineGeneratedContentService.java
@@ -1,10 +1,10 @@
-package edu.unc.lib.boxc.indexing.solr.utils;
+package edu.unc.lib.boxc.search.solr.services;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import edu.unc.lib.boxc.indexing.solr.exception.IndexingException;
 import edu.unc.lib.boxc.model.api.DatastreamType;
+import edu.unc.lib.boxc.model.api.exceptions.RepositoryException;
 import edu.unc.lib.boxc.model.api.ids.PID;
 import edu.unc.lib.boxc.model.fcrepo.services.DerivativeService;
 import org.slf4j.Logger;
@@ -85,7 +85,7 @@ public class MachineGeneratedContentService {
         try {
             return MAPPER.readTree(mgdJson);
         } catch (JsonProcessingException e) {
-            throw new IndexingException("Unable to deserialize machine generated JSON", e);
+            throw new RepositoryException("Unable to deserialize machine generated JSON", e);
         }
     }
 
@@ -135,6 +135,32 @@ public class MachineGeneratedContentService {
         }
         JsonNode scoreNode = mgdNode.path(RESULT_FIELD).path(MG_RISK_SCORE_FIELD);
         return scoreNode.isMissingNode() ? null : scoreNode.asInt();
+    }
+
+    /**
+     * Extracts the review assessment from the machine generated description JSON, if it exists.
+     * @param mgdNode the machine generated description JSON root node
+     * @return review assessment if it exists, otherwise null
+     */
+    public JsonNode extractReviewAssessment(JsonNode mgdNode) {
+        if (mgdNode == null) {
+            return null;
+        }
+        JsonNode scoreNode = mgdNode.path(RESULT_FIELD).path(MG_REVIEW_ASSESS_FIELD);
+        return scoreNode.isMissingNode() ? null : scoreNode;
+    }
+
+    /**
+     * Extracts the safety assessment from the machine generated description JSON, if it exists.
+     * @param mgdNode the machine generated description JSON root node
+     * @return safety assessment if it exists, otherwise null
+     */
+    public JsonNode extractSafetyAssessment(JsonNode mgdNode) {
+        if (mgdNode == null) {
+            return null;
+        }
+        JsonNode scoreNode = mgdNode.path(RESULT_FIELD).path(MG_SAFETY_ASSESS_FIELD);
+        return scoreNode.isMissingNode() ? null : scoreNode;
     }
 
     /**
--- a/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MachineGeneratedContentService.java
+++ b/search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MachineGeneratedContentService.java
@@ -146,8 +146,8 @@ public class MachineGeneratedContentService {
         if (mgdNode == null) {
             return null;
         }
-        JsonNode scoreNode = mgdNode.path(RESULT_FIELD).path(MG_REVIEW_ASSESS_FIELD);
-        return scoreNode.isMissingNode() ? null : scoreNode;
+        JsonNode assessmentNode = mgdNode.path(RESULT_FIELD).path(MG_REVIEW_ASSESS_FIELD);
+        return assessmentNode.isMissingNode() ? null : assessmentNode;
     }
 
     /**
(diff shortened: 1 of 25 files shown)


search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MachineGeneratedContentService.java
```diff
@@ -137,6 +137,32 @@ public Integer extractRiskScore(JsonNode mgdNode) {
...
+    /**
+     * Extracts the review assessment from the machine generated description JSON, if it exists.
+     * @param mgdNode the machine generated description JSON root node
+     * @return review assessment if it exists, otherwise null
+     */
+    public JsonNode extractReviewAssessment(JsonNode mgdNode) {
+        if (mgdNode == null) {
+            return null;
+        }
+        JsonNode scoreNode = mgdNode.path(RESULT_FIELD).path(MG_REVIEW_ASSESS_FIELD);
```
bbpennel: The name should either be more generic or should match the data be retrieved.

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MachineGeneratedContentService.java
```diff
@@ -137,6 +137,32 @@ public Integer extractRiskScore(JsonNode mgdNode) {
...
+    /**
+     * Extracts the safety assessment from the machine generated description JSON, if it exists.
+     * @param mgdNode the machine generated description JSON root node
+     * @return safety assessment if it exists, otherwise null
+     */
+    public JsonNode extractSafetyAssessment(JsonNode mgdNode) {
+        if (mgdNode == null) {
+            return null;
+        }
+        JsonNode scoreNode = mgdNode.path(RESULT_FIELD).path(MG_SAFETY_ASSESS_FIELD);
```
bbpennel: The name should either be more generic or should match the data be retrieved.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/MachineGeneratedSearchController.java
```diff
@@ -0,0 +1,97 @@
...
+            response.put("results", List.of());
+            response.put("errorMessage", NO_RESULTS);
+            return new ResponseEntity<>(response, HttpStatus.OK);
+        }
+
+        List<Map<String, Object>> results = new ArrayList<>(resultResponse.getResultList().size());
+        for (ContentObjectRecord metadata : resultResponse.getResultList()) {
+
+            Map<String, Object> data = new HashMap<>();
+            data.put("id", metadata.getId());
```
bbpennel: it's clunky, but I'd recommend using the constants like `SearchFieldKey.ID.getUrlParam()` for all of these field keys when they exist, particularly since this controller depends on generateSearchRequest which requires us to round trip consistent field names. I don't think any of the keys here are wrong, just good to keep them linked up

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/MachineGeneratedSearchController.java
```diff
@@ -0,0 +1,97 @@
...
+    private MachineGeneratedContentService machineGeneratedContentService;
+
+    @RequestMapping(value = "/machineGeneratedSearch/{parentId}")
+    public @ResponseBody ResponseEntity<Object> search(@PathVariable("parentId") String pidString, HttpServletRequest request) {
+        var pid = PIDs.get(pidString);
+        accessControlService.assertHasAccess("Insufficient permissions to access machine generated metadata for object " + pid,
+                pid, getAgentPrincipals().getPrincipals(), Permission.viewHidden);
+
+        SearchRequest searchRequest = generateSearchRequest(request);
+        SearchState searchState = searchRequest.getSearchState();
```
bbpennel, line 68: I believe we still need to filter to just File objects, and to just children of the parent id. There should be examples of both of these kinds of filters.

bbpennel (changes requested):

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/MachineGeneratedSearchController.java
sharonluong: ahhh that was the thing i wanted to ask you about but forgot yesterday. thanks!

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/MachineGeneratedSearchController.java
```diff
@@ -52,6 +56,11 @@ public class MachineGeneratedSearchController extends AbstractSolrSearchControll
         SearchRequest searchRequest = generateSearchRequest(request);
         SearchState searchState = searchRequest.getSearchState();
         searchState.setResultFields(MG_RESULT_FIELDS);
+        // filter to FileObjects
+        searchState.setFacet(new GenericFacet(SearchFieldKey.RESOURCE_TYPE.name(), ResourceType.File.name()));
+        // filter to children of parent ID
+        CutoffFacet cutoff = new CutoffFacetImpl(SearchFieldKey.ANCESTOR_PATH.name(), "1," + pid.getUUID());
```
bbpennel: Since `queryLayer.performSearch` is being used, I think we can limit to children of a particular container by calling the following:
```
searchRequest.setRootPid(pid); // Only children of this object will be returned
searchRequest.setApplyCutoffs(false); // Filters to all children, not just immediate
```
If we add a cutoff facet directly like in the current code, we'd need to figure out what the correct number prefix was, since "1," would only match if the id were the first id in the hierarchy (meaning it was the root of the repository).

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/MachineGeneratedControllerTest.java
```diff
@@ -70,16 +96,27 @@ public void init() {
...
+        searchState = new SearchState();
+        searchState.setFacetsToRetrieve(Collections.emptyList());
 
         TestHelper.setContentBase("http://localhost:48085/rest");
 
         when(searchStateFactory.createSearchState(any())).thenReturn(searchState);
+        doAnswer((Answer<Void>) invocation -> null)
+                .when(searchResultResponseDecorator).populateThumbnailUrls(any(), any());
+        doAnswer((Answer<Void>) invocation -> null)
+                .when(searchResultResponseDecorator).retrieveFacets(any(), any());
```
bbpennel: I don't think these are doing anything? The test passes without them

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/MachineGeneratedControllerTest.java
```diff
@@ -6,26 +6,41 @@
...
 import edu.unc.lib.boxc.search.solr.services.SearchStateFactory;
+import edu.unc.lib.boxc.search.solr.services.SetFacetTitleByIdService;
+import edu.unc.lib.boxc.search.solr.utils.SearchStateUtil;
 import edu.unc.lib.boxc.web.common.services.SolrQueryLayerService;
+import edu.unc.lib.boxc.web.common.utils.SerializationUtil;
 import edu.unc.lib.boxc.web.services.rest.exceptions.RestResponseEntityExceptionHandler;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
```
bbpennel: There are a bunch of unused imports in this class

bbpennel (changes requested):

web-common/src/main/java/edu/unc/lib/boxc/web/common/controllers/AbstractSolrSearchController.java
```diff
@@ -136,4 +136,12 @@ protected Map<String, Object> getResults(SearchResultResponse resp, String query
 
         return results;
     }
+
+//    public void setChildrenCountService(ChildrenCountService childrenCountService) {
```
bbpennel: Is this code not needed now?

bbpennel (approved):

PR #2132, opened 2026-04-28
BXC-5549 setup camel workflow for pdf4u

krwong:
[https://unclibrary.atlassian.net/browse/BXC-5549](https://unclibrary.atlassian.net/browse/BXC-5549)

- add `PDF_ACCESS_COPY` to DatastreamType
- add pdf4u to services-camel's `pom.xml` and `logback.xml`
- add pdf4u and pdf derivative processors and routers
- update `DestroyDerivativesRouter`, `EnhancementsRouter`, and `service-context.xml`
- add and update tests and test resources


operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+ * Service for generating a derivative PDF with OCR
+ * @author krwong
+ */
+public class PdfDerivativeService {
+    private static final Logger log = LoggerFactory.getLogger(PdfDerivativeService.class);
+
+    private SolrSearchService solrSearchService;
+    private RepositoryObjectLoader repositoryObjectLoader;
+
+    public Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
```
bbpennel: This is fine as default value (although I'm not sure if it needs to be public), but we should inject the value of the `${services.tempDirectory}` property in the spring context file like we are doing for most of the other processors in the camel app.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,213 @@
...
+ * @author krwong
+ */
+public class PdfDerivativeService {
+    private static final Logger log = LoggerFactory.getLogger(PdfDerivativeService.class);
+
+    private SolrSearchService solrSearchService;
+    private RepositoryObjectLoader repositoryObjectLoader;
+
+    public Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
+    public Path tmpFilesDir = tmpDir.resolve("pdf4u");
```
bbpennel, line 50: This subfolder will have to get setup after tmpDir is set. I'd suggest using an `init` method to do that and to do the call to `initializeTempImageFilesDir` that is currently in the constructor in the init method too. Spring might automatically find a method named "init", I can't remember for sure

model-api/src/main/java/edu/unc/lib/boxc/model/api/DatastreamType.java
```diff
@@ -26,6 +26,7 @@ public enum DatastreamType {
     MD_DESCRIPTIVE_HISTORY("md_descriptive_history", Constants.TEXT_XML, "xml", METADATA_CONTAINER, INTERNAL),
     MD_EVENTS("event_log", "application/n-triples", "nt", METADATA_CONTAINER, INTERNAL),
     ORIGINAL_FILE("original_file", null, null, DATA_FILE_FILESET, INTERNAL),
+    PDF_ACCESS_COPY("pdf", "application/pdf", "pdf", null, EXTERNAL),
```
bbpennel: I don't think this datastream type is needed anymore in that current setup

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+        var workPidString = request.getWorkPid();
+        var workPid = PIDs.get(workPidString);
+        var agent = request.getAgent();
+
+        var inputFilePath = prepareTempPath(workPidString + "_input", ".txt");
+        try {
+            var parentRec = getParentRecord(workPid, agent);
+            assertParentRecordValid(workPid, parentRec);
+
+            BufferedWriter writer = new BufferedWriter(new FileWriter(inputFilePath.toFile()));
```
bbpennel: use a `try (...)` block for ensuring this writer gets closed rather than explicitly calling close

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+        if (!Files.exists(path)) {
+            Files.createDirectories(path);
+        }
+    }
+
+    /**
+     * Create temporary file path and delete temporary file if it already exists
+     * @return tmpImageFilesDirectoryPath
+     */
+    private Path prepareTempPath(String fileName, String extension) throws Exception {
```
bbpennel:
```suggestion
    private Path prepareTempPath(String fileName, String extension) {
```
We should use the same approach for getting a temp path that we just added in pdf4u. And when we switch to that, then we wouldn't need to have this method throw any exceptions

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+        if (!supportsMemberOrdering(resourceType)) {
+            throw new InvalidOperationForObjectType(formatUnsupportedMessage(pid, resourceType));
+        }
+    }
+
+    /**
+     * Create tmp pdf4u files directory for temporary files
+     * @return tmpImageFilesDirectoryPath
+     */
+    private void initializeTempImageFilesDir() throws Exception {
```
bbpennel:
```suggestion
    private void initializeTempImageFilesDir() throws IOException {
```

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.List;
+
+import static edu.unc.lib.boxc.operations.api.order.MemberOrderHelper.formatUnsupportedMessage;
+import static edu.unc.lib.boxc.operations.api.order.MemberOrderHelper.supportsMemberOrdering;
```
bbpennel: We should use different methods here since these are both member order specific.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+            throw new Exception("Failed to generate input txt file for " + workPid, e);
+        }
+    }
+
+    /**
+     * Get path for transcript file(s)
+     * @param request PdfRequest
+     * @return .txt path to transcript files
+     */
+    public String getTranscriptFiles(PdfRequest request) throws Exception {
```
bbpennel:
```suggestion
    public Path createTranscriptListFile(PdfRequest request) throws IOException {
```
I find the method name/description a little confusing, with the plural I'd expect a list of files to get returned. It looks like this is creating a file that will contain a list of all the script paths. I also would suggest having it return a Path, turning it into a String is only needed for adding it as a parameter to the command, which is separate from creating the file

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Get text type from boxctron's alt text review
+     * @param request PdfRequest
+     * @return textType
+     */
+    public String getTextTypes(PdfRequest request) {
+        //TODO: get text type from the alt text review
+
+        return "HANDWRITTEN-PRINT";
```
bbpennel, line 176: Does pdf4u expect one value for all the files being added or a value for each one?

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,213 @@
...
+
+    // Query for all immediate children/members of the specified record, in default sort order
+    private List<ContentObjectRecord> getChildrenRecords(ContentObjectRecord parentRec, AgentPrincipals agent) {
+        SearchState searchState = new SearchState();
+        searchState.setIgnoreMaxRows(true);
+        searchState.setRowsPerPage(DEFAULT_PAGE_SIZE);
+        CutoffFacet selectedPath = parentRec.getPath();
+        searchState.addFacet(selectedPath);
+        searchState.setSortType("default");
+        searchState.setResultFields(FILE_REQUEST_FIELDS);
```
bbpennel, line 188: We should also filter the children to just images, we can use the `FILE_FORMAT_CATEGORY` field for that, or filter to files that have a jp2 datastream which i believe we are using as a proxy for finding displayable images elsewhere. Only issue with the latter is that the user would need to make sure to wait until after all the derivatives were generated before calling this workflow.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+            }
+        }
+    }
+
+    /**
+     * Get path for input file(s)
+     * @param request PdfRequest
+     * @return .txt path to input files
+     */
+    public String getInputFiles(PdfRequest request) throws Exception {
```
bbpennel:
```suggestion
    public Path createInputListFile(PdfRequest request) {
```
Same comments as for the getTranscriptFiles method.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+
+        var inputFilePath = prepareTempPath(workPidString + "_input", ".txt");
+        try {
+            var parentRec = getParentRecord(workPid, agent);
+            assertParentRecordValid(workPid, parentRec);
+
+            BufferedWriter writer = new BufferedWriter(new FileWriter(inputFilePath.toFile()));
+            List<ContentObjectRecord> children = getChildrenRecords(parentRec, agent);
+            for (var child : children) {
+                var original = child.getDatastreamObject(DatastreamType.ORIGINAL_FILE.getId());
```
bbpennel: I am guessing you need the actual absolute path to the file to pass into pdf4u? The value in the datastream field is the filename originally assigned to the file (used when downloading) but not the name/path of the file on disk. The simplest way to get the real path would be to use `RepositoryObjectLoader.getBinaryObject(...)` and then call getContentURI on the binary object. You can get the pid of the original_file datastream by using the DatastreamPids class helper methods, passing in the PID of the file object. We might not actually need to return the solr datastream field, but solr is still useful for filtering children.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }
+
+    /**
+     * Get path for transcript file(s)
+     * @param request PdfRequest
+     * @return .txt path to transcript files
+     */
+    public String getTranscriptFiles(PdfRequest request) throws Exception {
+        //TODO: get transcript files from the alt text review
```
bbpennel: Getting a file path to a transcript is not straight forward since it can come from two places ([see here](https://github.com/UNC-Libraries/box-c/blob/main/indexing-solr/src/main/java/edu/unc/lib/boxc/indexing/solr/filter/SetContentDescriptionMetadataFilter.java#L103) if you want to see how it gets retrieved), but you can get the value of the transcript directly from solr in the transcript field and write it to a temp file.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/PdfDerivativeService.java
```diff
@@ -0,0 +1,213 @@
...
+
+    public PdfDerivativeService() {
+        try {
+            initializeTempImageFilesDir();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public Path generatePdfDerivative(PdfRequest request) throws Exception {
```
bbpennel: For the most part we don't want to throw `Exception` in production code since its not very informative (that would apply to all the methods in this class). I'd suggest either throwing the specific exception classes (rather than rethrowing as an Exception), or rethrow as something like a ServiceException

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/PdfDerivativeProcessor.java
```diff
@@ -0,0 +1,148 @@
...
+ */
+public class PdfDerivativeProcessor implements Processor {
+    private static final Logger log = LoggerFactory.getLogger(PdfDerivativeProcessor.class);
+
+    private AccessControlService aclService;
+    private RepositoryObjectLoader repositoryObjectLoader;
+    private PdfDerivativeService pdfDerivativeService;
+    private String derivativeBasePath;
+
+    private static final Pattern MIMETYPE_PATTERN = Pattern.compile("^(image.*$|application.*?(photoshop|psd|pdf)$)");
```
bbpennel: I don't think this or DISALLOWED_PATTERN are used anymore since its now an ondemand service. You could move them into PdfDerivativeService for mimetype filtering of solr results. I think we do that in `AccessCopiesService`

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/PdfDerivativeProcessor.java
```diff
@@ -0,0 +1,148 @@
...
+            return "No mimetype provided for object " + pid;
+        }
+
+        if (!MIMETYPE_PATTERN.matcher(mimetype).matches()) {
+            return "File type " + mimetype + " on object " + pid + " is not applicable for pdf derivatives";
+        }
+
+        if (DISALLOWED_PATTERN.matcher(mimetype).matches()) {
+            return "File type " + mimetype + " on object " + pid + " is disallowed for pdf derivatives";
+        }
```
bbpennel: I don't think this necessarily makes sense now since the PID is a work object, and a work object can have many mimetypes (i'm not sure what one would be set in the message here, possibly none, or possibly the first one), some might be permitted and others might not. For example, a work containing a PDF generated by this processor would not be considered valid if you submitted it for processing again, since it would now contain a pdf. So I think the only thing that would make the pid valid is that it is a work and possibly that it contains at least one image (we probably should not create PDFs if there are no images, but I'm not sure the best place to handle that)

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/PdfDerivativeProcessor.java
```diff
@@ -0,0 +1,148 @@
...
+        Path derivativeFinalPath = setDerivativeFinalPath(request.getWorkPid());
+
+        aclService.assertHasAccess("User does not have permission to run enhancements",
+                pid, agent.getPrincipals(), Permission.runEnhancements);
+
+        assertValid(pid, mimetype);
+
+        try {
+            Path derivativeTmpPath = pdfDerivativeService.generatePdfDerivative(request);
+            moveFile(derivativeTmpPath, derivativeFinalPath);
```
bbpennel: We need to introduce the `AggregateFile` Resource type in the Cdr.java file, down with the other types like DescriptiveMetadata.

Then for this particular workflow we actually want to store the PDF as another file within the work rather than a derivative file. This is a bit complicated, but I think the workflow would be:
1. Mint a pid for a new file object, using PIDMinter
2. Get the PID for an originalFile using DatastreamPids
3. Get the storageLocation for writing the PDF to (see StorageLocationManager.getDefaultStorageLocation(workPid))
4. Get the storageUri by doing `storageLocation.getNewStorageUri(originalFilePid)`
5. Copy the generated PDF to this storageUri
6. Make a new empty jena Model, and add an RDF type relationship to it where the object of the relationship is the new `AggregateFile` type
7. Call `workObject.addDataFile(PID filePid, URI storageUri, String filename,            String mimetype, String sha1Checksum, String md5Checksum, Model model)`, passing in the storageUri as the first param, and the model as the model object created in step 6. This creates a new fileObject with a originalFile binary

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/PdfDerivativeProcessor.java
```diff
@@ -0,0 +1,148 @@
...
+        } catch (Exception e) {
+            log.error("Failed to generate pdf derivative for {}: {}", pid, "");
```
bbpennel: Rethrowing Exception isn't ideal, but I think you also mean to have `e` as the last param here (you don't need to put a `{}` into the message when displaying an exception)

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
krwong: pdf4u expects one value for all the files being added.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -53,21 +54,25 @@ public class PdfDerivativeService {
             SearchFieldKey.ID.name(), SearchFieldKey.ANCESTOR_PATH.name());
 
     private static final List<String> FILE_REQUEST_FIELDS = Arrays.asList(
-            SearchFieldKey.ID.name(), SearchFieldKey.FILE_FORMAT_TYPE.name(), SearchFieldKey.DATASTREAM.name(),
+            SearchFieldKey.ID.name(), SearchFieldKey.FILE_FORMAT_TYPE.name(),
             SearchFieldKey.ANCESTOR_PATH.name(), SearchFieldKey.TRANSCRIPT.name());
 
-    public PdfDerivativeService() {
+    public AggregatePdfService() {
         try {
```
bbpennel: the constructor probably shouldn't call this too, since it'll potentially mean folders get created in two different places, but only one used

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -96,49 +100,75 @@ public Path generatePdfDerivative(PdfRequest request) throws Exception {
...
+        var parentRec = getParentRecord(workPid, agent);
+        assertParentRecordValid(workPid, parentRec);
+
+        // retrieve transcript and write to temporary transcript file
+        List<ContentObjectRecord> children = getChildrenRecords(parentRec, agent);
+        for (var child : children) {
+            var transcriptFilePath = prepareTempPath(child.getId() + "_transcript", ".txt");
+            var transcriptValue = child.getTranscript();
+            try (BufferedWriter writer = new BufferedWriter(new FileWriter(transcriptFilePath.toFile()))) {
+                writer.write(transcriptValue);
```
bbpennel: not required, just wanted to point out that in current java versions you can write a string to file using, which also allows you to keep using the Path object
```
Files.write(transcriptFilePath, transcriptValue);
```

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
bbpennel: hmm, that might be an issue if a work contains multple images and some contain printed text while others contain cursive

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/AggregatePdfProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+    public void process(Exchange exchange) throws Exception {
+        var request = deserializeRequest(exchange);
+        var agent = request.getAgent();
+        var workPid = PIDs.get(request.getWorkPid());
+        var pdfPid = pidMinter.mintContentPid();
+        var originalFilePid = DatastreamPids.getOriginalFilePid(workPid);
+        var pdfStorageUri = Paths.get(locationManager.getDefaultStorageLocation(workPid)
+                .getNewStorageUri(originalFilePid));
+
+        aclService.assertHasAccess("User does not have permission to generate aggregate PDF",
```
bbpennel, line 55: do the access control check before any real work, so I'd suggest moving it to right after you set the `agent` variable

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/AggregatePdfProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    /**
+     *  Throws an IllegalArgumentException if the file is not eligible for having pdf derivatives generated from it
+     * @param pid work pid
+     */
+    private void assertValid(PID pid) {
+        try {
+            repositoryObjectLoader.getWorkObject(pid);
```
bbpennel: To save work, we should just load the workObject once. At the moment its loaded here and then loaded again in process. This method could be something like "loadWorkObject" which happens to also throw the IllegalArgumentException

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/PdfEnhancementsRouter.java
```diff
@@ -9,35 +9,35 @@
...
     @Override
     public void configure() throws Exception {
-        from(pdfDerivativesStreamCamel)
-                .routeId("PdfAccessCopy")
-                .log(LoggingLevel.DEBUG, log, "Received PDF with OCR generation request")
-                .bean(pdfDerivativeProcessor)
+        from(aggregatePdfStreamCamel)
+                .routeId("AggregatePdf")
+                .log(LoggingLevel.DEBUG, log, "Received aggregate PDF with OCR generation request")
+                .bean(aggregatePdfProcessor)
```
bbpennel, line 31: We'll need to test its behavior to see if we need an explicit indexing step, or if the fedora messages are sufficient to get the new PDF FileObject to show up.

services-camel-app/src/test/java/edu/unc/lib/boxc/services/camel/pdf/AggregatePdfProcessorTest.java
```diff
@@ -0,0 +1,305 @@
...
+        var exchange = createRequestExchange(workPid.getId(), "image/tiff");
+
+        when(aggregatePdfService.generateAggregatePdf(any()))
+                .thenThrow(new RuntimeException("conversion failed"));
+
+        var e = assertThrows(RuntimeException.class, () -> {
+            processor.process(exchange);
+        });
+
+        assertEquals(RuntimeException.class, e.getClass());
```
bbpennel: This is already verified by the assertThrows call

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
krwong: Potentially? kraken can process printed text so it might be okay

bbpennel: I guess we can see how it does, but i guess this effectively means we'd always use krakken. If we start basing this value off of what comes back from boxctron, we would have to include more logic to decide which type to use.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,238 @@
...
+        var transcriptListPath = prepareTempPath(workPid + "_transcriptlist", ".txt");
+        var transcriptList = new ArrayList<>();
+        var parentRec = getParentRecord(workPid, agent);
+        assertParentRecordValid(workPid, parentRec);
+
+        // retrieve transcript and write to temporary transcript file
+        List<ContentObjectRecord> children = getChildrenRecords(parentRec, agent);
+        for (var child : children) {
+            var transcriptFilePath = prepareTempPath(child.getId() + "_transcript", ".txt");
+            var transcriptValue = child.getTranscript();
```
bbpennel, line 144: We probably need a null check on this, since some children records won't have had transcripts generated

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,238 @@
...
+            CLIMain.runCommand(command);
+
+            return tempPath;
+        } catch (ObjectTypeMismatchException e) {
+            log.debug("Object {} is not a work object", request.getWorkPid(), e);
+            throw new IllegalArgumentException("Object " + workPid + " is not a work object");
+        } catch (Exception e) {
+            throw new ServiceException("Failed to generate aggregate PDF to " + tempPath + " for " + workPid, e);
+        } finally {
+            List<String> temporaryFiles = Arrays.asList(inputFiles, transcriptFiles);
```
bbpennel: we also need to delete all the transcript files listed in the transcript list file. we should be able to read the file into the temp files list here

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,238 @@
...
+     * Retrieve transcript value and write to temp file, then create .txt file with list of all transcript files
+     * @param request PdfRequest
+     * @return .txt path to transcript files
+     */
+    public Path createTranscriptListFile(PdfRequest request) {
+        var workPidString = request.getWorkPid();
+        var workPid = PIDs.get(workPidString);
+        var agent = request.getAgent();
+
+        var transcriptListPath = prepareTempPath(workPid + "_transcriptlist", ".txt");
```
bbpennel: use workPid.getId(), otherwise I think the filename is going to start with a url

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/AggregatePdfProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+
+        try {
+            Path pdfTmpPath = aggregatePdfService.generateAggregatePdf(request);
+            moveFile(pdfTmpPath, pdfStorageUri);
+
+            Model model = ModelFactory.createDefaultModel();
+            model.getResource("").addProperty(RDF.type, Cdr.AggregateFile);
+            workObject.addDataFile(pdfPid, pdfStorageUri.toUri(), pdfStorageUri.getFileName().toString(),
+                    "application/pdf", null, null, model);
+        } catch (IOException e) {
```
bbpennel: If any exceptions are thrown but pdfTmpPath isn't null, we should delete the pdfTmpPath. Since the likely exceptions would come from workObject.addDataFile we probably need to broaden the catch to any exception

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,238 @@
...
+                    + ", it either does not exist or is not accessible");
+        }
+    }
+
+    /**
+     * Create tmp pdf4u files directory for temporary files
+     * @return tmpImageFilesDirectoryPath
+     */
+    private void initializeTempImageFilesDir() throws IOException {
+        Path path = tmpFilesDir;
```
bbpennel: tmpFilesDir needs to be initialized in this method, otherwise it will be pointed at the java tmp dir even if we configure a different location via setters, since at the moment it gets initialized as soon as the class is created. There's also not really a reason to assign its value to the `path` variable here, since we don't do anything else with `path`.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,238 @@
...
+        var parentRec = getParentRecord(workPid, agent);
+        assertParentRecordValid(workPid, parentRec);
+
+        // retrieve transcript and write to temporary transcript file
+        List<ContentObjectRecord> children = getChildrenRecords(parentRec, agent);
+        for (var child : children) {
+            var transcriptFilePath = prepareTempPath(child.getId() + "_transcript", ".txt");
+            var transcriptValue = child.getTranscript();
+            try {
+                Files.write(transcriptFilePath, transcriptValue.getBytes());
```
bbpennel: Define the character set with a third param here as `StandardCharsets.UTF_8`

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/AggregatePdfProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+    public void process(Exchange exchange) throws Exception {
+        var request = deserializeRequest(exchange);
+        var agent = request.getAgent();
+        var workPid = PIDs.get(request.getWorkPid());
+
+        aclService.assertHasAccess("User does not have permission to generate aggregate PDF",
+                workPid, agent.getPrincipals(), Permission.runEnhancements);
+
+        var pdfPid = pidMinter.mintContentPid();
+        var originalFilePid = DatastreamPids.getOriginalFilePid(workPid);
```
bbpennel: this should use pdfPid rather than workPid

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -0,0 +1,238 @@
...
+    public Path generateAggregatePdf(PdfRequest request) throws IOException {
+        var workPid = request.getWorkPid();
+        String inputFiles = createInputListFile(request).toString();
+        String transcriptFiles = createTranscriptListFile(request).toString();
+        Path tempPath = prepareTempPath(workPid, ".pdf");
+        String textType = getTextTypes(request);
+
+        try {
+            // check that object is a work object
+            repositoryObjectLoader.getWorkObject(PIDs.get(workPid));
```
bbpennel: the processor is already explicitly verifying that the object is a work before this, we probably don't need to do it a second time here

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -205,12 +207,12 @@ private void assertParentRecordValid(PID pid, ContentObjectRecord parentRec) {
 
     /**
      * Create tmp pdf4u files directory for temporary files
-     * @return tmpImageFilesDirectoryPath
+     * @return tmpFilesDirectoryPath
```
bbpennel: This method doesn't have a return value, so this should be removed

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/pdf/AggregatePdfProcessor.java
bbpennel: This comment was referring to `var originalFilePid = DatastreamPids.getOriginalFilePid(workPid);`, since the original file datastream belongs to the pdf fileObject, not the work. Changing the pid in the getDefaultStorageLocation as you already did is fine too, but they will probably always resolve to the same storage location.

bbpennel (approved):

PR #2137, opened 2026-05-04
BXC-5664 full description controller and tests

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-5664


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/FullDescriptionController.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Controller
+public class FullDescriptionController {
```
bbpennel, line 22: javadoc

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/FullDescriptionController.java
```diff
@@ -0,0 +1,63 @@
...
+    @Autowired
+    FullDescriptionUpdateService fullDescriptionUpdateService;
+    @Autowired
+    BinaryTransferService transferService;
+    @Autowired
+    RepositoryObjectLoader repositoryObjectLoader;
```
bbpennel: instance variables should be private here

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/FullDescriptionController.java
```diff
@@ -0,0 +1,63 @@
...
+    @PostMapping(value = "/edit/fullDescription/{id}")
+    @ResponseBody
+    public ResponseEntity<Object> updateFullDescription(@PathVariable("id") String id, @RequestParam("fullDescription") String fullDescription) {
+        var agent = AgentPrincipalsImpl.createFromThread();
+        var repoObject = repositoryObjectLoader.getRepositoryObject(PIDs.get(id));
+        var request = new FullDescriptionUpdateRequest();
+        request.setPidString(id);
+        request.setFullDescriptionText(fullDescription);
+        request.setAgent(agent);
+        request.setTransferSession(transferService.getSession(repoObject));
```
bbpennel: Looking at the alt text controller, I don't think we need to set a transferSession here for a single object update, shared sessions are for minimizing overhead in bulk updates. The service will start one if needed, and its a good idea to minimize the amount of work we do before the access control check anyways. So we shouldn't need to load repoObject either since this was its only usage.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/utils/MachineUpdateServiceTestHelper.java
```diff
@@ -0,0 +1,18 @@
...
+
+import edu.unc.lib.boxc.model.api.ids.PID;
+import edu.unc.lib.boxc.web.services.rest.MvcTestHelpers;
+import org.springframework.test.web.servlet.MvcResult;
+
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MachineUpdateServiceTestHelper {
```
bbpennel: The method in here is a fairly generic check that a response contains an action and a pid which could apply to a number of controllers, so I don't think this really needs to be scoped for MachineUpdates (I suppose alt text and full description aren't really machine updates either since you could submit totally human generated values for either). So I think this class should either have a different name, or put the helper method in FullDescriptionControllerTest since its only saving 2 lines in AltTextControllerTest but adding a layer of indirection.

bbpennel (changes requested):

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/utils/MachineUpdateServiceTestHelper.java
sharonluong: Okay I'll just repeat the lines, I think that would be clearer. Also I can't think of what name it would be to encompass all the tests I'd use it in (I was planning on doing it in the transcript controller test too)

bbpennel (approved):

PR #2141, opened 2026-05-05
Turnstile updates

lfarrell:


web-access-app/src/main/java/edu/unc/lib/boxc/web/access/controllers/BotChallengeController.java
```diff
@@ -46,9 +44,10 @@ public class BotChallengeController {
     public @ResponseBody
     String  turnstileJson(@RequestBody CfTurnstileToken token, HttpServletRequest request, HttpServletResponse response) throws IOException {
         HttpSession session = request.getSession(true);
-        String ipAddress = request.getRemoteAddr();
-
-        if (hasUncAddress(ipAddress, session) || hasValidTurnstileToken(session)){
+        String ipAddress = getClientIpAddress( request);
```
bbpennel: This doesn't appear to be used since the ip matching was used to the other controller, so I think we can remove it and getClientIpAddress

web-access-app/src/test/java/edu/unc/lib/boxc/web/access/controllers/BotChallengeControllerTest.java
```diff
@@ -105,7 +108,7 @@ public void turnstileJsonVerificationSuccessSetsExpiryAndTokenFlag() throws Exce
         assertTrue(result.contains("\"success\":true"));
         assertEquals(Boolean.TRUE, Objects.requireNonNull(request.getSession()).getAttribute("validCfTurnstileToken"));
         assertNotNull(request.getSession().getAttribute("turnstileTokenExpiresIn"));
-        assertEquals(NON_UNC_IP_ADDRESS, request.getSession().getAttribute("userIPAddress"));
+        assertNull(request.getSession().getAttribute("userIPAddress"));
```
bbpennel: This controller doesn't set userIPAddress, so we probably don't need the two null assertions on it here

web-common/src/main/java/edu/unc/lib/boxc/web/common/auth/filters/StoreAccessLevelFilter.java
```diff
@@ -109,6 +97,20 @@ public void doFilterInternal(HttpServletRequest request, HttpServletResponse res
...
         chain.doFilter(request, response);
+
+        // Set headers after request processing so session-backed values are current.
+        var tokenExpiration = session.getAttribute("turnstileTokenExpiresIn");
+        response.setHeader("token-expiration", String.valueOf(tokenExpiration));
+
+        var uncIPAddress = session.getAttribute("uncIPAddress");
+        var hasUncIpAddress = (uncIPAddress != null) ? String.valueOf(uncIPAddress) : "false";
+        response.setHeader("unc-ip-address", hasUncIpAddress);
+        response.setHeader("ip-address", String.valueOf(session.getAttribute("userIPAddress")));
```
bbpennel: doesn't seem like this header is ever used, is it directly used by turnstile or can it be removed?

web-common/src/main/java/edu/unc/lib/boxc/web/common/auth/filters/StoreAccessLevelFilter.java
```diff
@@ -109,6 +97,20 @@ public void doFilterInternal(HttpServletRequest request, HttpServletResponse res
         }
 
         chain.doFilter(request, response);
+
+        // Set headers after request processing so session-backed values are current.
+        var tokenExpiration = session.getAttribute("turnstileTokenExpiresIn");
+        response.setHeader("token-expiration", String.valueOf(tokenExpiration));
```
bbpennel: this header doesn't seem to be referenced anywhere, is it somehow used by turnstile?

bbpennel (changes requested):

web-common/src/main/java/edu/unc/lib/boxc/web/common/auth/filters/StoreAccessLevelFilter.java
lfarrell: Ah yes, that was just there for testing.

bbpennel (approved):

PR #2145, opened 2026-05-07
Add head request to check if a manifest exists and the user has access to it

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-5656

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestController.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestController.java
@@ -13,6 +13,8 @@ import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.CrossOrigin;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.ResponseBody;
 
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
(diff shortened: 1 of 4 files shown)


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/IiifV3ManifestController.java
```diff
@@ -59,6 +61,22 @@ public ResponseEntity<Object> getCanvas(@PathVariable("id") String id) {
...
+     * @return Response containing custom header about the manifest's existence and whether the user can access it.
+     */
+    @CrossOrigin
+    @GetMapping(value = "/iiif/v3/{id}/access", produces = APPLICATION_JSON_VALUE)
+    @ResponseBody
+    public ResponseEntity<Object> hasManifestAccess(@PathVariable("id") String id) {
+        PID pid = PIDs.get(id);
+        var hasManifestAccess = manifestService.hasManifestAccess(pid, AgentPrincipalsImpl.createFromThread());
+        return ResponseEntity.ok()
+                .header("Can-Access-Manifest", String.valueOf(hasManifestAccess))
```
bbpennel, line 74: From an API perspective I find communicating this information via a header rather than in the body or just via the response code to be kind of weird. Is it most helpful as a header for some reason?

It also doesn't totally agree with some other parts of the method. Like there is not body being returned in the response, and it doesn't produce any json, so the `produces`, `@ResponseBody` and `<Object>` aren't really needed unless this method is going to return a json body.

So I think we should pick one of the following:
1. Stick with the header and remove all the body related stuff and have the response code by 204 since there is no content in the body, 
2. switch to having this method return a JSON response and communicate that way,
3. Have the method not set a header or body, and then use a 204 response to indicate that the user has access, and a 403 or 404 if the user can't,
4. Switch to being a HEAD method, then return 200 for success or 4xx otherwise (this might not really be accurate though since we wouldn't have the other info about the manifest, like size, dates, etc without doing more work)

bbpennel (changes requested):

bbpennel (approved):

PR #2155, opened 2026-05-27
BXC-5644 correct mimetypes that don't match actual file type

krwong:
[https://unclibrary.atlassian.net/browse/BXC-5644](https://unclibrary.atlassian.net/browse/BXC-5644)

- add `CorrectMimetypesService` and `CorrectMimetypesController` which receives a csv of file object ids and mimetypes to update in fedora
- add tests


operations/src/main/java/edu/unc/lib/boxc/operations/impl/mimetype/CorrectMimetypesService.java
```diff
@@ -0,0 +1,220 @@
...
+        // Update a copy of the model for this object
+        Model model = binaryObject.getModel(true);
+        Resource resource = model.getResource(binaryObject.getPid().getRepositoryPath());
+
+        // Clear out existing mimetypes and add new mimetypes
+        resource.removeAll(CdrDeposit.mimetype);
+        resource.addProperty(CdrDeposit.mimetype, mimetype);
+
+        // Push the updated object back to fedora
+        repositoryObjectFactory.createOrTransformObject(binaryObject.getMetadataUri(), model);
```
bbpennel: I'd recommend using repositoryObjectFactory.createExclusiveRelationship, it deletes all previous values of a property and inserts a new value, and updates the object in fedora.

operations/src/main/java/edu/unc/lib/boxc/operations/impl/mimetype/CorrectMimetypesService.java
```diff
@@ -0,0 +1,220 @@
...
+            throw new AccessRestrictionException("Permission denied for " + id);
+        } catch (RuntimeException e) {
+            tx.cancel(e);
+            throw e;
+        } finally {
+            tx.close();
+        }
+
+        // Send message that the action completed
+        operationsMessageSender.sendUpdateDescriptionOperation(agent.getUsername(),
```
bbpennel: I'd suggest using sendAddOperation, which is a full reindex of the object. sendUpdateDescriptionOperation is a smaller reindexing job that is focused on the descriptive metadata of the object, so it's possible it could miss updating some aspects related to the mimetype

operations/src/main/java/edu/unc/lib/boxc/operations/impl/mimetype/CorrectMimetypesService.java
```diff
@@ -0,0 +1,220 @@
...
+        aclService.assertHasAccess(
+                "User does not have permissions to edit mimetypes",
+                PIDs.get(id), agent.getPrincipals(), Permission.editDescription);
+
+        RepositoryObject obj = repositoryObjectLoader.getRepositoryObject(PIDs.get(id));
+
+        // Verify object is a file object
+        if (!(obj instanceof FileObject)) {
+            throw new IllegalArgumentException("Cannot update mimetype for non-file object " + obj.getPid());
+        }
```
bbpennel: we could use repositoryObjectLoader.getFileObject to directly get the object as a FileObject, it would throw an ObjectTypeMismatchException if its not a file object which maps to a 400 response in the services app.

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/mimetype/CorrectMimetypesService.java
```diff
@@ -0,0 +1,193 @@
...
+    private PID processRow(String id, String mimetype, AgentPrincipals agent) {
+        if (!isValidMimetype(mimetype)) {
+            throw new InvalidMimeTypeException(mimetype, "Invalid mimetype");
+        }
+
+        aclService.assertHasAccess(
+                "User does not have permissions to edit mimetypes",
+                PIDs.get(id), agent.getPrincipals(), Permission.editDescription);
+
+        FileObject obj = repositoryObjectLoader.getFileObject(PIDs.get(id));
```
bbpennel: Lets make the PID once and reuse it throughout the method rather than making multiple `PIDs.get(id)` calls

operations/src/main/java/edu/unc/lib/boxc/operations/impl/mimetype/CorrectMimetypesService.java
```diff
@@ -0,0 +1,193 @@
...
+
+        FedoraTransaction tx = txManager.startTransaction();
+
+        try (Timer.Context context = timer.time()) {
+            // Get original_file and its mimetype
+            BinaryObject binaryObject = obj.getOriginalFile();
+            String oldMimetype = binaryObject.getMimetype();
```
bbpennel, line 114: This mostly applies to resuming in case the job gets interrupted, but could you add a check to make sure the old and new mimetype are different, and if they are the same then do an early return so we aren't making unnecessary updates?

operations/src/main/java/edu/unc/lib/boxc/operations/impl/mimetype/CorrectMimetypesService.java
```diff
@@ -0,0 +1,193 @@
...
+        // Send message that the action completed
+        operationsMessageSender.sendAddOperation(agent.getUsername(), Collections.singletonList(obj.getParentPid()),
+                Collections.singletonList(PIDs.get(id)), null, null);
+        return PIDs.get(id);
+    }
+
+    private boolean isValidMimetype(String mimetype) {
+        try {
+            return mimetype.contains("/");
+        } catch (Exception e) {
```
bbpennel: i think the only exception that could be thrown here would be a null pointer exception, but that seems to already be handled by the blank check up in `correctMimetypes`

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/CorrectMimetypesTest.java
```diff
@@ -0,0 +1,196 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class CorrectMimetypesTest {
```
bbpennel:
```suggestion
public class CorrectMimetypesControllerTest {
```

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/CorrectMimetypesTest.java
```diff
@@ -0,0 +1,196 @@
...
+        assertFalse(body.containsKey("error"));
+    }
+
+    @Test
+    public void authorizationExceptionTest() throws Exception {
+        when(multipartFile.isEmpty()).thenReturn(false);
+        when(multipartFile.getInputStream())
+                .thenReturn(new ByteArrayInputStream("id,mimetype\n".getBytes(StandardCharsets.UTF_8)));
+
+        Exception serviceException = new RuntimeException(
```
bbpennel: I don't think it should be wrapped in a RuntimeException, it seems like the AuthorizationException is thrown directly

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/CorrectMimetypesController.java
```diff
@@ -0,0 +1,74 @@
...
+            result.put("error", "CSV file is required");
+            result.put("timestamp", System.currentTimeMillis());
+            return new ResponseEntity<>(result, HttpStatus.BAD_REQUEST);
+        }
+
+        List<PID> pids = new ArrayList<>();
+        try (InputStream correctMimetypeStream = file.getInputStream()) {
+            pids = correctMimetypesService.correctMimetypes(correctMimetypeStream, AgentPrincipalsImpl.createFromThread());
+            result.put("status", "Corrected mimetypes for " + pids);
+        } catch (StateUnmodifiedException e) {
```
bbpennel: After skimming through all the usages of StateUnmodifiedException, I don't think it can actually get thrown here. It appears to only be thrown when a datastream (like the MODS description) doesn't change. For a bulk operation, I think it's fine for no-op updates to just be ignored.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/CorrectMimetypesTest.java
```diff
@@ -0,0 +1,196 @@
...
+        assertFalse(body.containsKey("timestamp"));
+    }
+
+    @Test
+    public void accessRestrictionExceptionCauseReturnsForbiddenTest() throws Exception {
+        when(multipartFile.isEmpty()).thenReturn(false);
+        when(multipartFile.getInputStream())
+                .thenReturn(new ByteArrayInputStream("id,mimetype\n".getBytes(StandardCharsets.UTF_8)));
+
+        Exception serviceException = new RuntimeException(
```
bbpennel: Same here, I don't think the exception needs to be wrapped

bbpennel (changes requested):

static/templates/admin/CorrectMimetypesForm.html
```diff
@@ -0,0 +1,18 @@
+<iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
+<form id="correct_mimetypes_form" method="post" action="/services/api/edit/correctMimetypes" enctype="multipart/form-data" target="upload_file_container" class="edit_form">
+    <p>Select the correct mimetypes CSV document you wish to import.</p>
+    <p>You will be sent an email notification when the import completes.</p>
```
bbpennel: I don't think this is true?

static/js/admin/src/CorrectMimetypesForm.js
```diff
@@ -0,0 +1,47 @@
...
+        CorrectMimetypesForm.prototype.validationErrors = function() {
+            var errors = [];
+            var dataFile = $("input[type='file']", this.$form).val();
+            if (!dataFile)
+                errors.push("You must select a file to import");
+            return errors;
+        };
+
+        CorrectMimetypesForm.prototype.getSuccessMessage = function(data) {
+            return this.ingestFile.name + " has been successfully uploaded for updating.  You will receive an email when it completes.";
```
bbpennel: I don't think it sends an email does it?

static/templates/admin/CorrectMimetypesForm.html
```diff
@@ -0,0 +1,18 @@
+<iframe id="upload_file_frame" name="upload_file_frame" height="0" width="0" style="display: none;"></iframe>
+<form id="correct_mimetypes_form" method="post" action="/services/api/edit/correctMimetypes" enctype="multipart/form-data" target="upload_file_container" class="edit_form">
+    <p>Select the correct mimetypes CSV document you wish to import.</p>
```
bbpennel: I think it should be "corrected"

static/js/admin/src/ImportMenu.js
```diff
@@ -21,6 +21,11 @@ define('ImportMenu', [ 'jquery', 'jquery-ui', 'underscore', 'ImportMemberOrderFo
                 items["importRefIds"] = {name : "Bulk Ref Ids"};
             }
 
+            if (this.container !== undefined && this.container.type === 'File'
```
bbpennel: the container type will almost never be File, I think it's fine to remove the `this.container.type === 'File'` condition since its fine to run the update from anywhere, the items being updated aren't necessarily the item being viewed since this is a bulk import, and the server should already be validating that the user is only submitting valid objects.

bbpennel (approved):

PR #2159, opened 2026-06-09
BXC-5550 on-demand PDF generation service APIs

krwong:
[https://unclibrary.atlassian.net/browse/BXC-5550](https://unclibrary.atlassian.net/browse/BXC-5550)

- update javadoc and names to "aggregate pdf"
- add `extractTextType` method to `MachineGeneratedContentService` to retrieve the `textType`
- update `getTextTypes` method to use `extractTextType`
- add `AggregatePdfController` 
- add `Aggregate PDF` action to `ResultObjectActionMenu`
- add/update tests and test resources


operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -171,9 +175,34 @@ public Path createTranscriptListFile(PdfRequest request) {
...
      */
     public String getTextTypes(PdfRequest request) {
-        //TODO: get text type from the alt text review
+        var textType = RESULT_HANDWRITTEN_CURSIVE;
 
-        return "HANDWRITTEN-PRINT";
+        var workPid = PIDs.get(request.getWorkPid());
+        var agent = request.getAgent();
+        var parentRec = getParentRecord(workPid, agent);
+        List<ContentObjectRecord> children = getChildrenRecords(parentRec, agent);
```
bbpennel, line 185: Weren't we planning to pass in a list of types, one per image?

search-solr/src/main/java/edu/unc/lib/boxc/search/solr/services/MachineGeneratedContentService.java
```diff
@@ -168,6 +168,31 @@ public JsonNode extractSafetyAssessment(JsonNode mgdNode) {
...
+        if (mgdNode == null) {
+            return null;
+        }
+
+        JsonNode textTypeNode = extractSafetyAssessment(mgdNode).path("text_characteristics")
+                .path("text_type");
+
+        if (!textTypeNode.isMissingNode()) {
+            String textType = textTypeNode.asText();
+            if (RESULT_HANDWRITTEN_PRINT.equalsIgnoreCase(textType)
```
bbpennel: There are 3 other possible values: N/A, PRINTED, and TYPED. N/A should probably not return a type, but all the others should make it through.

static/js/admin/src/action/AggregatePdfAction.js
```diff
@@ -0,0 +1,72 @@
...
+    function AggregatePdfAction(context) {
+        this._create(context);
+    };
+
+    AggregatePdfAction.prototype.constructor = AggregatePdfAction;
+    AggregatePdfAction.prototype = Object.create( AbstractBatchAction.prototype );
+
+    AggregatePdfAction.prototype.isValidTarget = function(target) {
+        return target.isSelected() && target.isEnabled() && $.inArray("editResourceType", target.metadata.permissions) != -1
+            && "Folder" == target.getMetadata().type;
```
bbpennel: We should also be able to run this one Work objects

static/js/admin/src/ResultObjectActionMenu.js
```diff
@@ -257,6 +253,9 @@ define('ResultObjectActionMenu', [ 'jquery', 'jquery-ui', 'StringUtilities',  'A
 			if (!isContentRoot && $.inArray('destroy', metadata.permissions) != -1) {
 				items["destroy"] = {name : 'Destroy', disabled :  !resultObject.isDeleted};
 			}
+			if (metadata.type === 'Folder' && $.inArray('reindex', metadata.permissions) != -1) {
```
bbpennel: This should be available to works as well

static/js/admin/src/action/AggregatePdfAction.js
```diff
@@ -0,0 +1,72 @@
...
+
+        this.targets = this.getTargets();
+        if (this.targets.length == 1) {
+            title = "Create aggregrate PDFs for " + this.targets[0].metadata.title.substring(0, 30);
+        } else {
+            title = "Create aggregate PDFs for " + this.targets.length + " objects";
+        }
+
+        var aggregatePdfForm = aggregatePdfTemplate();
+        this.dialog = $("<div class='containingDialog'>" + aggregatePdfForm + "</div>");
```
bbpennel: I believe there is a confirmation dialogue option that we could use instead of creating a new form, you could check the "Reindex" or "Destroy" actions to see how those work

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/AggregatePdfController.java
```diff
@@ -0,0 +1,59 @@
...
+    @RequestMapping(value = "edit/aggregatePdf")
+    @ResponseBody
+    public ResponseEntity<Object> aggregatePdf(@RequestParam("ids") String ids) {
+        if (isEmpty(ids)) {
+            throw new IllegalArgumentException("Must provide one or more ids");
+        }
+
+        Map<String, Object> result = new HashMap<>();
+        var agent = AgentPrincipalsImpl.createFromThread();
+        result.put("action", "generate aggregate PDF");
```
bbpennel:
```suggestion
        result.put("action", "generate aggregate PDFs");
```

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/AggregatePdfController.java
```diff
@@ -0,0 +1,59 @@
...
+
+        for (String id : ids.split("\n")) {
+            PdfRequest pdfRequest = new PdfRequest();
+            pdfRequest.setAgent(agent);
+            pdfRequest.setWorkPid(id);
```
bbpennel, line 55: this request object seems to assume that it is getting work ids, but the UI components seem to assume folder objects. Works are were the actual aggregates will go, so it makes sense for the requests to be focused around those. But we need to either make the UI stuff only apply to Works, or the controller needs to be able to receive a folder id and expand it into work IDs. We could check with Anna to see if she thinks she will want to be able to submit entire folders.

bbpennel (changes requested):

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/AggregatePdfController.java
krwong: Yeah, this sentence in the ticket description really threw me `This would need to be able to be run on a folder to process all the works in it.` I checked with Anna and she said just works.

operations/src/test/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfServiceTest.java
```diff
@@ -125,7 +125,7 @@ public void generateAggregatePdfTest() throws Exception {
             assertNotNull(cmd);
             assertTrue(cmd.contains("pdf4u"));
             assertTrue(FilenameUtils.getBaseName(cmd.get(3)).startsWith(PARENT_UUID));
-            assertEquals(RESULT_HANDWRITTEN_CURSIVE, cmd.get(9));
+            assertEquals(List.of(RESULT_HANDWRITTEN_CURSIVE).toString(), cmd.get(9));
```
bbpennel: this is expecting the parameter to have value `[HANDWRITTEN_CURSIVE]`, but that seems pretty unusual for passing a list into a CLI, I would expect it to be a comma delimited list like `HANDWRITTEN_CURSIVE,PRINTED`

static/js/admin/src/action/AggregatePdfAction.js
```diff
@@ -9,8 +8,9 @@ define('AggregatePdfAction', [ 'jquery', 'AbstractBatchAction', "tpl!templates/a
     AggregatePdfAction.prototype = Object.create( AbstractBatchAction.prototype );
 
     AggregatePdfAction.prototype.isValidTarget = function(target) {
-        return target.isSelected() && target.isEnabled() && $.inArray("editResourceType", target.metadata.permissions) != -1
-            && "Folder" == target.getMetadata().type;
+        return target.isSelected() && target.isEnabled()
+            && $.inArray("reindex", target.metadata.permissions) !== -1
```
bbpennel: I checked in with Anna about who should have permission to run this workflow. I think we should switch back to `editResourceType` (here and in AggregatePdfProcessor). Reindex and Run Enhancements are only available to admins, but she suggested that canManager role be able to trigger this workflow.

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -201,7 +202,7 @@ public List<String> createTextTypeList(PdfRequest request) {
             }
         }
 
-        return textTypeList;
+        return textTypeList.stream().map(Object::toString).collect(Collectors.joining(","));
```
bbpennel: I would suggest having the service return a List of paths still, and serializing it into a string where it needs to be a string for presenting it to the user. Returning a list of Paths is cleaner and more usable programmatically.

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
```diff
@@ -166,14 +195,95 @@ public Path createTranscriptListFile(PdfRequest request) {
...
+
+        // get collectionId from parent collection and get hookId from work
+        var workFields = getFilenameRecord(workPid, agent);
+        var parentFields = getFilenameRecord(PIDs.get(workFields.getParentCollectionId()), agent);
+
+        String collectionId = parentFields.getCollectionId();
+
+        // if hookId field is empty, get hookId from the identifier field
+        String hookId = workFields.getHookId();
+        if (hookId == null || hookId.equals("null")) {
```
bbpennel: Do we really have hookids that are set to the string "null"? I wonder where those are coming from

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/AggregatePdfController.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    private ResponseEntity<Object> aggregatePdfs(String... ids) {
+        Map<String, Object> result = new HashMap<>();
+        var agent = AgentPrincipalsImpl.createFromThread();
+
+        result.put("action", "generate aggregate PDFs");
+
+        for (String id : ids) {
+            PdfRequest pdfRequest = new PdfRequest();
```
bbpennel, line 61: I would suggest adding an access control check into this method before we send the requests so it can fail fast and protect the AMQ queues. At the moment, there is no immediate indication to users that they don't have permission to perform this action, it only fails after its already been pushed to the queue and picked up for processing.

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/modify/AggregatePdfControllerTest.java
```diff
@@ -0,0 +1,112 @@
...
+import static edu.unc.lib.boxc.web.services.rest.MvcTestHelpers.getMapFromResponse;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.MockitoAnnotations.openMocks;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+public class AggregatePdfControllerTest {
```
bbpennel, line 37: We will want to add a permission denied test

bbpennel (changes requested):

operations/src/main/java/edu/unc/lib/boxc/operations/impl/pdf/AggregatePdfService.java
krwong: Oops I don't think we do, this was from me typing "null" instead of null in the service tests. Fixed it!

bbpennel (approved):

PR #2161, opened 2026-06-12
BXC-5684 add new param for run enhancement requests

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-5684

### DIFF
--- a/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
+++ b/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
@@ -35,40 +35,6 @@ public class RunEnhancementsMessageHelpers {
      * @param userid
      * @param pid
      * @param force
-     * @return
-     */
-    public static Document makeEnhancementOperationBody(String userid, PID pid, Boolean force) {
-        Document msg = new Document();
-        Element entry = new Element("entry", ATOM_NS);
-        entry.addContent(new Element("author", ATOM_NS)
-                .addContent(new Element("name", ATOM_NS).setText(userid)));
-
-        Element paramForce = new Element("force", CDR_MESSAGE_NS);
-        if (force) {
-            paramForce.setText("true");
-        } else {
-            paramForce.setText("false");
-        }
-
-        Element paramEnhancementList = new Element("enhancementList", CDR_MESSAGE_NS);
-        paramEnhancementList.setText(DEFAULT_ENHANCEMENTS_STRING);
-
-        Element enhancements = new Element(CDRActions.RUN_ENHANCEMENTS.getName(), CDR_MESSAGE_NS);
-        enhancements.addContent(new Element("pid", CDR_MESSAGE_NS).setText(pid.getRepositoryPath()));
-        enhancements.addContent(paramForce);
-        entry.addContent(enhancements);
-
-        msg.addContent(entry);
-
-        return msg;
-    }
-
-    /**
-     * Generate the body for a run enhancement request message
-     *
-     * @param userid
-     * @param pid
-     * @param force
      * @param enhancementList list of enhancements to run
      * @return
      */
@@ -85,15 +51,17 @@ public class RunEnhancementsMessageHelpers {
             paramForce.setText("false");
         }
 
-        Element paramEnhancementList = new Element("enhancementList", CDR_MESSAGE_NS);
-        paramEnhancementList.setText(String.join(",",enhancementList));
-
         Element enhancements = new Element(CDRActions.RUN_ENHANCEMENTS.getName(), CDR_MESSAGE_NS);
         enhancements.addContent(new Element("pid", CDR_MESSAGE_NS).setText(pid.getRepositoryPath()));
         enhancements.addContent(paramForce);
-        enhancements.addContent(paramEnhancementList);
-        entry.addContent(enhancements);
 
+        if (enhancementList != null) {
+            Element paramEnhancementList = new Element("enhancementList", CDR_MESSAGE_NS);
+            paramEnhancementList.setText(String.join(",",enhancementList));
+            enhancements.addContent(paramEnhancementList);
+        }
+
+        entry.addContent(enhancements);
         msg.addContent(entry);
 
         return msg;
--- a/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
+++ b/operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
@@ -9,12 +9,22 @@ import org.jdom2.Element;
 import edu.unc.lib.boxc.model.api.ids.PID;
 import edu.unc.lib.boxc.operations.jms.JMSMessageUtil.CDRActions;
 
+import java.util.List;
+
 /**
  * Helper methods for run enhancement messages
  *
  * @author bbpennel
  */
 public class RunEnhancementsMessageHelpers {
+    public static final String ENHANCEMENT_LIST = "enhancementList";
+    public static final String IMAGE_ACCESS_COPY = "imageAccessCopy";
+    public static final String EXTRACT_FULLTEXT = "extractFulltext";
+    public static final String AUDIO_ACCESS_COPY = "audioAccessCopy";
+    public static final String VIDEO_ACCESS_COPY = "videoAccessCopy";
+    public static final String MACHINE_GEN_DESCRIPTION = "machineGenDescription";
+    public static final List<String> DEFAULT_ENHANCEMENTS = List.of(IMAGE_ACCESS_COPY,EXTRACT_FULLTEXT,
+            AUDIO_ACCESS_COPY, VIDEO_ACCESS_COPY, MACHINE_GEN_DESCRIPTION);
 
     private RunEnhancementsMessageHelpers() {
     }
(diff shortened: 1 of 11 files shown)


operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
```diff
@@ -27,23 +27,30 @@ private RunEnhancementsMessageHelpers() {
      * @param force
      * @return
      */
-    public static Document makeEnhancementOperationBody(String userid, PID pid, Boolean force) {
+    public static Document makeEnhancementOperationBody(String userid, PID pid, Boolean force, Boolean regenerateDescription) {
```
bbpennel: this works, but when we met I feel like we talked about having a "enhancementsSet" type field instead of a boolean. It would usually be null/empty (to run everything), and if populated it would override the set of enhancements that get run. The nice part about the enhancementsList approach is that we could use it for targeting any of the enhancements in the future without needing to add flags for each enhancement, like if we needed to only regenerate all the jp2s.

Also, we could keep the previous `public static Document makeEnhancementOperationBody(String userid, PID pid, Boolean force)` signature which calls an override of the method with the extra field. That way all the callers that don't specify an enhancementSet (or regenerateDescription) wouldn't need to be changed.

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/BinaryEnhancementProcessor.java
```diff
@@ -20,14 +21,17 @@
...
+
 /**
  * Sets headers related to identifying binary objects to run enhancement operations on
  *
  * @author lfarrell
  */
 public class BinaryEnhancementProcessor implements Processor {
     private static final Logger log = LoggerFactory.getLogger(BinaryEnhancementProcessor.class);
-
+    public static final String DEFAULT_ENHANCEMENTS = "imageAccessCopy,extractFulltext,audioAccessCopy," +
```
bbpennel: We should consider breaking the names of all the enhancements out into their own enum in operations-jms, so that other modules can use the names (relating to the enhancementSet message field).

operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
bbpennel: the implementation for this method should just be:
```
return makeEnhancementOperationBody(userid, pid, force, null);
```
that should DRY things up a good bit since they should construct the same message, with on optional field

operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
```diff
@@ -34,16 +44,54 @@ public static Document makeEnhancementOperationBody(String userid, PID pid, Bool
...
+                .addContent(new Element("name", ATOM_NS).setText(userid)));
+
+        Element paramForce = new Element("force", CDR_MESSAGE_NS);
         if (force) {
             paramForce.setText("true");
         } else {
             paramForce.setText("false");
         }
 
+        Element paramEnhancementList = new Element("enhancementList", CDR_MESSAGE_NS);
```
bbpennel: we should only add the enhancementList field if `enhancementList` is not null or empty

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/BinaryEnhancementProcessor.java
```diff
@@ -56,11 +60,19 @@ public void process(final Exchange exchange) throws Exception {
                     in.setHeader(FcrepoJmsConstants.RESOURCE_TYPE, String.join(",", repoObj.getTypes()));
 
                     Element forceText = enhancementsEl.getChild("force", CDR_MESSAGE_NS);
-                    if (forceText != null) {
+                    Element enhancementsList = enhancementsEl.getChild(ENHANCEMENT_LIST, CDR_MESSAGE_NS);
+                    var enhancements = enhancementsList.getTextTrim();
+                    // if it's a regen request set force to true
+                    if (Objects.equals(enhancements, MACHINE_GEN_DESCRIPTION)) {
```
bbpennel: I'm not sure we need to always set the force flag to true for machine generation requests, it seems helpful to allow for skipping over items that already have the machine gen data present when doing a bulk run enhancements action on a folder or something, in case we are trying to get it to fill missed item.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/RunEnhancementsService.java
```diff
@@ -105,33 +108,34 @@ private void recursiveEnhancements(PID pid, AgentPrincipals agent, Boolean force
...
 
-    private void createMessage(ContentObjectRecord metadata, String username, Boolean force) {
+    private void createMessage(ContentObjectRecord metadata, String username, Boolean force, Boolean regenerateDescription) {
         PID pid = metadata.getPid();
         Datastream originalDs = metadata.getDatastreamObject(ORIGINAL_FILE.getId());
         String resourceType = metadata.getResourceType();
         PID originalPid = (ResourceType.File.equals(resourceType) && originalDs != null) ?
                 DatastreamPids.getOriginalFilePid(pid) : pid;
-        Document msg = makeEnhancementOperationBody(username, originalPid, force);
+        var enhancements = regenerateDescription ? List.of(MACHINE_GEN_DESCRIPTION) : DEFAULT_ENHANCEMENTS;
```
bbpennel: If the user hasn't specified what enhancements to run, it should be fine for us to pass in no enhancement list since we set the list of default enhancements later on in the workflow

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/RunEnhancementsRequest.java
```diff
@@ -12,6 +12,7 @@ public class RunEnhancementsRequest {
     private boolean force;
     private AgentPrincipals agent;
     private boolean recursive;
+    private boolean regenerateDescription;
```
bbpennel: Since we have updated the other parts of the pipeline to allow for user defined enhancement lists, I'd suggest doing that here too rather than tightly coupling the request object to the description generation enhancement specifically.

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/processing/RunEnhancementsService.java
```diff
@@ -105,33 +108,34 @@ private void recursiveEnhancements(PID pid, AgentPrincipals agent, Boolean force
...
-        createMessage(metadata, agent.getUsername(), force);
+        createMessage(metadata, agent.getUsername(), force, regenerateDescription);
     }
 
     public void setAclService(AccessControlService aclService) {
         this.aclService = aclService;
     }
 
-    private void createMessage(ContentObjectRecord metadata, String username, Boolean force) {
+    private void createMessage(ContentObjectRecord metadata, String username, Boolean force, Boolean regenerateDescription) {
```
bbpennel: This method would reflect the enhancementList suggestion on the request object

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/BinaryEnhancementProcessor.java
sharonluong: But I thought we wanted to force a regeneration? That isn't the use case?

operations-jms/src/main/java/edu/unc/lib/boxc/operations/jms/RunEnhancementsMessageHelpers.java
```diff
@@ -25,16 +35,16 @@ private RunEnhancementsMessageHelpers() {
      * @param userid
      * @param pid
      * @param force
+     * @param enhancementList list of enhancements to run
      * @return
      */
-    public static Document makeEnhancementOperationBody(String userid, PID pid, Boolean force) {
+    public static Document makeEnhancementOperationBody(String userid, PID pid, Boolean force, List<String> enhancementList) {
```
bbpennel, line 41: looks like quilty is mad because there is no coverage of this method. I know there's no existing test, but since this is one of the central changes in this PR I'd suggest adding one, even if it just targets your modifications. It looks like it should be very testable

services-camel-app/src/main/java/edu/unc/lib/boxc/services/camel/BinaryEnhancementProcessor.java
```diff
@@ -61,6 +66,15 @@ public void process(final Exchange exchange) throws Exception {
                     } else {
                         in.setHeader("force", "false");
                     }
+
+                    Element enhancementsList = enhancementsEl.getChild(ENHANCEMENT_LIST, CDR_MESSAGE_NS);
```
bbpennel: In order to allow us to catch empty enhancementLists, not just the absence of the XML element, we could directly use `getChildTextTrim` here and then do a blank/null check in the conditional

web-services-app/src/test/java/edu/unc/lib/boxc/web/services/rest/RunEnhancementsIT.java
```diff
@@ -220,6 +220,28 @@ public void runEnhancementsNoAccess() throws Exception {
...
+    @Test
+    public void setRunEnhancementsRegenerateDescription() throws Exception {
+        FileObject fileObj = repositoryObjectFactory.createFileObject(null);
+        PID filePid = fileObj.getPid();
+        fileObj.addOriginalFile(makeContentUri(filePid, BINARY_CONTENT), "file.png", "image/png", null, null);
+        setResultMetadataObject(filePid, ResourceType.File.name());
+
+        MvcResult result = mvc.perform(post("/runEnhancements")
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .content("{\"force\":true,\"pids\":[\"" + filePid.getId() + "\"],\"regenerateDescription\":true}")
```
bbpennel: I believe the parameters need to be updated here since its still using regenerateDescription. That the test still passes probably demonstrates that we need to inspect the details of the message a bit more to verify that it passes along the enhancements we tell it to in the message

bbpennel (changes requested):

bbpennel (approved):

PR #2174, opened 2026-06-26
Bxc 5691 wcag compliance fedora property

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-5691

### DIFF
--- a/operations/src/main/java/edu/unc/lib/boxc/operations/impl/utils/PropertiesUtil.java
+++ b/operations/src/main/java/edu/unc/lib/boxc/operations/impl/utils/PropertiesUtil.java
@@ -1,0 +1,14 @@
+package edu.unc.lib.boxc.operations.impl.utils;
+
+import edu.unc.lib.boxc.model.api.objects.RepositoryObject;
+import org.apache.jena.rdf.model.Property;
+
+public class PropertiesUtil {
+    public static String getValue(RepositoryObject repositoryObject, Property property) {
+        var propertyValue = repositoryObject.getResource().getProperty(property);
+        if (propertyValue == null) {
+            return null;
+        }
+        return propertyValue.getString();
+    }
+}
--- a/operations/src/main/java/edu/unc/lib/boxc/operations/impl/wcagCompliance/WcagComplianceService.java
+++ b/operations/src/main/java/edu/unc/lib/boxc/operations/impl/wcagCompliance/WcagComplianceService.java
@@ -1,7 +1,97 @@
 package edu.unc.lib.boxc.operations.impl.wcagCompliance;
 
-public class WcagComplianceService {
-    public void updateWcagCompliance() {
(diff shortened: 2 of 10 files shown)


operations/src/main/java/edu/unc/lib/boxc/operations/impl/utils/PropertiesUtil.java
```diff
@@ -0,0 +1,14 @@
+package edu.unc.lib.boxc.operations.impl.utils;
+
+import edu.unc.lib.boxc.model.api.objects.RepositoryObject;
+import org.apache.jena.rdf.model.Property;
+
+public class PropertiesUtil {
```
bbpennel: Please add javadoc for the class and for getValue. Also, PropertiesUtil is a very broad name, I'd suggest narrowing it to something related to fedora object properties. There should also be a unit test. There should probably also be a private constructor since this is a utility method class

operations/src/main/java/edu/unc/lib/boxc/operations/impl/wcagCompliance/WcagComplianceService.java
```diff
@@ -0,0 +1,109 @@
...
+
+        var currentLevel = PropertiesUtil.getValue(repoObj, Cdr.wcagCompliance);
+        // if current level is null (does not exist) and the requested level is blank, do nothing
+        if (currentLevel == null && StringUtils.isBlank(requestLevel)) {
+            log.debug("The current WCAG compliance level is null and the requested level is blank");
+            return;
+        }
+        // if we're just updating to the same existing compliance level, do nothing
+        if (!StringUtils.isBlank(currentLevel) && Objects.equals(currentLevel, requestLevel)) {
+            log.debug("The current WCAG compliance level is null and the requested level is blank");
```
bbpennel: Need to update this message, looks like it is the same as the previous block

operations/src/test/java/edu/unc/lib/boxc/operations/impl/wcagCompliance/WcagComplianceServiceTest.java
```diff
@@ -0,0 +1,196 @@
...
+        var request = buildRequest(LEVEL_A_10);
+        doThrow(new AccessRestrictionException("Access Denied")).when(aclService)
+                .assertHasAccess(any(), eq(pid), any(), eq(Permission.editResourceType));
+        Assertions.assertThrows(AccessRestrictionException.class, () -> {
+            service.updateWcagCompliance(request);
+        });
+    }
+
+    @Test
+    public void testNotAWork() {
```
bbpennel:
```suggestion
    public void testNotAFile() {
```
it is a work!

bbpennel (changes requested):

bbpennel (approved):

PR #2179, opened 2026-06-30
BXC-5699 add wcag compliance controller and DRY up code

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-5699

### DIFF
--- a/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/WcagComplianceController.java
+++ b/web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/WcagComplianceController.java
@@ -1,0 +1,47 @@
+package edu.unc.lib.boxc.web.services.rest.modify;
+
+import edu.unc.lib.boxc.auth.fcrepo.models.AgentPrincipalsImpl;
+import edu.unc.lib.boxc.operations.impl.wcagCompliance.WcagComplianceService;
+import edu.unc.lib.boxc.operations.jms.wcagCompliance.WcagComplianceRequest;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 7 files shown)


web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/WcagComplianceController.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@Controller
+public class WcagComplianceController {
```
bbpennel, line 24: javadoc

web-services-app/src/main/java/edu/unc/lib/boxc/web/services/rest/modify/WcagComplianceController.java
```diff
@@ -0,0 +1,47 @@
...
+    @Autowired
+    private WcagComplianceService service;
+
+    @PostMapping(value = "/edit/wcagCompliance/{id}", produces = APPLICATION_JSON_VALUE)
+    @ResponseBody
+    public ResponseEntity<Object> updateWcagCompliance(@PathVariable("id") String id, @RequestParam("level") String level) {
```
bbpennel: I haven't tried it myself, but it sounds like we might be able to do `updateWcagCompliance(@ModelAttribute WcagComplianceRequest wcagRequest)` it would map in both of the params (although I suspect the url pattern would need to match the names in the request object, so `"/edit/wcagCompliance/{pidString}"`). If that's an option it'd save a little bit of setup but also might be better from the perspective of keeping the request params consistent throughout the workflow

jp24u
bbpennel, line 1: Not sure what these changes to jp24u are, could you check to make sure its not reverting something that kara committed here? https://github.com/UNC-Libraries/box-c/pull/2172

bbpennel (changes requested):

bbpennel (approved):

PR #2182, opened 2026-07-06
Add option to admin UI to add/update WCAG compliance

lfarrell:
https://unclibrary.atlassian.net/browse/BXC-5692

### DIFF
--- a/static/js/admin/src/EditWcagComplianceForm.js
+++ b/static/js/admin/src/EditWcagComplianceForm.js
@@ -5,7 +5,11 @@ define('EditWcagComplianceForm', [ 'jquery', 'jquery-ui', 'underscore', 'RemoteS
         var defaultOptions = {
             title : 'Edit WCAG Compliance Level',
             createFormTemplate : editWcagComplianceForm,
-            submitMethod: 'POST'
+            submitMethod: 'POST',
+            complianceLevel: function(current_value, form_value) {
+                return current_value !== undefined && Array.isArray(current_value) &&
+                    current_value[0] === form_value;
+            }
         };
 
         function EditWcagComplianceForm(options) {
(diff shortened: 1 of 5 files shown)


static/js/admin/src/EditWcagComplianceForm.js
```diff
@@ -0,0 +1,56 @@
...
+        EditWcagComplianceForm.prototype.preprocessForm = function(resultObject) {
+            var pid = resultObject.metadata.id;
+            this.action_url = "/services/api/edit/wcagCompliance/" + pid;
+        };
+
+        EditWcagComplianceForm.prototype.validationErrors = function() {
+            var errors = [];
+            var wcagComplianceLevel = $("#edit_wcag_compliance", this.$form).val();
+            // Validate input
+            if (wcagComplianceLevel === "")
```
bbpennel: This check means that there's no way to clear the wcag compliance level once it is set. The server side services are setup to clear the wcag field is a blank value is provided, so we might want to permit this.

web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/SerializationUtil.java
```diff
@@ -238,6 +238,10 @@ public static Map<String, Object> metadataToMap(ContentObjectRecord metadata, Ac
             result.put(SearchFieldKey.TRANSCRIPT.getUrlParam(), metadata.getTranscript());
         }
 
+        if (metadata.getWcagCompliance() != null) {
```
bbpennel: This is actually redundant, it's already present https://github.com/UNC-Libraries/box-c/blob/ebfae48a8f0730a997af3e820ab9fcccdbc934d1/web-common/src/main/java/edu/unc/lib/boxc/web/common/utils/SerializationUtil.java#L266
however, your version of it is using `SearchFieldKey.WCAG_COMPLIANCE.getUrlParam` which is more correct than the existing implementation. Could you remove your change but update the existing conditional to use SearchFieldKey?

bbpennel (changes requested):

bbpennel (approved):

PR #2203, opened 2026-07-23
BXC-5725 Modify BinaryTransferClient to allow keeping old versions

krwong:
[https://unclibrary.atlassian.net/browse/BXC-5725](https://unclibrary.atlassian.net/browse/BXC-5725)

- add `transferAndKeep` method to keep old versions and use in `transferVersion`
- add test

### DIFF
--- a/persistence/src/main/java/edu/unc/lib/boxc/persist/impl/transfer/FSToFSTransferClient.java
+++ b/persistence/src/main/java/edu/unc/lib/boxc/persist/impl/transfer/FSToFSTransferClient.java
@@ -45,6 +45,8 @@ public class FSToFSTransferClient implements BinaryTransferClient {
 
     private static final Logger log = LoggerFactory.getLogger(FSToFSTransferClient.class);
 
+    public enum TransferMode { REJECT_EXISTING, REPLACE_EXISTING, KEEP_EXISTING_VERSIONS }
+
     public FSToFSTransferClient(IngestSource source, StorageLocation destination) {
         this.source = source;
         this.destination = destination;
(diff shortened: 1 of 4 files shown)


persistence/src/main/java/edu/unc/lib/boxc/persist/impl/transfer/FSToFSTransferClient.java
```diff
@@ -148,9 +148,39 @@ private static String copyFile(Path srcPath, Path destPath) throws IOException {
         return digest;
     }
 
+    /**
+     * Similar to transfer, but keeps old versions of original_files (no cleanup)
+     * @param binPid
+     * @param sourceFileUri
+     */
+    public BinaryTransferOutcome transferAndKeep(PID binPid, URI sourceFileUri) {
```
bbpennel: this method is the same as `transfer(PID binPid, URI sourceFileUri, boolean allowOverwrite)` but with the lines related to cleanup removed. For maintenance/DRY purposes, I'd suggest a couple changes:
1. We could introduce a new enum like "TransferMode" that had values like REJECT_EXISTING (same as what `transfer` currently does with allowOverwrite=false), REPLACE_EXISTING (equivalent to allowOverwrite=true), and KEEP_EXISTING_VERSIONS (equivalent to transferAndKeep) so that we wouldn't need to add more boolean params for each variation. So the new signature would be `transfer(PID binPid, URI sourceFileUri, TransferMode transferMode)`, and it would conditionally skip all the behaviors that were removed in transferAndKeep when transferMode was KEEP_EXISTING_VERSIONS.
2. `transfer(PID binPid, URI sourceFileUri, TransferMode transferMode)` should become `protected` instead of public, since its not part of the interface and is only used internally by these clients.

bbpennel (changes requested):

bbpennel (approved):

PR #2229, opened 2026-08-25
BXC-4984 update transaction classes to fedora 7

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4984](https://unclibrary.atlassian.net/browse/BXC-4984)

- update `build.yml` and `docker-compose`yml` to include fedora 5 and fedora 7 in parallel
- update `fedora.client.version` to `6.0.0` in `pom.xml`
- cleanup `FcrepoClientFactory`
- update endpoint urls for committing/keeping/canceling transactions in `TransactionManager`
- use txt id to supply `Atomic-ID` header for requests and greatly simplify `TransactionalFcrepoClient`
- use `accept("text/turtle")` instead of `preferMinimal()`, which seems to not be supported in fedora 7
- in `RepositoryInitializer`, wrap `createOrTransformObject` in a try/catch block in `createContainer` method and add `Cdr.ContentRoot` RDF property in `createContentRoot`
- add `RepositoryObjectLoaderFactory` and `RepositoryObjectDriverFactory` to fix circular spring bean dependency error
- fix tests


model-fcrepo/src/main/java/edu/unc/lib/boxc/model/fcrepo/services/ContentPathFactoryImpl.java
```diff
@@ -120,10 +120,10 @@ private class ChildToParentCacheLoader extends CacheLoader<PID, PID> {
         @Override
         public PID load(PID pid) {
             try (FcrepoResponse resp = fcrepoClient.get(pid.getRepositoryUri())
-                    .preferMinimal()
+                    .accept("text/turtle")
```
bbpennel: while including text/turtle here is fine, i'm not sure why any of the other changes in this file would be necessary, like i think we still want this to return "preferMinimal"

model-fcrepo/src/main/java/edu/unc/lib/boxc/model/fcrepo/services/RepositoryInitializer.java
```diff
@@ -75,11 +83,13 @@ private URI createContentRoot(URI contentUri) {
         log.warn("Initializing content root object {}", contentRootUri);
 
         Model model = ModelFactory.createDefaultModel();
-        Resource resc = model.createResource(contentRootString);
-        resc.addProperty(DC.title, "Content Collections Root");
-
+        model.createResource(contentRootString)
+                .addProperty(DC.title, "Content Collections Root")
+                .addProperty(RDF.type, Cdr.ContentRoot);
```
bbpennel: This already gets set in edu.unc.lib.boxc.model.fcrepo.services.RepositoryObjectFactoryImpl#createContentRootObject so i don't think we need to set it here

model-fcrepo/src/main/java/edu/unc/lib/boxc/model/fcrepo/services/RepositoryObjectFactoryImpl.java
```diff
@@ -355,7 +355,7 @@ public BinaryObject createOrUpdateBinary(PID pid, URI storageUri, String filenam
         // Track the URI where metadata updates would be made to for this binary
         URI describedBy;
         log.debug("Updating binary {} with fields:\n    uri = '{}'\n    mimetype = '{}', " +
-                "filename = '{}'\n    sha1 = '{}', m5 = '{}'", pid.getRepositoryPath(),
+                "filename = '{}'\n    sha1 = '{}', m5 = '{}'", pid.getRepositoryUri().toString(),
```
bbpennel: why is getRepositoryPath getting replaced by getRepositoryUri().toString() everywhere? getRepositoryPath is essentially the same thing already

operations/src/test/java/edu/unc/lib/boxc/operations/impl/mimetype/CorrectMimetypesServiceIT.java
```diff
@@ -226,7 +227,8 @@ private FileObject addFileObject(WorkObject workObject, PID filePid, String file
     }
 
     private void assertOriginalFileMimetype(FileObject fileObject, String expectedMimetype) {
-        BinaryObject originalFile = fileObject.getOriginalFile();
+        PID originalPid = DatastreamPids.getOriginalFilePid(fileObject.getPid());
```
bbpennel: I'm not sure why this would have needed to change? I think this is basically what getOriginalFile does.

operations/src/test/java/edu/unc/lib/boxc/operations/impl/versioning/VersionedDatastreamServiceIT.java
```diff
@@ -228,9 +228,6 @@ public void addVersion_DatastreamWithHistory() throws Exception {
         Document headDoc = inputStreamToDocument(dsObjFinal.getBinaryStream());
         assertHasModsTitle("lets leave it here", headDoc);
 
-        assertNotEquals(version2Date, dsObjFinal.getLastModified(),
```
bbpennel, line 224: I don't think anything should have changed with versioned datastream handling

pom.xml
```diff
@@ -71,7 +71,7 @@
         <junit.jupiter.version>5.14.1</junit.jupiter.version>
         <mockito.version>5.20.0</mockito.version>
         <mockito.junit.version>5.20.0</mockito.junit.version>
-        <fcrepo.client.version>5.0.0</fcrepo.client.version>
+        <fcrepo.client.version>6.0.0</fcrepo.client.version>
```
bbpennel: I think there might be a 6.2.0 in maven central

model-fcrepo/src/main/java/edu/unc/lib/boxc/model/fcrepo/services/RepositoryObjectFactoryImpl.java
krwong: getRepositoryPath returns the repository uri as a string but getRepositoryUri returns the full repository uri. I'm not sure what the exact difference is between the repository uri and the full repository uri, but it was causing problems. Based on [this page](https://wiki.lyrasis.org/spaces/FEDORA6x/pages/178882389/External+Content) and [this page](https://wiki.lyrasis.org/spaces/FEDORAM6M0P0/pages/222593827/RDF+Namespaces), Fedora 6 and beyond are expecting absolute uris, so I'm guessing getRepositoryPath is relative

bbpennel (approved):

